PR #161, opened 2016-11-16
App extension support

clpolet:
Here comes the app extension support.
Some use cases needs to be discussed further, for instance: What is a session for an app extension? There is no out of the box life cycle to hook upon. Some extension does not even have UI.
Thus it is not officially supported as for now but at least most of the work is already done and we'll keep the app extension support in mind while coding.


MobileCenter/MobileCenter/Internals/Utils/MSApplicationHelper.h
```diff
@@ -0,0 +1,47 @@
+/*
```
jaelim-ms: Let's remove this. 😉

globalconfig.xcconfig
```diff
@@ -14,4 +14,6 @@ OTHER_CFLAGS[sdk=iphonesimulator9.*] = $(MOBILE_CENTER_WARNING_FLAGS) -fembed-bi
 OTHER_CFLAGS[sdk=iphoneos10.*] = $(MOBILE_CENTER_WARNING_FLAGS) -fembed-bitcode
 OTHER_CFLAGS[sdk=iphonesimulator10.*] = $(MOBILE_CENTER_WARNING_FLAGS) -fembed-bitcode-marker
 
+APPLICATION_EXTENSION_API_ONLY = YES;
```
jaelim-ms: Shouldn't it be NO?

clpolet: This means, when set to YES, that we guarantee to use App extension API only. Which is true while integrated on app extensions.

PR #163, opened 2016-11-16
Remove UUID validation for appSecret

clpolet:
Application secret is now processed as a string and not a UUID.

### DIFF
--- a/MobileCenter/MobileCenter/MSMobileCenter.m
+++ b/MobileCenter/MobileCenter/MSMobileCenter.m
@@ -154,7 +154,7 @@ static NSString *const kMSDefaultBaseUrl = @"https://in.mobile.azure.com";
   }
 
   // Validate and set the app secret.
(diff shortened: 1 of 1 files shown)


MobileCenter/MobileCenter/MSMobileCenter.m
```diff
@@ -154,7 +154,7 @@ - (BOOL)configure:(NSString *)appSecret {
   }
 
   // Validate and set the app secret.
-  else if ([appSecret length] == 0 || ![[NSUUID alloc] initWithUUIDString:appSecret]) {
+  else if ([appSecret length] == 0) {
```
jaelim-ms: We also need to check null/nil here.

clpolet: It's there already, if appSecret is nil then we end up with nil==0 which is true in objective c.

TroubleMakerBen: +1

PR #172, opened 2016-11-17
Changing the download link for iOS modules

elamalani:


elamalani:
Change already in another branch.

PR #174, opened 2016-11-18
Feature/add crash probe cases

TroubleMakerBen:


Puppet/Puppet/ MSFakeCXXClass.h
```diff
@@ -0,0 +1,13 @@
+/*
```
jaelim-ms: The filename contains blank.

Puppet/Puppet/ MSFakeCXXClass.mm
```diff
@@ -0,0 +1,13 @@
+/*
```
jaelim-ms: The filename contains blank.

PR #175, opened 2016-11-18
Swift demo with more test case

vineetchoudhary:
I've added some basic case in swift demo. 
- Track Event
- Generate Test Crash
- Enable or disable Crashes
- Enable or disable Analytics
- App crash in the last session?

simulator screen shot 18-nov-2016 7 34 56 pm

This is my first pull request. :)

### DIFF
--- a/SwiftDemo/SwiftDemo.xcodeproj/project.pbxproj
+++ b/SwiftDemo/SwiftDemo.xcodeproj/project.pbxproj
@@ -157,6 +157,7 @@
 				TargetAttributes = {
 					B2CDD0E81DCCFAAC00610725 = {
 						CreatedOnToolsVersion = 8.1;
+						DevelopmentTeam = 2CG2Z45AZQ;
 						LastSwiftMigration = 0810;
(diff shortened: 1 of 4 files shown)


SwiftDemo/SwiftDemo.xcodeproj/project.pbxproj
```diff
@@ -343,6 +345,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
+				DEVELOPMENT_TEAM = 2CG2Z45AZQ;
```
lumaxis: These probably shouldn't be in this PR 🙂

clpolet: That's right :), I was about to say the same.

vineetchoudhary: Reverted development team to none.

PR #177, opened 2016-11-21
Fix/wrapper crash reporter

achocron:
This is necessary to fix an issue where projects using the crashes service in Xamarin.iOS projects could crash in certain situations when it shouldn't have been. The solution involves setting up signal handlers and enabling PLCrashReporter in a particular order, so that functionality has been refactored slightly, with the addition of a delegate that the wrapper SDK needs to implement in order to know when PLCrashReporter is going to be initialized


achocron:
@TroubleMakerBen There don't seem to be any side effects with native crashes, but I'm going to do a bit more testing before merging

jaelim-ms:
Your build is currently broken. Please take a look.

PR #194, opened 2016-11-30
Add script to publish podspec to internal repo

jaelim-ms:
This PR is for Bitrise scripts which will automatically upload newer spec to internal repo after release process is done.

### DIFF
--- a/bitrise/publish-internal-podspec.sh
+++ b/bitrise/publish-internal-podspec.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+## I. Get publish version for information
+publish_version="$(grep "VERSION_STRING" $BITRISE_SOURCE_DIR/$VERSION_FILE | head -1 | awk -F "[= ]" '{print $4}')"
+echo "Publishing podspec for version" $publish_version
+
+## II. Get path of internal podspec local repo
+local_repo_path="$(pod repo | grep "$GIT_SPEC_REPO_NAME" | grep Path | head -1 | awk -F ": " '{print $2}')"
+
+## III. Update podspec to the internal podspec local repo
+resp="$(pod repo push $GIT_SPEC_REPO_NAME $BITRISE_SOURCE_DIR/$PODSPEC_FILENAME --allow-warnings)"
+echo $resp
+error="$(echo $resp | grep -i error\|fatal)"
+if [ "$error" ]; then
+  echo "Cannot publish to private repo"
+  exit 1
+fi
+
+## IV. Push podspec to the interanl podspec remote repo
+cd $local_repo_path
+git push
+cd $BITRISE_SOURCE_DIR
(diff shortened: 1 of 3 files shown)


bitrise/publish-internal-podspec.sh
```diff
@@ -0,0 +1,52 @@
...
+publish_version="$(grep "VERSION_STRING" $BITRISE_SOURCE_DIR/$VERSION_FILE | head -1 | awk -F "[= ]" '{print $4}')"
+echo "Publishing podspec for version" $publish_version
+
+if [ "$1" == "internal" ]; then
+
+  ## 1. Get path of internal podspec local repo
+  local_repo_path="$(pod repo | grep "$GIT_SPEC_REPO_NAME" | grep Path | head -1 | awk -F ": " '{print $2}')"
+
+  ## 2. Update podspec to the internal podspec local repo
+  resp="$(pod repo push $GIT_SPEC_REPO_NAME $BITRISE_SOURCE_DIR/$PODSPEC_FILENAME --allow-warnings)"
```
TroubleMakerBen: hm...why the `--allow-warnings`-flag?

bitrise/publish-internal-podspec.sh
```diff
@@ -0,0 +1,52 @@
...
+  cd $local_repo_path
+  git push
+  cd $BITRISE_SOURCE_DIR
+
+  echo "Podspec published to internal repo successfully"
+
+else
+
+  ## 1. Run lint to validate podspec. This step will automatically push the spec to CocoaPods if the spec is valid
+  resp="$(pod spec lint $BITRISE_SOURCE_DIR/$PODSPEC_FILENAME --allow-warnings)"
```
TroubleMakerBen: hm...why the --allow-warnings-flag?

TroubleMakerBen: `pod spec lint` does not automatically publish a spec to cocoapods. You have to run `trunk` for that. I'm validation podspecs with `pod spec lint` several times locally before I call `pod trunk push` or whatever the command for publishing is.

bitrise/publish-internal-podspec.sh
jaelim-ms: Yes, that is a bad behavior. We have to fix this warning to eliminate that flag.
`[!] The validator for Swift projects uses Swift 3.0 by default, if you are using a different version of swift you can use a `.swift-version` file to set the version for your Pod. For example to use Swift 2.3, run:`

jaelim-ms: The warning was my machine only issue. It actually failed due to not be able to validating https url. My ruby version was old so I updated and everything looks good now. Removing that flag.

jaelim-ms:
Yes. Still not sure why that happened during the previous release.
I added `pod trunk push` for publishing podspec to CocoaPods.

PR #205, opened 2016-12-08
Implement analytics callbacks for testing purposes

achocron:


TroubleMakerBen:
Looks good to me. I just don't know the motivation behind this.

achocron:
@TroubleMakerBen The reason it helps is that it enables us to do better UITests from Xamarin. Without it, there is no way to make sure that events are being sent.

TroubleMakerBen:
Thx for clarifying. Feel free to merge =)

MobileCenterAnalytics/MobileCenterAnalytics/Internals/MSAnalyticsDelegate.h
```diff
@@ -0,0 +1,27 @@
...
+
+@class MSEventLog;
+@class MSPageLog;
+@class MSAnalytics;
+
+@protocol MSAnalyticsDelegate <NSObject>
+
+@optional
+
+- (void)analytics:(MSAnalytics *)analytics willSendEventLog:(MSEventLog *)eventLog;
```
jaelim-ms: docs please.

MobileCenterAnalytics/MobileCenterAnalytics/MSAnalytics.m
```diff
@@ -208,4 +216,62 @@ - (void)sessionTracker:(id)sessionTracker processLog:(id<MSLog>)log withPriority
...
+  }
+}
+
+- (void)channel:(id<MSChannel>)channel didSucceedSendingLog:(id<MSLog>)log {
+  if (!self.delegate) {
+    return;
+  }
+  NSObject *logObject = (NSObject *)log;
+  if ([logObject isKindOfClass:[MSEventLog class]] &&
+      [self.delegate respondsToSelector:@selector(analytics:didSucceedSendingEventLog:)]) {
```
jaelim-ms: Do you have any reasons of having different delegate methods per log type?

MobileCenterAnalytics/MobileCenterAnalytics/Internals/MSAnalyticsDelegate.h
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ */
+
+#import <Foundation/Foundation.h>
+
+@class MSEventLog;
+@class MSPageLog;
+@class MSAnalytics;
+
+@protocol MSAnalyticsDelegate <NSObject>
```
jaelim-ms: The reason of the separate delegate class for Crashes is that we expose the delegate to customers.
I don't think this is the case, you can just implement MSChannelDelegate for Analytics. Maybe it is required because you have other methods per log type. Look at below comment.

MobileCenterAnalytics/MobileCenterAnalytics/MSAnalytics.m
achocron: @jaelim-ms I just figured that since they are different types, they could be handled differently. And would save any client code from having to deal with MSLogWithProperties and casting. I don't really have any strong opinions on the matter though. (The interface is not public, anyways. It's only going to be used by Xamarin UITests.)

MobileCenterAnalytics/MobileCenterAnalytics/Internals/MSAnalyticsDelegate.h
jaelim-ms: Discussed offline. This will only be used for Xamarin.

MobileCenterAnalytics/MobileCenterAnalytics/MSAnalytics.m
jaelim-ms: Discussed offline. The only thing is to have one method for all log types. Don't duplicate methods per type.

PR #221, opened 2017-01-03
Feature/fix bitcode warnings

clpolet:
Fix bitcode and nullability warnings + small fix on tracking APIs signature.


Global.xcconfig
```diff
@@ -8,11 +8,18 @@ MS_SIM_ARCHS = x86_64 i386
 ARCHS = $(MS_ARM_ARCHS)
 VALID_ARCHS = $(MS_ARM_ARCHS)
 
+// This flag adds -fembed-bitcode-marker in the release configuration
```
jaelim-ms: If you rename the file, you should update environment variable in Bitrise build.

MobileCenterAnalytics/MobileCenterAnalytics/Internals/MSAnalyticsInternal.h
```diff
@@ -23,7 +25,7 @@
  * @param pageName  page name.
  * @param properties dictionary of properties.
  */
-+ (void)trackPage:(NSString *)pageName withProperties:(NSDictionary *)properties;
++ (void)trackPage:(NSString *)pageName withProperties:(nullable NSDictionary<NSString *, NSString *> *)properties;
```
jaelim-ms: What is this change? Merging issue?

clpolet: To fix the nullable warnings we had here I added NS_ASSUME_NONNULL_BEGIN/END but doing so I had to explicitly specify nullable parameters.

jaelim-ms: Don't mind. I just checked previous PR and found that I missed the change in the header file. 😄 
BTW, I was talking about the key/value types for NSDictionary, not nullable.

Global.xcconfig
jaelim-ms: In addition to the comment, you might need to change two environment variables.
`VERSION_FILE` and `VERSION_FILENAME`. Eventually we are going to use `VERSION_FILENAME` but just not to break the build, I keep both variables. Please update before you merge to develop.

clpolet:
More unit tests added to improve code coverage.

TroubleMakerBen:
Looking good.

PR #222, opened 2017-01-03
Fix crash sent on disabled

clpolet:
Crashs captured during disabled state are now deleted.

### DIFF
--- a/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
+++ b/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
@@ -157,17 +157,17 @@ static void uncaught_cxx_exception_handler(const MSCrashesUncaughtCXXExceptionIn
   // Enabling
   if (isEnabled) {
     
-    // PLCrashReporter keeps collecting crash reports even when the SDK is disabled,
-    // delete them only if current state is disabled.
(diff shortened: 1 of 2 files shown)


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
```diff
@@ -154,7 +154,15 @@ - (instancetype)init {
 - (void)applyEnabledState:(BOOL)isEnabled {
   [super applyEnabledState:isEnabled];
 
+  // Enabling
   if (isEnabled) {
+    
+    // PLCrashReporter keeps collecting crash reports even when the SDK is disabled,
+    // delete them only if current state is disabled.
+    if (!self.isEnabled){
+      [self.plCrashReporter purgePendingCrashReport];
```
jaelim-ms: Is this actually disable (or stop) `PLCrashReporter`? Would be good to have a unit test for this scenario.

clpolet: This is just purging old reports as we can't disable PLCrasReporter.

clpolet: I'll add a new unit test.

PR #225, opened 2017-01-06
Fix/fix cxxexception report

TroubleMakerBen:


TroubleMakerBen:
I'll improve unit tests for this (should be done anyway) to make codecov happy.

PR #235, opened 2017-01-11
Fix/suspend http tasks

clpolet:
Http session tasks should be suspended but not cancelled when the sender is suspended. Particularly because task cancellation is leveraged in another case where  the sender receives a fatal error.

This also stops/resets call's retry when sender is suspended.


MobileCenter/MobileCenterTests/MSHttpSenderTests.m
```diff
@@ -234,60 +195,193 @@ - (void)testNetworkUpAgain {
...
+  // Send the container.
+  [self.sut sendAsync:container
+    completionHandler:^(NSString *batchId, NSError *error, NSUInteger statusCode) {
+      XCTFail(@"Completion handler shouldn't be called as test will finish before the retry will expire.");
+    }];
+  
+  // Mock the call to intercept the retry.
+  MSRetriableCall *mockedCall = OCMPartialMock(self.sut.pendingCalls[@"1"]);
+  OCMStub([mockedCall sender:self.sut callCompletedWithStatus:MSHTTPCodesNo500InternalServerError error:[OCMArg any]]).andForwardToRealObject().andDo(^(NSInvocation *invocation){
+           NSLog(@"TEST1: %@", ((MSRetriableCall*)self.sut.pendingCalls[@"1"]).timerSource);//TODO remove
```
clpolet: Just noticed that I forgot to remove this line, will do ^^

clpolet:
Weird the failed unit test was working locally, I'll have a look.

PR #238, opened 2017-01-12
Fix/fix crash callbacks on 404

clpolet:
Fix other non sent crash callback not triggered on HTTP fatal error
    
    * Callback didFailSendingLog now also called on logs stored and but not sent in case of non retriable error
    * Optimize channel enabling/disabling process within log manager
    * Return deleted logs while deleting logs from a storage key
    * Add unit test

### DIFF
--- a/MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
+++ b/MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
@@ -85,7 +85,7 @@ static NSUInteger const MSDefaultLogCountLimit = 50;
   [self deleteFile:[bucket fileWithId:logsId] fromBucket:bucket];
 }
 
-- (NSArray<MSLog> *)deleteFile:(MSFile *)file fromBucket:(MSStorageBucket *)aBucket {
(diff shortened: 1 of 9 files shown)


MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
```diff
@@ -61,32 +61,45 @@ - (void)saveLog:(id<MSLog>)log withStorageKey:(NSString *)storageKey {
...
+  return deletedLogs;
 }
 
 - (void)deleteLogsForId:(NSString *)logsId withStorageKey:(NSString *)storageKey {
   MSStorageBucket *bucket = self.buckets[storageKey];
-  MSFile *file = [bucket fileWithId:logsId];
+  [self deleteFile:[bucket fileWithId:logsId] fromBucket:bucket];
+}
 
+- (NSArray<MSLog> *)deleteFile:(MSFile *)file fromBucket:(MSStorageBucket *)aBucket {
```
jaelim-ms: Can we just say `bucket`?

clpolet: Yes I'll do the change.

PR #240, opened 2017-01-13
Do not log Crashes enabled/disabled when a debugger is attached

jaelim-ms:


jaelim-ms:
What about the log in [line 398](https://github.com/Microsoft/mobile-center-sdk-ios/blob/feature/hide-enabled-disabled-logs-when-debugger-attached/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m#L398)?
Please DO NOT MERGE unless the label is removed.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
```diff
@@ -187,7 +187,8 @@ - (void)applyEnabledState:(BOOL)isEnabled {
     if (self.crashFiles.count > 0) {
       [self startDelayedCrashProcessing];
     }
-    MSLogInfo([MSCrashes getLoggerTag], @"Crashes service has been enabled.");
+    if (![MSMobileCenter isDebuggerAttached])
+      MSLogInfo([MSCrashes getLoggerTag], @"Crashes service has been enabled.");
```
lumaxis: One-line `if`s without curly braces?! Sacrilege! 🙂

lumaxis:
I'd just like to chime in here real quick and provide a bit of experience from maintaining the HockeyApp SDKs.
In my experience, figuring out if crash reporting works/trying it with Xcode/debugger connected has been a big point of confusion in the past. Even though we logged a big message that crash reporting is disabled with the debugger attached, many devs still wondered and contacted support about why HockeyApp didn't report any crashes when the app they were debugging crashed.

This just as an advice to make this super clear as this is something that easily confuses people 👍

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
jaelim-ms: We have coding styles for that. If any statements have one-line body, we don't use curly braces unless it is not easy to read. I know there are some statements with curly braces, we are currently mix-using them. We will have an internal discussion and make it consistent.

TroubleMakerBen: That shouldn't be in there at all.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.m
```diff
@@ -188,6 +188,14 @@ - (void)applyEnabledState:(BOOL)isEnabled {
       [self startDelayedCrashProcessing];
     }
     MSLogInfo([MSCrashes getLoggerTag], @"Crashes service has been enabled.");
+
+    // More details on log if a debugger is attached.
+    if ([MSMobileCenter isDebuggerAttached]) {
+      MSLogInfo([MSCrashes getLoggerTag], @"Crashes service has been enabled but the service cannot detect crashes due to the application with a debugger attached.");
```
TroubleMakerBen: It should be "due to running the application with a debugger attached"

PR #261, opened 2017-01-31
Keychain utility in core

jaelim-ms:


MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.h
```diff
@@ -0,0 +1,79 @@
...
+//  HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+//  WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+//  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+//  OTHER DEALINGS IN THE SOFTWARE.
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+#ifndef MSKeychainUtil_h
```
TroubleMakerBen: Just something I only noticed now:
we're adding `#ifndef` as a `#include`-guard. As Objc usually uses `#import`, this is redundant (it already does the `guard` for us). The guard doesn't really hurt, but I think we should either have it in all .h-files, or nowhere except where necessary.

jaelim-ms: Sorry, I removed it from other places but forgot to do here. Will make a change for that.

TroubleMakerBen: 👍🏻

PR #263, opened 2017-02-01
Feature/refactor core

jaelim-ms:


MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -334,12 +327,20 @@ - (NSString *)hideSecret:(NSString *)secret {
...
+          (secret.length > kMSMaxCharactersDisplayedForAppSecret ? secret.length - kMSMaxCharactersDisplayedForAppSecret
+                  : secret.length);
   NSString *appSecretHiddenPart =
-      [@"" stringByPaddingToLength:appSecretHiddenPartLength withString:kMSHidingStringForAppSecret startingAtIndex:0];
+          [@"" stringByPaddingToLength:appSecretHiddenPartLength withString:kMSHidingStringForAppSecret startingAtIndex:0];
   return [secret stringByReplacingCharactersInRange:NSMakeRange(0, appSecretHiddenPart.length)
                                          withString:appSecretHiddenPart];
 }
 
+- (NSArray *)retryIntervals {
```
TroubleMakerBen: Please document that this is the "empty" implementation that is intended to be overwritten.

jaelim-ms: Done.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -334,12 +327,26 @@ - (NSString *)hideSecret:(NSString *)secret {
...
-      [@"" stringByPaddingToLength:appSecretHiddenPartLength withString:kMSHidingStringForAppSecret startingAtIndex:0];
+          [@"" stringByPaddingToLength:appSecretHiddenPartLength withString:kMSHidingStringForAppSecret startingAtIndex:0];
   return [secret stringByReplacingCharactersInRange:NSMakeRange(0, appSecretHiddenPart.length)
                                          withString:appSecretHiddenPart];
 }
 
+/**
+ * This is a method without implementation and is intended to be overwritten in all sub classes.
+ */
+- (NSArray *)retryIntervals {
```
clpolet: I would suggest something else: `retryIntervals` and `apiPath` being readonly properties and being set via a new initializer. That way we don't need the new `MSIngestionHttpSender` class and the empty implementations of `retryIntervals` and `apiPath`.

jaelim-ms: As discussed offline, we will use properties as @clpolet suggested. We need to come back if we have many properties configurable in the future.

PR #264, opened 2017-02-01
Make sure we don't loose any events in case of a crash

TroubleMakerBen:


MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -82,6 +82,10 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
 
 - (void)processLog:(id<MSLog>)log withPriority:(MSPriority)priority {
 
+  if(!log) {
```
jaelim-ms: Can you reformat this file?

MobileCenter/MobileCenter/Internals/Storage/MSStorage.h
```diff
@@ -37,8 +37,10 @@ typedef void (^MSLoadDataCompletionBlock)(BOOL succeeded, NSArray<MSLog> *logArr
  *
  * @param log The log item that should be written to disk
  * @param storageKey The key used for grouping
+ *
```
jaelim-ms: Probably you don't need a new line here. And run reformat for this file please.

MobileCenter/MobileCenterTests/MSLogManagerDefaultTests.m
```diff
@@ -14,6 +14,17 @@ @interface MSLogManagerDefaultTests : XCTestCase
 
 @implementation MSLogManagerDefaultTests
 
+
+#pragma mark - Housekeeping
+
+- (void)setUp {
```
jaelim-ms: Do you need this and `tearDown` now? If not, please remove.

MobileCenter/MobileCenterTests/MSLogManagerDefaultTests.m
```diff
@@ -49,4 +60,23 @@ - (void)testProcessingWithNewPriorityWillCreateNewChannel {
   assertThat(sut.channels[@(priority)], notNilValue());
 }
 
+- (void)testProcessingLogWillTriggerOnProcessingCall {
+  // If
```
jaelim-ms: New line before comment.

MobileCenter/MobileCenter/Internals/Storage/MSStorage.h
TroubleMakerBen: We have newlines after `@param` and before `@return` quite often.

jaelim-ms: OK. Then let's take this rule for other places.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -240,6 +272,61 @@ - (MSPriority)priority {
   return MSPriorityHigh;
 }
 
+#pragma mark - MSLogManagerDelegate
```
clpolet: Also might be good for our future selves to add a comment saying: We had no choice than to implement the last chance storage/flush logic in the crash module even if it concerns logs from any services because we can't contact any other module while crashing.

If I am not mistaken ^^.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -432,6 +517,28 @@ - (void)processCrashReport {
...
+  NSArray *files = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:self.logBufferDir
+                                                                       error:NULL];
+  for (NSString *tmp in files) {
+    if ([[tmp pathExtension] isEqualToString:kMSLogBufferFileExtension]) {
+      NSString *filePath = [self.logBufferDir stringByAppendingPathComponent:tmp];
+      NSData *serializedLog = [NSData dataWithContentsOfFile:filePath];
+
+      id <MSLog> item = [NSKeyedUnarchiver unarchiveObjectWithData:serializedLog];
+      if (item) {
+        if ([((NSObject *) item) isKindOfClass:[MSAppleErrorLog class]]) {
```
TroubleMakerBen: The problem is that crashes doesn't know about other types of logs at all. The only reliable way that I can think of right now would be to introduce a priority-property on MSLog, which I think is out of scope for now.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
TroubleMakerBen: I'll add the comment.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
TroubleMakerBen: Created new tasks to improve this

PR #266, opened 2017-02-03
Pass appSecret to each service when it starts

jaelim-ms:
This PR will merge to `feature/refactor-core` branch since it has implementation dependencies with that.

### DIFF
--- a/MobileCenter/MobileCenter/Internals/MSServiceCommon.h
+++ b/MobileCenter/MobileCenter/Internals/MSServiceCommon.h
@@ -21,7 +21,12 @@
 /**
  * Log manager.
  */
-@property(nonatomic) id<MSLogManager> logManager;
+@property(nonatomic) id <MSLogManager> logManager;
+
+/**
+ * The app secret for the SDK.
+ */
+@property(nonatomic, nonnull) NSString *appSecret;
 
 /**
  * Apply the enabled state to the service.
@@ -38,7 +43,7 @@
  * @discussion: IMPORTANT, This string is used to point to the right storage value for this service.
  * Changing this string results in data lost if previous data is not migrated.
  */
-@property(nonatomic, readonly) NSString *storageKey;
+@property(nonatomic) NSString *storageKey;
 
 /**
  * The channel priority for this service.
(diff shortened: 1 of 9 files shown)


MobileCenter/MobileCenter/Internals/MSServiceCommon.h
```diff
@@ -38,7 +43,7 @@
  * @discussion: IMPORTANT, This string is used to point to the right storage value for this service.
  * Changing this string results in data lost if previous data is not migrated.
  */
-@property(nonatomic, readonly) NSString *storageKey;
+@property(nonatomic) NSString *storageKey;
```
TroubleMakerBen: We should think about adding the `copy` attribute to properties of type `NSString` everywhere, including this one. I'm creating a task for this. It's really quick to do.

MobileCenter/MobileCenter/Internals/MSServiceCommon.h
```diff
@@ -38,7 +43,7 @@
  * @discussion: IMPORTANT, This string is used to point to the right storage value for this service.
  * Changing this string results in data lost if previous data is not migrated.
  */
-@property(nonatomic, readonly) NSString *storageKey;
```
clpolet: Is there a reason why the storage key lost its readonly attribute?

MobileCenterAnalytics/MobileCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -31,19 +29,21 @@ @interface MSAnalyticsTests : XCTestCase<MSAnalyticsDelegate>
...
 @end
 
 @interface MSServiceAbstract ()
 
-- (BOOL) isEnabled;
-- (void) setEnabled:(BOOL)enabled;
+- (BOOL)isEnabled;
 
-@end
+- (void)setEnabled:(BOOL)enabled;
```
clpolet: This one is not on you ;) but you can remove the whole MSAnalytics and MSServiceAbstract class extensions from here. You'll probably need to add the `#import "MSServiceAbstractProtected.h"` import.

PR #275, opened 2017-02-09
Refactor HttpSender to take generic object for request body

jaelim-ms:


MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -23,7 +22,10 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Pending http calls.
  */
-@property(atomic, strong) NSMutableDictionary<NSString *, id <MSSenderCall>> *pendingCalls;
+@property(atomic, strong) NSMutableDictionary<NSString *, MSSenderCall *> *pendingCalls;
+
+// TODO (jaelim): Add doc here.
+- (void)sendAsync:(NSObject *)container callId:(NSString *)callId completionHandler:(MSSendAsyncCompletionHandler)handler;
```
clpolet: I would try to use a generic type on the class itself and use it here instead of `NSObject`. Then you may want to restrict the type within the `MSIngestionSender`.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -60,33 +59,8 @@ - (id)initWithBaseUrl:(NSString *)baseUrl
...
-      call.logContainer = container;
-      call.completionHandler = handler;
-
-      // Store call in calls array.
-      self.pendingCalls[batchId] = call;
-    }
-    [self sendCallAsync:call];
-  }
+- (void)sendAsync:(NSObject *)data completionHandler:(MSSendAsyncCompletionHandler)handler {
+  [self sendAsync:data callId:MS_UUID_STRING completionHandler:handler];
```
clpolet: Might be overkill to generate a UUID on each request, a simple increment with same life than this object would be enough I guess.

MobileCenter/MobileCenter/Internals/Sender/MSIngestionSender.h
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ */
```
clpolet: Got it copy past from existing file ^^

TroubleMakerBen: we have the header more or less everywhere but not in tests. If we don't need it, let's just get rid of it in a sep. PR.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
jaelim-ms: This is a default behavior but we don't expect anyone uses this direct call.

clpolet: But that's what we'll be using for distribution right?

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -23,7 +22,12 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Pending http calls.
  */
-@property(atomic, strong) NSMutableDictionary<NSString *, id <MSSenderCall>> *pendingCalls;
+@property(atomic, strong) NSMutableDictionary<NSString *, MSSenderCall *> *pendingCalls;
+
+// TODO (jaelim): Add doc here.
+- (void)sendAsync:(NSObject *)container
```
clpolet: I would suggest to use a generic type on the class and then use it there instead of an `NSObject`. Then you'll be able to restrict the type on the `MSIngestionSender` class.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
jaelim-ms: Oh.. Yes. But it is still required because distribution http sender will have retry rules we have to maintain `MSSenderCall`.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.h
jaelim-ms: I tried but it looks I couldn't. `NSSender` is a protocol that can't have generics. Maybe I can try with `__kindof` but I'm not sure how it will work at the moment.
I'd like to leave this as it is for now, we can use generics later.

jaelim-ms:
@clpolet wants testing on his side. Waiting for his feedback.

clpolet:
I can see some new warnings on the Puppet:
`2017-02-10 09:16:27.523 Puppet[308:15274] *** -[NSKeyedUnarchiver initForReadingWithData:]: data is empty; did you forget to send -finishEncoding to the NSKeyedArchiver?`

Maybe something wrong with log deserialization?

jaelim-ms:
The issue was caused by some changes in `develop`. I merged the latest `develop` to the working branch and the issue is gone.

PR #276, opened 2017-02-09
Fix/warning settings and fixes

TroubleMakerBen:


Global.xcconfig
```diff
@@ -53,4 +52,10 @@ GCC_PREPROCESSOR_DEFINITIONS = $(inherited) $(XCODEBUILD_GCC_PREPROCESSOR_DEFINI
 // -padded
 // This isn’t really an issue to us, since we’re not programming embedded systems.
 
-WARNING_CFLAGS = -Weverything -Wno-objc-missing-property-synthesis -Wno-float-equal -Wno-pedantic -Wno-padded  -Wno-sign-conversion
+// -Wc++98
```
jaelim-ms: -c++98-compat?

TroubleMakerBen: Yes, if we don't add `-Wno-c++98-compat`, the compiler will ommit warnings that we use C++11 features. We're deliberately using C++11 and don't care about C++98 ;)

TroubleMakerBen: Oooooh, I now get what you are saying. I'll update the comment 🙈

jaelim-ms: 😄

PR #286, opened 2017-02-14
Save buffered logs in buckets per priority and...

TroubleMakerBen:
... sends them in their original priority on the next launch.


MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -14,6 +14,7 @@
 #import "MSServiceAbstractPrivate.h"
 #import "MSServiceAbstractProtected.h"
 #import "MSUtil.h"
+#import "MSCrashesUtil.h"
```
jaelim-ms: You probably need one more unit test that verifies your implementation that `PriorityHigh` should go first.

MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
```diff
@@ -28,15 +29,18 @@ struct MSCrashesBufferedLog {
                                             &reinterpret_cast<const char *>(data.bytes)[data.length]) {}
 };
 
+
+
```
jaelim-ms: Unnecessary lines. 😄

MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
```diff
@@ -88,8 +92,12 @@ typedef struct MSCrashesCallbacks {
 
 /**
  * The index for our log buffer. It keeps track of where we want to buffer our next event.
+ * The keys are MSPriority cast to NSNumber.
```
jaelim-ms: Good to learn. enum cannot be a type of key.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -43,16 +43,18 @@ static void ms_save_log_buffer_callback(siginfo_t *info, ucontext_t *uap, void *
...
-    if (fd < 0) {
-      return;
+  for (auto it = msCrashesLogBuffer.begin(), end = msCrashesLogBuffer.end(); it != end; ++it) {
+    for (int i = 0; i < ms_crashes_log_buffer_size; i++) {
+      // Make sure not to allocate any memory (e.g. copy).
+      const std::string &data = it->second[i].buffer;
+      const std::string &path = it->second[i].bufferPath;
+      int fd = open(path.c_str(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+      if (fd < 0) {
+        return;
```
jaelim-ms: `return` or `break`? You are exiting the method.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -319,14 +322,16 @@ - (void)onProcessingLog:(id<MSLog>)log withPriority:(MSPriority)priority {
   @synchronized(self) {
     NSData *serializedLog = [NSKeyedArchiver archivedDataWithRootObject:log];
     if (serializedLog && (serializedLog.length > 0)) {
-      if (self.bufferIndex > (ms_crashes_log_buffer_size - 1)) {
-        self.bufferIndex = 0;
+      if ([[self.bufferIndex objectForKey:@(priority)] integerValue] > (ms_crashes_log_buffer_size - 1)) {
```
jaelim-ms: `>= ms_crashes_log_buffer_size`??

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -319,14 +322,16 @@ - (void)onProcessingLog:(id<MSLog>)log withPriority:(MSPriority)priority {
...
+        // Reset the counter to 0 for a priority.
```
jaelim-ms: New line before comment

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -43,16 +43,18 @@ static void ms_save_log_buffer_callback(siginfo_t *info, ucontext_t *uap, void *
...
-  for (int i = 0; i < ms_crashes_log_buffer_size; i++) {
-
-    // Make sure not to allocate any memory (e.g. copy).
-    const std::string &data = msCrashesLogBuffer[i].buffer;
-    int fd = open(msCrashesLogBuffer[i].bufferPath.c_str(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
-    if (fd < 0) {
-      return;
+  for (auto it = msCrashesLogBuffer.begin(), end = msCrashesLogBuffer.end(); it != end; ++it) {
+    for (int i = 0; i < ms_crashes_log_buffer_size; i++) {
+      // Make sure not to allocate any memory (e.g. copy).
```
jaelim-ms: New line before comment

MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
jaelim-ms: Also one more for verifying files more than 20. 😄

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -43,16 +43,18 @@ static void ms_save_log_buffer_callback(siginfo_t *info, ucontext_t *uap, void *
...
+      const std::string &data = it->second[i].buffer;
+      const std::string &path = it->second[i].bufferPath;
+      int fd = open(path.c_str(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+      if (fd < 0) {
+        return;
+      }
```
gwynne: Use `continue` instead of `return` here, since failing to open any one path doesn't imply that all of them will fail.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
gwynne: I would write `self.bufferIndex[@(priority)].integerValue` here, visually simpler to parse IMO

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -319,14 +322,16 @@ - (void)onProcessingLog:(id<MSLog>)log withPriority:(MSPriority)priority {
   @synchronized(self) {
     NSData *serializedLog = [NSKeyedArchiver archivedDataWithRootObject:log];
     if (serializedLog && (serializedLog.length > 0)) {
-      if (self.bufferIndex > (ms_crashes_log_buffer_size - 1)) {
-        self.bufferIndex = 0;
+      if ([[self.bufferIndex objectForKey:@(priority)] integerValue] > (ms_crashes_log_buffer_size - 1)) {
+        // Reset the counter to 0 for a priority.
+        [self.bufferIndex setObject:@0 forKey:@(priority)];
```
gwynne: Likewise, `self.bufferIndex[@(priority)] = @0;` here

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -319,14 +322,16 @@ - (void)onProcessingLog:(id<MSLog>)log withPriority:(MSPriority)priority {
...
       }
-      msCrashesLogBuffer[self.bufferIndex].buffer =
+      NSInteger index = [[self.bufferIndex objectForKey:@(priority)] integerValue];
+      msCrashesLogBuffer[priority][index].buffer =
           std::string(&reinterpret_cast<const char *>(serializedLog.bytes)[0],
                       &reinterpret_cast<const char *>(serializedLog.bytes)[serializedLog.length]);
-
-      self.bufferIndex += 1;
+      index += 1;
+      [self.bufferIndex setObject:@(index) forKey:@(priority)];
```
gwynne: `self.bufferIndex[@(priority)] = @(++index);`

Well, YMMV on whether to use the preincrement form, it's a little unobvious.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -524,18 +529,22 @@ - (void)processCrashReports {
...
+    NSString *dirForPriority = [self bufferDirectoryForPriority:(MSPriority)priority];
+    NSArray *files = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:dirForPriority error:NULL];
+    for (NSString *tmp in files) {
+      if ([[tmp pathExtension] isEqualToString:kMSLogBufferFileExtension]) {
+        NSString *filePath = [dirForPriority stringByAppendingPathComponent:tmp];
+        NSData *serializedLog = [NSData dataWithContentsOfFile:filePath];
+        if (serializedLog && serializedLog.length && serializedLog.length > 0) {
+          id<MSLog> item = [NSKeyedUnarchiver unarchiveObjectWithData:serializedLog];
+          if (item) {
+            [self.logManager processLog:item withPriority:(MSPriority)priority];
```
gwynne: For this whole stanza, rewrite using URL versions of the APIs in question, per Apple recommendations, convert to string path only when required.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -640,75 +649,84 @@ - (void)createAnalyzerFile {
...
+      // Init the counter for each priority.
+      [self.bufferIndex setObject:@0 forKey:@(priority)];
 
-    // Initialize the bufferIndex and write pre-fill the filepaths of the MSCrashesBufferedLog
-    self.bufferIndex = 0;
-    for (NSUInteger i = 0; i < count; i++) {
-      MSCrashesBufferedLog emptyLog = MSCrashesBufferedLog{bufferFiles[i], nil};
-      msCrashesLogBuffer[i] = emptyLog;
+      // Create a buffer for the priority. Making use of `{}` as we're using C++11.
+      msCrashesLogBuffer[(MSPriority)priority][{}];
```
gwynne: While it works, using the "`[]` implicitly emplaces" form is technically executing a void expression for its side effects. Prefer an explicit insert or emplace - `msCrashesLogBuffer.insert(static_cast<MSPriority>(priority), {});`. Also, due to that same side effect, this initialization is unnecessary anyhow; it will happen automatically the first time through the loop below.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -640,75 +649,84 @@ - (void)createAnalyzerFile {
...
-  // Create missing buffer files if needed.
-  if (logBufferFiles.count < ms_crashes_log_buffer_size) {
-    NSInteger missingFileCount = ms_crashes_log_buffer_size - logBufferFiles.count;
-    for (int i = 0; i < missingFileCount; i++) {
-      NSString *logId = MS_UUID_STRING;
-      NSString *path = [self createBufferFileWithName:logId];
-      [logBufferFiles addObject:path];
-    }
+    // Files are named N.mscrasheslogbuffer where N is between 0 and ms_crashes_log_buffer_size.
+    NSString *logId = [NSString stringWithFormat:@"%ld", i];
```
gwynne: `NSString *logId = @(i).stringValue;` <-- faster and type-safer than invoking a full `stringWithFormat:`

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
gwynne: `>` is more semantically expressive and no more or less confusing IMO 😆

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
TroubleMakerBen: I guess YMMV applies here. 
I think with `self.bufferIndex[@(priority)] = @(++index);`, I'd add a comment to point out the increment to make it less likely that someone doesn't see it. In this case, I can also keep `index += 1`. ;)

gwynne: Yeah, that's definitely one where clarity wins over brevity, as a rule. I have a tendency to write extremely dense code because I can read extremely dense code due to ~brain damage~ lots of practice 😅

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
jaelim-ms: Yeah.. But this is a different case I think. For `>`, it is doing unnecessary subtraction. 😄

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
TroubleMakerBen: I have created a new task for this as this applies to the whole SDK, not just this module.

MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
TroubleMakerBen: About your first comment:
Crashes is just looping over the buffered logs which should happen really fast, channel takes care of the prioritization. I don't think Crashes should care about what a priority actually means.

2nd comment: Adding one!

PR #287, opened 2017-02-14
Fix warnings related to direct access of ivars

erychagov:
I created a special property for class MSCrashesCXXExceptionWrapperException.m, because I don't know an another way for fix warning.

Also I moved two properties from MSCrashesPrivate.h to MSCrashes.mm because this properties have been using only inside class MSCrashes and I found only this way for fix.

Let me know if you have other suggestion how I could fix it


TroubleMakerBen:
Hey @erychagov,
Thx for the PR! Please check the unit tests and make sure to grab the latest changes from /develop.

🖖
Benny

erychagov:
@TroubleMakerBen 
Unit tests pass

PR #288, opened 2017-02-14
Rename setters for URL

TroubleMakerBen:
Update with latest names.


jaelim-ms:
Can we hold merging this branch? I have a big change that will cause a big pain of merge conflict. Since this is small so resolving conflicts on this PR would be much easier. What do you think?

TroubleMakerBen:
@jaelim-ms sure thing

PR #290, opened 2017-02-15
Ignore -Wc++98-compat-pedantic

TroubleMakerBen:
We can ignore this warning, too, as we don't care about C++98 compatibility.


Global.xcconfig
```diff
@@ -55,7 +55,10 @@ GCC_PREPROCESSOR_DEFINITIONS = $(inherited) $(XCODEBUILD_GCC_PREPROCESSOR_DEFINI
 // -c++98
 // We don't want to compile our code for C++98, no need to be warned about incompatibility.
 
+// -Wc++98-compat-pedantic
```
jaelim-ms: No `W`. Just `-c++98-compat-pedantic`

TroubleMakerBen:
Fixed the comment. Merging.

PR #298, opened 2017-02-21
Feature/url scheme

jaelim-ms:


MobileCenterUpdates/MobileCenterUpdates/MSUpdates.h
```diff
@@ -16,4 +16,11 @@
  */
 + (void)setInstallUrl:(NSString *)installUrl;
 
+/**
+ * Take a request via custom URL scheme from browser.
```
clpolet: I would rephrase it like this : "Process URL related to this module". May be triggered not only by browsers in the future. 

Also may be add a discussion saying that default integration is to place this call into the app delegate openUrl method.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
```diff
@@ -106,14 +96,53 @@ - (void)applyEnabledState:(BOOL)isEnabled {
...
+
++ (void)setApiUrl:(NSString *)apiUrl {
+  [[self sharedInstance] setApiUrl:apiUrl];
+}
+
++ (void)setInstallUrl:(NSString *)installUrl {
+  [[self sharedInstance] setInstallUrl:installUrl];
+}
+
++ (void)openUrl:(NSURL *)url {
```
clpolet: This one may be protected by `if ([[self sharedInstance] canBeUsed])` if it's not supposed to be called when the  module is not started yet.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
```diff
@@ -365,39 +362,80 @@ - (BOOL)isNewerVersion:(MSReleaseDetails *)details {
...
+      return;
+    }
+
+    // Parse query parameters
+    NSString *requestedId = [MS_USER_DEFAULTS objectForKey:kMSUpdateTokenRequestIdKey];
+    NSString *queryRequestId = nil;
+    NSString *queryUpdateToken = nil;
+    NSURLComponents *components = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:NO];
+
+    for (NSURLQueryItem *item in components.queryItems) {
```
clpolet: new line without comment ;)

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
```diff
@@ -365,39 +362,80 @@ - (BOOL)isNewerVersion:(MSReleaseDetails *)details {
...
+- (void)openUrl:(NSURL *)url {
+  if ([self isEnabled]) {
+
+    // If the request is not for Mobile Center Updates, ignore.
+    if (![kMSUpdtsDefaultCustomScheme isEqualToString:[url scheme]]) {
+      return;
+    }
+
+    // Parse query parameters
+    NSString *requestedId = [MS_USER_DEFAULTS objectForKey:kMSUpdateTokenRequestIdKey];
```
clpolet: We may delete this request Id from the persist once processed.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
```diff
@@ -365,39 +362,80 @@ - (BOOL)isNewerVersion:(MSReleaseDetails *)details {
...
+    }
+
+    // If the request ID doesn't match, ignore.
+    if (!(requestedId && queryRequestId && [requestedId isEqualToString:queryRequestId])) {
+      return;
+    }
+
+    // Store update token
+    if (queryUpdateToken) {
+      [MSKeychainUtil storeString:queryUpdateToken forKey:kMSUpdateTokenKey];
```
clpolet: Maybe add a log here saying the update token has been successfully retrieved. No need to print it I think or hide it as it's a secret.

MobileCenterUpdates/MobileCenterUpdates/MSUpdatesPrivate.h
```diff
@@ -47,4 +62,31 @@ static NSString *const kMSUpdtsDefaultCustomScheme = @"msupdt";
  */
 - (void)openURLInSafariApp:(NSURL *)url;
 
+/**
+ * Take a request via custom URL scheme from browser.
+ *
+ * @param url  The url with parameters.
+ */
+- (void)openUrl:(NSURL *)url;
```
clpolet: I would rephrase it like this : "Process URL related to this module". May be triggered not only by browsers in the future.

Also may be add a discussion saying that default integration is to place this call into the app delegate openUrl method.

MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.h
```diff
@@ -13,7 +13,18 @@ NS_ASSUME_NONNULL_BEGIN
  *
```
jaelim-ms: My bad on copy and paste.

MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.h
```diff
@@ -22,7 +33,17 @@ NS_ASSUME_NONNULL_BEGIN
  * Delete a key and a string for the given service name.
```
jaelim-ms: Revised.

MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.h
```diff
@@ -31,15 +52,33 @@ NS_ASSUME_NONNULL_BEGIN
  * Get a string from Keychain with the given key and service name.
```
jaelim-ms: Revised.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
jaelim-ms: It shouldn't matter either the service has been started or not. It should persist update token at any time. And I'll take care of enable/disable/canBeUsed in another task that is already assigned to me.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
jaelim-ms: Good point. I will be deleted only if requested id can be verified.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
jaelim-ms: The rule is that we have a new line before comments. This doesn't mean no line breaking if there is no comment. If we don't have line breaks, readability will be terrible.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
jaelim-ms: Added.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.h
jaelim-ms: Done.

MobileCenterUpdates/MobileCenterUpdates/MSUpdatesPrivate.h
jaelim-ms: Done.

MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.m
```diff
@@ -3,20 +3,30 @@
 
 @implementation MSKeychainUtil
```
jaelim-ms: Our rule is that there should be a new line before comment, not new line only before comment. No?

MobileCenter/MobileCenter/Internals/Utils/MSKeychainUtil.h
```diff
@@ -31,15 +52,33 @@ NS_ASSUME_NONNULL_BEGIN
...
  * @return A string data if exists.
  */
 + (NSString *)stringForKey:(NSString *)key service:(NSString *)service;
 
 /**
  * Clear all keys and strings associated with the given service name.
  *
+ * @return YES if cleared successfully, NO otherwise.
+ */
++ (BOOL)clearForService;
```
jaelim-ms: Changing this to use mainBundle. We don't need to have service name as a parameter.
This needs to be done in `develop`.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.h
```diff
@@ -3,7 +3,7 @@
 @interface MSUpdates : MSServiceAbstract
```
jaelim-ms: Completely forgot about this one. Will be added to PR.

jaelim-ms: For now we only have Updates service that requires custom URL scheme. I'm pretty sure that it should be per service because core doesn't know about each service. If we have only one scheme for the SDK, core should understand what the request means and then send it over to a proper service. I'm pretty sure during this process, it will go through all started services to find an appropriate service. The benefit of simple integration would be less than expect I think.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
clpolet: The rule is no new line without comment within code blocks. For code clarity you can add a comment here.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.h
jaelim-ms: We are going to use pre-define format so it is not necessary. I'm make a relevant change for this one.

jaelim-ms: Same thing. Ignore this one. We will use pre-defined format.

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.m
jaelim-ms: Took offline and added a comment for the block

MobileCenterUpdates/MobileCenterUpdates/MSUpdates.h
clpolet: Actually I didn't realized but it can't be customized since it has to be unique per app ^^

PR #303, opened 2017-02-22
Restore build UUID for update module

clpolet:
We can now reuse a build UUID to request the update token. The code use CodeSignatureExtraction framework directly, it's a workaround. It doesn't have all the constraints that we have on the SDK + switch frameworks doesn't have ABI compatibility. We should rewrite it to Objective C and use this source code directly. 

- Fix buildUUID format.
- Bring back CodeSignatureExtraction framework.
- Update demos. 

[Edit] CodeSignatureExtraction rewriting is done now, CodeSignatureExtraction framework removed.


clpolet:
It takes too much time to fix CI issue caused by the import of the CodeSignatureExtraction framework, I'll re-write it in Objective-C now. That was planned anyway, we'll only port the code that we use + it will be easier to match all the restrictions we have on the SDK.

clpolet:
Will do unit tests in another PR.

MobileCenter/MobileCenter/Internals/Utils/MSUtil.m
```diff
@@ -92,4 +94,25 @@ + (NSTimeInterval)nowInMilliseconds {
...
+
++ (NSString *)formatToUUIDString:(NSString *)aString {
+  NSMutableString *stringToFormat = [aString mutableCopy];
+  short dashesCount = (sizeof kMSUUIDDashIndexes) / (sizeof kMSUUIDDashIndexes[0]);
+
+  // Pre-validate string.
+  if (aString.length != (NSUInteger)(kMSUUIDLength - dashesCount)) {
+    return nil;
+  }
+  for (short i = 0; i < dashesCount; i++) {
```
jaelim-ms: Minor thing but you can iterate from at the end of string to modify. This case the index would have better readability like `8, 12, 16, 20`.

clpolet: Good idea, though we actually don't need this anymore :).
I removed it.

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -464,10 +473,14 @@
 		6E2307ED1D22083C0031C598 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = NO;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME = "Brand Assets";
 				ENABLE_BITCODE = NO;
-				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
+				FRAMEWORK_SEARCH_PATHS = (
```
TroubleMakerBen: Why this change? This used to be done on a project level right?

clpolet: Was because of the CodeSignatureExtraction framework. But don't hurt to leave it. Also I harmonized to all apps and it's now at the project level ;)

PR #318, opened 2017-03-01
Feature/download update

jaelim-ms:
Implementation of downloading updates + fixed alert controller issue.


jaelim-ms:
Tried but no luck. Created a task for the unit test.

PR #336, opened 2017-03-03
Feature/rename to distribute

jaelim-ms:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.h
```diff
@@ -1,9 +1,9 @@
...
 
 /**
- * Mobile Center updates service.
+ * Mobile Center Distribute service.
  */
-@interface MSUpdates : MSServiceAbstract
+@interface MSDistribute : MSServiceAbstract
 
 /**
  * Change The URL that will be used for generic update related tasks, e.g. fetching the auth token..
```
clpolet: 'Distribute service' instead of 'update' here maybe?

MobileCenter.podspec
```diff
@@ -47,10 +47,10 @@ Pod::Spec.new do |s|
...
 
-#  s.subspec 'MobileCenterUpdates' do |ss|
+#  s.subspec 'MobileCenterDistribute' do |ss|
 #   ss.frameworks = 'CoreTelephony', 'Foundation'
 #   ss.dependency 'MobileCenter/MobileCenter'
-#  ss.resource_bundle = { 'MobileCenterUpdatesResources' => ['MobileCenter-SDK-iOS/MobileCenterUpdatesResources.bundle/*.lproj'] }
+#  ss.resource_bundle = { 'MobileCenterDistributeResources' => ['MobileCenter-SDK-iOS/MobileCenterDistributeResources.bundle/*.lproj'] }
 # end
 
 end
```
clpolet: And also update the changelog in the mean time.

jaelim-ms: It will be determined when we merge to develop. The version doesn't mean anything in this stage.

jaelim-ms: I think the change log would be something like introducing new features. I think This would be done in develop branch as well.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.h
jaelim-ms: This is for `setApiUrl` which only takes care of update (not install). Hard to say distribute here since distribute consists of install as well as update.

PR #339, opened 2017-03-06
Fix/set log url anytime

clpolet:
Allow updating the log URL at runtime. This is done in the base HTTP sender, will need to do some small work to port this to base/updates branch.

### DIFF
--- a/MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
+++ b/MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
@@ -1,8 +1,8 @@
-#import <Foundation/Foundation.h>
 #import "MSLogManager.h"
 #import "MSMobileCenter.h"
 #import "MSServiceInternal.h"
 #import "MobileCenter+Internal.h"
+#import <Foundation/Foundation.h>
 
 // Persisted storage keys.
 static NSString *const kMSInstallIdKey = @"MSInstallId";
@@ -16,7 +16,7 @@ static NSString *const kMSMobileCenterIsEnabledKey = @"MSMobileCenterIsEnabled";
 @property(nonatomic, copy) NSString *logUrl;
 @property(nonatomic, readonly) NSUUID *installId;
 @property(nonatomic, copy) NSString *apiVersion;
-@property BOOL sdkConfigured;
+@property(atomic) BOOL sdkConfigured;
 @property(atomic) BOOL enabledStateUpdating;
 
 /**
--- a/MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
+++ b/MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
@@ -1,8 +1,9 @@
+#import <Foundation/Foundation.h>
+
 #import "MSLogManager.h"
 #import "MSMobileCenter.h"
 #import "MSServiceInternal.h"
 #import "MobileCenter+Internal.h"
-#import <Foundation/Foundation.h>
 
 // Persisted storage keys.
 static NSString *const kMSInstallIdKey = @"MSInstallId";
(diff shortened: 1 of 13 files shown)


MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -6,9 +6,14 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
+ * Base URL (schema + authority + port only) used to communicate with the server.
+ */
+@property(nonatomic) NSString *baseURL;
+
+/**
  *	Send Url.
  */
-@property(nonatomic, strong, readonly) NSURL *sendURL;
+@property(nonatomic, strong) NSURL *sendURL;
```
TroubleMakerBen: we don't need to specify `strong` here. Should we add "Do not specify default attributes for properties" as a coding rule?

clpolet: Why not. I know there are two schools about this, people that wants everything specified to be explicit but some other prefer the concise way for aesthetic.

clpolet: Ok so we'll skip the defaults for now on. But for consistency we'll do that on another PR for the whole code.

TroubleMakerBen: Yep.

MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
```diff
@@ -1,8 +1,8 @@
-#import <Foundation/Foundation.h>
```
jaelim-ms: OK. So what is a rule for `#import <header>` and `#import "header"`? I remember we do `<>` and then `""`.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -62,7 +68,12 @@ - (id)initWithBaseUrl:(NSString *)baseUrl
          queryStrings:(NSDictionary *)queryStrings
          reachability:(MS_Reachability *)reachability
        retryIntervals:(NSArray *)retryIntervals {
-  return [self initWithBaseUrl:baseUrl apiPath:@"" headers:headers queryStrings:queryStrings reachability:reachability retryIntervals:retryIntervals];
+  return [self initWithBaseUrl:baseUrl
```
jaelim-ms: Not related to your change but we can remove this method.

MobileCenter/MobileCenter/Internals/Sender/MSSender.h
```diff
@@ -11,7 +11,17 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol MSSender <NSObject, MSSenderCallDelegate, MSEnable>
 
 /**
- *  Reachability library.
+ * Base URL (schema + authority + port only) used to communicate with the server.
```
jaelim-ms: This is a generic sender. I don't think we need to place those properties in here. Need to go to http sender, what do you think?

clpolet: I moved apiPath which is only used by senders.

MobileCenter/MobileCenter/Internals/Sender/MSHttpSender.m
clpolet: done

MobileCenter/MobileCenter/Internals/MSMobileCenterInternal.h
clpolet: done

MobileCenter/MobileCenter/Internals/Sender/MSSender.h
jaelim-ms: As per offline conversation, you need to make one more change for baseURL.

PR #340, opened 2017-03-06
Add error message

TroubleMakerBen:
We want to make sure that our SDK was integrated correctly, including the resource bundle for Distribute.
Compile-time checks for resources don't exist so the alternative would be assertions (checks at runtime). I don't like the idea of our SDK causing any crashes, not even when running in debug config or with a debugger attached, so my proposal is to log something.

### DIFF
--- a/MobileCenterDistribute/MobileCenterDistribute/MSDistributeUtil.m
+++ b/MobileCenterDistribute/MobileCenterDistribute/MSDistributeUtil.m
@@ -1,6 +1,7 @@
 #import "MSDistribute.h"
 #import "MSDistributeInternal.h"
 #import "MSDistributeUtil.h"
+#import "MSLogger.h"
 
 NSBundle *MSDistributeBundle(void) {
   static NSBundle *bundle = nil;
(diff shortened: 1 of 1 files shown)


MobileCenterDistribute/MobileCenterDistribute/MSDistributeUtil.m
```diff
@@ -11,6 +12,12 @@
     NSString *mainBundlePath = [[NSBundle bundleForClass:[MSDistribute class]] resourcePath];
     NSString *frameworkBundlePath = [mainBundlePath stringByAppendingPathComponent:MOBILE_CENTER_DISTRIBUTE_BUNDLE];
     bundle = [NSBundle bundleWithPath:frameworkBundlePath];
+    
+    // Log to console in case the bundle is nil.
+    if(!bundle) {
+      MSLogWarning([MSDistribute logTag], @"The MobileCenterDistributeResources.bundle file could not be found in your"
```
jaelim-ms: Should it be a warning or an error?

TroubleMakerBen: Good question. I was not sure. Thinking about it now, I'm changing this to error as the SDK wasn't integrated properly.

PR #345, opened 2017-03-07
Fix linker flags to inherit from our xcconfig file

TroubleMakerBen:
I don't know when this got thrown under the bus but I just noticed and fixed it.


jaelim-ms:
Took offline to discuss branch issues. We merge into base/updates.

PR #382, opened 2017-03-16
Fix/bring safari view to front

clpolet:
- Show Safari view controller as a modal view

- Dismiss it while disabling or if update token succeeded

### DIFF
--- a/MobileCenterDistribute/MobileCenterDistribute/MSDistributePrivate.h
+++ b/MobileCenterDistribute/MobileCenterDistribute/MSDistributePrivate.h
@@ -43,6 +43,11 @@ static NSString *const kMSUpdateTokenKey = @"MSUpdateToken";
 @interface MSDistribute ()
 
 /**
+ * View controller presenting the `SFSafariViewController`.
+ */
+@property(nonatomic) UIViewController *safariHostingViewController;
+
+/**
  * Build the install URL for token request with the given application secret.
  *
  * @param appSecret Application secret.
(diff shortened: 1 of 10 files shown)


MobileCenterDistribute/MobileCenterDistribute/MSDistributePrivate.h
```diff
@@ -43,6 +43,11 @@ static NSString *const kMSUpdateTokenKey = @"MSUpdateToken";
 @interface MSDistribute ()
 
 /**
+ * View controller presenting the `SFSafariViewController`.
+ */
+@property(nonatomic) UIViewController *safariHostingViewController;
```
jaelim-ms: Any reason that you keep this as a property? Opening a browser is just one time thing, you can dispose it after it is done.

clpolet: We have to automatically dispose it later for instance if the module is disabled or if the openURL method is called with the right token. To do so we need to access the hosting view controller hence the property :)

MobileCenterDistribute/MobileCenterDistribute/Support/MobileCenterDistribute.xcconfig
```diff
@@ -1,3 +1,3 @@
 #include "../../../Global.xcconfig"
 
-OTHER_LDFLAGS=$(inherited) -framework Foundation -framework UIKit -framework CoreTelephony -framework Security
+OTHER_LDFLAGS=$(inherited) -framework Foundation -framework UIKit -framework CoreTelephony
```
TroubleMakerBen: AFAIK this needs to include SafariServices, too. (It's added in the podspec but not here)

clpolet: Right

PR #383, opened 2017-03-16
Fix/fix openurl freeze

clpolet:
Fix openURL iOS API freezing the app launch.


clpolet:
Base branch messed up

PR #385, opened 2017-03-16
Fix openURL freeze application start

clpolet:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -383,7 +383,11 @@ - (void)openURLInEmbeddedSafari:(NSURL *)url fromClass:(Class)clazz {
 
 - (void)openURLInSafariApp:(NSURL *)url {
   MSLogDebug([MSDistribute logTag], @"Using Safari browser to open URL: %@", url);
-  [MSUtil sharedAppOpenUrl:url options:@{} completionHandler:nil];
+  
+  /* Dispatch the open url call to the next loop to avoid freezing the App new instance start up */
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [MSUtil sharedAppOpenUrl:url options:@{} completionHandler:nil];
```
jaelim-ms: Isn't it better to have it in `sharedAppOpenUrl`?
What if it is called from main thread? You don't need to check current thread?

clpolet: Right I can move it to "sharedAppOpenUrl" for future use. 
The only thing we have to make sure is it's not using openURL in the same loop than the init one which is what we are doing here.

jaelim-ms: Then you can add TODO comment there.

clpolet:
Forgot to push the small change yesterday @jaelim-ms :)

jaelim-ms:
Not unit test is failing. Looks the previous implementation is wrong.

PR #387, opened 2017-03-17
Rewrite file-access using NSURL-API as per Apple's reccommendation

MatkovIvan:
Replace NSString for paths everywhere and use NSURL instead.


MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
```diff
@@ -165,38 +165,36 @@ - (void)renewCurrentFileForStorageKey:(NSString *)storageKey {
...
-      _baseDirectoryPath = [appSupportPath stringByAppendingPathComponent:kMSLogsDirectory];
+- (NSURL *)baseDirectoryURL {
+  if (!_baseDirectoryURL) {
+    NSURL *appSupportURL = [[[NSFileManager defaultManager] URLsForDirectory:NSApplicationSupportDirectory inDomains:NSUserDomainMask] lastObject];
+    if (appSupportURL) {
+      _baseDirectoryURL = (NSURL * _Nonnull)[appSupportURL URLByAppendingPathComponent:kMSLogsDirectory];
     }
 
-    MSLogVerbose([MSMobileCenter logTag], @"Storage Path:\n%@", _baseDirectoryPath);
+    MSLogVerbose([MSMobileCenter logTag], @"Storage Path:\n%@", self->_baseDirectoryURL);
```
clpolet: Is there a reason to not call _baseDirectoryURL directly here.

MobileCenter/MobileCenter/Internals/Storage/MSFileUtil.m
```diff
@@ -43,87 +43,86 @@ - (NSFileManager *)fileManager {
...
+  if (![directoryURL checkResourceIsReachableAndReturnError:&error]) {
     return nil;
   }
 
   NSMutableArray<MSFile *> *files;
-  NSError *error;
-  NSArray *allFiles = [fileManager contentsOfDirectoryAtPath:path error:&error];
+  NSArray *allFiles = [fileManager contentsOfDirectoryAtURL:(NSURL * _Nonnull)directoryURL
+                                 includingPropertiesForKeys:@[NSURLCreationDateKey]
+                                                    options:NSDirectoryEnumerationSkipsHiddenFiles
```
clpolet: I would not skip hidden files to be more future proof.
@jaelim-ms @TroubleMakerBen Any though?

MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
TroubleMakerBen: Can this be done without direct ivar access?

MobileCenter/MobileCenter/Internals/Storage/MSFileUtil.m
TroubleMakerBen: I agree that we shouldn't skip hidden files.

MatkovIvan: [link](https://developer.apple.com/reference/foundation/nsfilemanager/1413768-contentsofdirectoryaturl) - "the only supported option is `NSDirectory​Enumeration​Skips​Hidden​Files`"

jaelim-ms: I think what the others say is to pass `nil`. @MatkovIvan. Doesn't it work?

MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
TroubleMakerBen: why not `self.baseDirectoryURL`?

MobileCenter/MobileCenter/Internals/Storage/MSFileUtil.m
TroubleMakerBen: I didn't look at the API, if `NSDirectory​Enumeration​Skips​Hidden​Files` this is fine. We don't create any hidden files anyway.

MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
clpolet: Yep but it's not the only var accessed directly, if we go that way we should propertify them all in this class.

clpolet: Yep should be a property you are right.

MobileCenter/MobileCenter/Internals/Storage/MSFileUtil.m
clpolet: At the end if the options can't be `nil` then yes we can use that option.

MobileCenter/MobileCenter/Internals/Storage/MSFileStorage.m
MatkovIvan: This is auto-fix by XCode for fixing warning:
`Block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior`
Also with `self.baseDirectoryURL` this is recursion.

MatkovIvan:
@clpolet please let me know if there is anything which should be updated.

PR #395, opened 2017-03-18
Feature/use group id for storage

TroubleMakerBen:


TroubleMakerBen:
This introduces the groupID for persistence and includes some cleanup for all classes that was touched by the groupID-change.

One thing that I want to point out is that the CrashBuffer uses it's own groupID, which is okay in my opinion as it will never contain more than 20 logs. We could make the buffer use the log's own groupIDs but need to adjust the LogManagerDelegate-callbacks.

MobileCenter/MobileCenter/Internals/Channel/MSLogManager.h
```diff
@@ -35,40 +35,51 @@ NS_ASSUME_NONNULL_BEGIN
...
  *  @param deleteData A boolean value set to YES to delete data or NO to keep it.
- *  @param priority A priority to enable/disable.
+ *  @param groupID A groupID to enable/disable.
+ *  @param priority The priority of the groupID to enable/disable.
  */
-- (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData forPriority:(MSPriority)priority;
+- (void)setEnabled:(BOOL)isEnabled
+    andDeleteDataOnDisabled:(BOOL)deleteData
+                 forGroupID:(NSString *)groupID
+               withPriority:(MSPriority)priority;
```
jaelim-ms: Do you still need to know priority? `groupID` is enough to delete data.

MobileCenter/MobileCenter/Internals/Channel/MSLogManager.h
```diff
@@ -35,40 +35,51 @@ NS_ASSUME_NONNULL_BEGIN
...
  * Remove a delegate to each channel that has a certain priority.
  *
  * @param channelDelegate A delegate for the channel.
+ * @param groupID The groupID of a channel.
  * @param priority The priority of a channel.
  */
-- (void)removeChannelDelegate:(id<MSChannelDelegate>)channelDelegate forPriority:(MSPriority)priority;
+- (void)removeChannelDelegate:(id<MSChannelDelegate>)channelDelegate
+                   forGroupID:(NSString *)groupID
+                 withPriority:(MSPriority)priority;
```
jaelim-ms: Need priority?

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -82,7 +86,7 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
 
 #pragma mark - Process items
 
-- (void)processLog:(id<MSLog>)log withPriority:(MSPriority)priority {
+- (void)processLog:(id<MSLog>)log withPriority:(MSPriority)priority andGroupID:(NSString *)groupID {
```
jaelim-ms: I wonder if you can get rid of a priority parameter from all (or most of) methods.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,3 +1,7 @@
+
```
jaelim-ms: Please delete lines.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -658,7 +659,9 @@ - (void)processLogBufferAfterCrash {
         if (serializedLog && serializedLog.length && serializedLog.length > 0) {
           id<MSLog> item = [NSKeyedUnarchiver unarchiveObjectWithData:serializedLog];
           if (item) {
-            [self.logManager processLog:item withPriority:MSPriority(priority)];
+            
+            // Buffered logs are used sending their own channel. It will never contain more than 20 logs
+            [self.logManager processLog:item withPriority:(MSPriority)priority andGroupID:@"CrashBuffer"];
```
jaelim-ms: I doubt this is a reason of why you are keeping priority even after introducing groupID.

TroubleMakerBen:
As discussed: All API that use MSPriority create new channels on the fly, so the answer is “no, I can’t get rid of MSPriority without completely changing our batching logic".

MobileCenter/MobileCenter/Internals/MSServiceInternal.h
```diff
@@ -11,13 +11,6 @@
 @protocol MSServiceInternal <MSService, MSServiceCommon>
```
clpolet: `+@property(nonatomic, copy, readonly) NSString *groupID;` Should be there too as required so that any feature must implement it.

TroubleMakerBen: I just added that – I'm very sure I had that at some point. Good catch!

PR #397, opened 2017-03-20
Create categories for utility

bmourat:
MSUtil class was refactored in MSUtility class divided into different categories


TroubleMakerBen:
LGTM

PR #399, opened 2017-03-20
Fix unused parameter warnings

MatkovIvan:


TroubleMakerBen:
Good job here!

Could you go ahead and check for usage of `#pragma GCC diagnostic ignored "-Wunused-parameter"` just in case and move to your version? I think we have at least 1 occurrence.  Thanks!

MatkovIvan:
@clpolet this PR related to unused **parameters**. Method `(UIApplicationState)sharedAppState` don't have parameters and you talk about unused **variable**.

clpolet:
@MatkovIvan you are right, not the same warning. Though I would still have fixed this one because it's a really quick thing to do. But I am fine with doing this later.

MatkovIvan:
@clpolet It's should already done in last (35d616c) commit.

PR #401, opened 2017-03-20
Update endpoint to PROD

jaelim-ms:


Puppet/Puppet/AppDelegate.m
```diff
@@ -21,9 +21,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
   // Start Mobile Center SDK.
   [MSMobileCenter setLogLevel:MSLogLevelVerbose];
-  [MSMobileCenter setLogUrl:@"https://in-integration.dev.avalanch.es"];
 
-  [MSMobileCenter start:@"65dc3680-7325-4000-a0e7-dbd2276eafd1"
+  [MSMobileCenter start:@"7dfb022a-17b5-4d4a-9c75-12bc3ef5e6b7"
```
clpolet: Is it normal to have puppet and Sasquatch swift having the same value here?
Also the plist should be updated.

jaelim-ms: Nope. The values are now different.
info.plist got updated.

PR #404, opened 2017-03-21
Mock NSUserDefaults

erychagov:
Remove MS_USER_DEFAULTS from tests.

### DIFF
--- a/MobileCenter/MobileCenterTests/Util/MSMockUserDefaults.h
+++ b/MobileCenter/MobileCenterTests/Util/MSMockUserDefaults.h
@@ -1,13 +1,4 @@
-//
-//  MSMockUserDefaults.h
-//  MobileCenter
-//
-//  Created by Евгений Рычагов on 23/03/2017.
-//  Copyright © 2017 Microsoft. All rights reserved.
-//
-
-#ifndef MSMockUserDefaults_h
-#define MSMockUserDefaults_h
+#import <Foundation/Foundation.h>
 
 @interface MSMockUserDefaults : NSObject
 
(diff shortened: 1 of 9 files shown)


MobileCenter/MobileCenterTests/Util/MSMockUserDefaults.h
```diff
@@ -0,0 +1,25 @@
+//
+//  MSMockUserDefaults.h
+//  MobileCenter
+//
+//  Created by Евгений Рычагов on 23/03/2017.
+//  Copyright © 2017 Microsoft. All rights reserved.
+//
```
clpolet: We usually don't add headers, could you remove this one please?

MobileCenter/MobileCenterTests/Util/MSMockUserDefaults.m
```diff
@@ -0,0 +1,68 @@
...
+@property(nonatomic) id mockUserDefaults;
+
+@end
+
+@implementation MSMockUserDefaults
+
+@synthesize dictionary;
+@synthesize mockUserDefaults;
+
+- (instancetype)init {
```
clpolet: It terms of code convention we have this rule:  Empty lines not preceded by a comment within a code block is not allowed.

For instance this:
```
       id object;
       [invocation getArgument:&object atIndex:2];
 
       if (!object) {
         return;
       }
```

Should be this:
```
       id object;
       [invocation getArgument:&object atIndex:2];

       // Don't store nil objects.
       if (!object) {
         return;
       }
```

PR #405, opened 2017-03-21
Improve code coverage in Crashes module

MatkovIvan:
Added more tests for Crashes module.

PR #416, opened 2017-03-22
Revert change to MSCrashes.mm’s log buffer feature that were introduc…

TroubleMakerBen:
…ed in PR #405 

I overlooked this change in PR #405.
a) my comment explains why we init with `nil` and it also fixes the issue. Imho the old crash buffer logic actually renders the buffer useless. This would actually crash the app if we weren't already crashing (the data of msLogBuffer is accessed in post_crash_callback). 

cc @MatkovIvan

### DIFF
--- a/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
+++ b/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
@@ -790,16 +790,18 @@ static void uncaught_cxx_exception_handler(const MSCrashesUncaughtCXXExceptionIn
       // Create a buffer for the priority. Making use of `{}` as we're using C++11.
       for (NSUInteger i = 0; i < ms_crashes_log_buffer_size; i++) {
 
+        // We need to convert the NSURL to NSString as we cannot safe NSURL to our async-safe log buffer.
+        NSString *path = files[i].path;
+        
         /**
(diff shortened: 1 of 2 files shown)


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -771,7 +772,17 @@ - (void)setupLogBuffer {
...
+        /**
+         * Some explanation why we are initializing the msCrashLogBuffer with `files[i], nil` instead of `files[i]`:
+         * Not assigning `nil` leads to potentially unpredictable runtime behavior if you have any code which assumes
+         * the second field is always some valid or `nil` value (since failing to explicitly initialize, it permits the
+         * compiler to leave garbage in that field - zero-init is only guaranteed for Objective-C objects and static
+         * data, it is especially NOT guaranteed for a stack temporary like here. Even if zero-init were guaranteed,
+         * this is [Objective-]C++ so it's not legal to borrow ObjC's "nil is always zero"-rule so technically it would
+         * still not be properly initialized despite being a trivial type.
+         */
+        msCrashesLogBuffer[MSPriority(priority)][i] = MSCrashesBufferedLog{files[i], nil};
```
jaelim-ms: Please refresh your branch. This line has a conflict, `files[i]` -> `[files[i] path]`.

TroubleMakerBen:
Fixed regression.

PR #420, opened 2017-03-23
Fix unused parameter warnings for Push module

MatkovIvan:


MatkovIvan:
In objective c methods it not possible. Its consistent with PR #399. I missed that warning because its placed in preprocessor condition:
'''
#if !(TARGET_OS_SIMULATOR)
'''

PR #426, opened 2017-03-24
Feature/block app for mandatory update

jaelim-ms:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -474,11 +475,16 @@ - (void)showConfirmationAlert:(MSReleaseDetails *)details {
...
-    [alertController addCancelActionWithTitle:MSDistributeLocalizedString(@"Download")
-                                      handler:^(__attribute__((unused)) UIAlertAction *action) {
-                                        MSLogDebug([MSDistribute logTag], @"Start download and install the update.");
-                                        [self startDownload:details];
-                                      }];
+    [alertController
+        addCancelActionWithTitle:MSDistributeLocalizedString(@"Download")
+                         handler:^(__attribute__((unused)) UIAlertAction *action) {
+                           MSLogDebug([MSDistribute logTag], @"Start download and install the update.");
+#if TARGET_IPHONE_SIMULATOR
```
TroubleMakerBen: Maybe put in a comment that this case should not happen normally?

jaelim-ms: It could happen if it is running on simulator. Adding a comment for why we are doing it instead.

PR #431, opened 2017-03-27
Shared ui for sasquatch applications.

bmourat:
Sasquatch applications reimplemented as a single project Sasquatch with two different targets: SasquatchSwift and SasquatchObjC. This allows to share ui code and avoid its duplication.


Sasquatch/Sasquatch/Info.plist
```diff
@@ -16,14 +16,12 @@
 	<string>APPL</string>
```
clpolet: Looks like this info.plist under Sasquatch/ is not used and can be deleted. There is already two other under each apps

clpolet: Or maybe the other way around, if this one can be shared as it's using project constants then probably the others could be removed.

Sasquatch/Sasquatch/Info.plist
```diff
@@ -43,20 +41,5 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>CFBundleURLTypes</key>
-	<array>
-		<dict>
-			<key>CFBundleTypeRole</key>
-			<string>None</string>
-			<key>CFBundleURLName</key>
```
jaelim-ms: This should stay.

Sasquatch/Sasquatch/Info.plist
```diff
@@ -43,20 +41,5 @@
...
-			<string>com.microsoft.azure.mobile.sasquatchobjc</string>
-			<key>CFBundleURLSchemes</key>
```
jaelim-ms: This should stay.

Sasquatch/SasquatchObjC/Info.plist
```diff
@@ -34,20 +34,12 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>CFBundleURLTypes</key>
+	<key>UISupportedInterfaceOrientations~ipad</key>
 	<array>
-		<dict>
-			<key>CFBundleTypeRole</key>
-			<string>None</string>
-			<key>CFBundleURLName</key>
```
jaelim-ms: This should stay.

Sasquatch/SasquatchObjC/Info.plist
```diff
@@ -34,20 +34,12 @@
...
-			<string>com.microsoft.azure.mobile.sasquatchswift</string>
-			<key>CFBundleURLSchemes</key>
```
jaelim-ms: This should stay.

Sasquatch/Sasquatch/Info.plist
jaelim-ms: We should maintain two different info.plist for URL scheme. Probably it is good to remove the one in Sasquatch/

bmourat: @clpolet @jaelim-ms  So i'am deleting Info.plist in Sasquatch/ living two different plist files for Swift and ObjC versions?

Sasquatch/SasquatchObjC/Info.plist
```diff
@@ -41,5 +41,18 @@
...
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>None</string>
+			<key>CFBundleURLName</key>
+			<string>com.microsoft.azure.mobile.sasquatchobjc</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>mobilecenter-3ccfe7f5-ec01-4de5-883c-f563bbbe147a</string>
```
clpolet: This one should be mobilecenter-0dbca56b-b9ae-4d53-856a-7c2856137d85 to match the App key from the start call in the app delegate.

jaelim-ms: Good point, thanks @clpolet. Unfortunately this is right. @bmourat, you need to change it in AppDelegate. Let me see if I can do this.

PR #448, opened 2017-03-31
Fix serializing error with fatal property.

erychagov:
Added test for this case. Also I sorted files and folders by type and name under MobileCenterCrashesTests folder

### DIFF
--- a/MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
+++ b/MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
@@ -509,14 +509,20 @@ static NSString *const kMSFatal = @"fatal";
 - (void)testAbstractErrorLogSerialization {
   MSAbstractErrorLog *log = [MSAbstractErrorLog new];
 
-  // If
-  log.fatal = NO;
-
   // When
   NSDictionary *serializedLog = [log serializeToDictionary];
 
   // Then
-  XCTAssertNotNil([serializedLog objectForKey:kMSFatal]);
+  XCTAssertFalse([serializedLog objectForKey:kMSFatal]);
+
+  // If
+  log.fatal = NO;
+
+  // When
+  serializedLog = [log serializeToDictionary];
+
+  // Then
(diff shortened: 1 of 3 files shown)


MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -505,6 +506,28 @@ - (void)testWrapperCrashCallback {
...
+  MSAbstractErrorLog *log = [MSAbstractErrorLog new];
+
+  // If
+  log.fatal = NO;
+
+  // When
+  NSDictionary *serializedLog = [log serializeToDictionary];
+
+  // Then
+  XCTAssertNotNil([serializedLog objectForKey:kMSFatal]);
```
jaelim-ms: Check the value is NO.

MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -505,6 +506,28 @@ - (void)testWrapperCrashCallback {
...
+
+  // If
+  log.fatal = YES;
+
+  // When
+  serializedLog = [log serializeToDictionary];
+
+  // Then
+  XCTAssertNotNil([serializedLog objectForKey:kMSFatal]);
```
jaelim-ms: Check the value is YES.

MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -505,6 +506,28 @@ - (void)testWrapperCrashCallback {
   XCTAssertTrue(first.count == 1);
 }
 
+- (void)testAbstractErrorLogSerialization {
+  MSAbstractErrorLog *log = [MSAbstractErrorLog new];
```
jaelim-ms: You probably want to test the value is NO after serialization even though you don't assign any value to log.fatal.

dhei:
Looks good to me

PR #450, opened 2017-03-31
Persist mandatory updates

clpolet:
* Last mandatory update is persisted
* If no network on next app start then the persisted mandatory update will be used
* Add replaceAlert feature to MSAlertController to replace any existing alert with another one
* If network is back then replace current mandatory alert with new one


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -233,6 +243,22 @@ - (void)checkLatestRelease:(NSString *)updateToken releaseHash:(NSString *)relea
                                                 withString:@"/"];
                 }
                 MSLogDebug([MSDistribute logTag], @"Received a response of update request:\n%@", jsonString);
+
+                /*
```
jaelim-ms: Can you look at the line feeds in the comment?

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -189,6 +188,17 @@ - (void)checkLatestRelease:(NSString *)updateToken releaseHash:(NSString *)relea
...
+          [[MSReleaseDetails alloc] initWithDictionary:[MS_USER_DEFAULTS objectForKey:kMSMandatoryReleaseKey]];
+      if (self.mandatoryRelease && ![self handleUpdate:self.mandatoryRelease]) {
+
+        // This release is no more a candidate for update, deleting it.
+        [self deleteMandatoryRelease];
+      }
+    }
+
     // Check if sender is still waiting for a response of the previous request.
     if (self.sender == nil) {
```
jaelim-ms: else if?

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -503,7 +543,8 @@ - (void)showConfirmationAlert:(MSReleaseDetails *)details {
 
     // Show the alert controller.
     MSLogDebug([MSDistribute logTag], @"Show update dialog.");
-    [alertController show];
+    [alertController replaceAlert:self.updateAlertController];
```
jaelim-ms: Why do you need to keep this one? Alert is one time only so it can stay as a local variable.

clpolet: This is to cover the case where: You have no network > start the app > the persisted mandatory update alert is displayed > then the network comes back > a new release is there > we need to replace the old alert with the new one.

That's it, we need to remember the old alert to be able to replace it later if there is need.

clpolet:
Some issues with the NSUserDefaults mocking in test, going to fix it.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
clpolet: We still enqueue the request to the sender so that when the network will be back then it will automatically fetch the new release.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
clpolet: Done ;)

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
```diff
@@ -378,6 +380,143 @@ - (void)testShowConfirmationAlertForMandatoryUpdate {
...
+                                 OCMReject([alertControllerMock alertControllerWithTitle:[OCMArg any]
+                                                                                 message:[OCMArg any]]);
+                                 OCMReject(
+                                           [alertControllerMock addDefaultActionWithTitle:[OCMArg any] handler:[OCMArg any]]);
+                                 OCMReject(
+                                           [alertControllerMock addCancelActionWithTitle:[OCMArg any] handler:[OCMArg any]]);
+                               }];
+}
+
+// FIXME: Somehow this test is failing. NSUserDefaults mock seems not to be working properly.
```
jaelim-ms: I think this is caused by using NSUserDefaults instead of mocking itself. We've observed this issue before and I think we can fix it by mocking NSUserDefaults. Not a merge blocker and we can address the issue soon.

TroubleMakerBen: We're mocking NSUserDefaults elsewhere I think. Let's merge and address this in a sep. PR that also adds nullability where required.

clpolet: Yes will be part of another PR.

clpolet: That's right, added a new comment.

PR #465, opened 2017-04-06
Remove the filename auto-generation in error attachment

MatkovIvan:


MobileCenterCrashes/MobileCenterCrashes/Internals/MSErrorAttachmentLog.m
```diff
@@ -77,7 +75,7 @@ - (NSMutableDictionary *)serializeToDictionary {
   if (self.contentType) {
     dict[kMSContentType] = self.contentType;
   }
-  if (self.filename) {
+  if (self.filename && self.filename.length > 0) {
```
clpolet: We discussed with the team about allowing empty strings in the payload and we do allow empty string there. Could you please remove the `&& self.filename.length > 0`?

clpolet:
@MatkovIvan you can also remove the `+ (NSString *)generateFilenameForMimeType:(NSString *)mimeType` method from the MSCrashUtil class since it's not used anywhere else.

PR #466, opened 2017-04-06
Fixed issue with test methods in distribute module

bmourat:
Fixed issue with test methods in distribute module that were failing while running in batch mode.

### DIFF
--- a/MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
+++ b/MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
@@ -71,6 +71,10 @@ static NSURL *sfURL;
   [MSKeychainUtil clear];
   self.sut = [MSDistribute new];
   self.settingsMock = [MSMockUserDefaults new];
+  
+  // Mock MSUserDefaults shared method to return settingsMock.
+  id userDefaultsMock = OCMClassMock([MSUserDefaults class]);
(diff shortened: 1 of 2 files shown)


MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
```diff
@@ -71,6 +71,10 @@ - (void)setUp {
   [MSKeychainUtil clear];
   self.sut = [MSDistribute new];
   self.settingsMock = [MSMockUserDefaults new];
+  
+  // Mock MSUserDefaults shared method to return settingsMock.
+  id userDefaultsMock = OCMClassMock([MSUserDefaults class]);
```
clpolet: Do you think you can add this stub to MSMockUserDefaults init? That way we'll not face the issue on other classes that's using it.

bmourat: Moved mocking to MSMockUserDefaults. Also, i think a little bit of refactoring should be made to MSDistributeTests, cause there are couple of methods such as testOpenUrl, testNotDeleteUpdateoken, testDeleteUpdateTokenAfterReinstall, and probably some others, that implement their own mocking of MSUserDefaults. To avoid redundancy and inconsistent use, I think they should all use this mock. So, if you agree with that, i can do this refactoring :)

PR #467, opened 2017-04-07
Fix a few warnings

MatkovIvan:


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -47,11 +47,6 @@ static void ms_save_log_buffer_callback(__attribute__((unused)) siginfo_t *info,
                                         __attribute__((unused)) ucontext_t *uap,
                                         __attribute__((unused)) void *context) {
 
-  // Do not save the buffer if it is empty.
-  if (msCrashesLogBuffer.size() == 0) {
```
MatkovIvan: Check is always true, because size is constant.

PR #470, opened 2017-04-09
MSDistributeTests refactored

bmourat:
MSDistributeTests refactored to avoid redundancies and to unify interactions with MSUserDefaults and [NSBundle mainBundle].


MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
```diff
@@ -773,13 +760,10 @@ - (void)testWithoutNetwork {
 }
 
 - (void)testPackageHash {
-
```
jaelim-ms: Don't remove this line.

PR #474, opened 2017-04-12
Feature/update dialog

jaelim-ms:
This PR still requires an additional change for release notes URL.

### DIFF
--- a/MobileCenterDistribute/MobileCenterDistribute/Internals/Model/MSReleaseDetails.m
+++ b/MobileCenterDistribute/MobileCenterDistribute/Internals/Model/MSReleaseDetails.m
@@ -16,6 +16,7 @@ static NSString *const kMSUploadedAt = @"uploaded_at";
 static NSString *const kMSDownloadUrl = @"download_url";
 static NSString *const kMSAppIconUrl = @"app_icon_url";
 static NSString *const kMSInstallUrl = @"install_url";
+static NSString *const kMSReleaseNotesUrl = @"release_notes_url";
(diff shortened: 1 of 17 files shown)


MobileCenterDistribute/MobileCenterDistribute/Internals/Model/MSReleaseDetails.m
```diff
@@ -152,4 +153,35 @@ - (NSDictionary *)serializeToDictionary {
...
+          [self.provisioningProfileName isEqualToString:details.provisioningProfileName]) &&
+         ((!self.size && !details.size) || [self.size isEqualToNumber:details.size]) &&
+         ((!self.minOs && !details.minOs) || [self.minOs isEqualToString:details.minOs]) &&
+         (self.mandatoryUpdate == details.mandatoryUpdate) &&
+         ((!self.fingerprint && !details.fingerprint) || [self.fingerprint isEqualToString:details.fingerprint]) &&
+         ((!self.uploadedAt && !details.uploadedAt) || [self.uploadedAt isEqual:details.uploadedAt]) &&
+
+         // Don't compare downloadUrl. downloadUrl contains pltoken param which will have different values every time.
+         // ((!self.downloadUrl && !details.downloadUrl) || [self.downloadUrl isEqual:details.downloadUrl]) &&
+         ((!self.appIconUrl && !details.appIconUrl) || [self.appIconUrl isEqual:details.appIconUrl]) &&
```
jaelim-ms: The reason of the change is that download URL only contains pltoken param which will be changed every request. other URLs won't be changed like download URL since those don't contain something like request token.

PR #480, opened 2017-04-13
Conditionally start Distribute depending on config + update README

clpolet:


README.md
```diff
@@ -640,6 +640,30 @@ If you want the Mobile Center SDK to be disabled completely, use the `setEnabled
 ```swift
 MSMobileCenter.setEnabled(false)
 ```
+
+### Skip Distribute while developing
+ 
+Distribute does popup UI/browser at application start. While this is an expected behavior for your end users it could be disruptive for you during the development stage of your application. In this case you can simply skip Distribute from the SDK start:
```
jaelim-ms: Distribute service does popup...
...you can simply skip Distribute service...

clpolet:
@jaelim-ms I think for demos we should have this because it needs to reflect customer integration, not sure about puppet. I can remove it for Puppet.

README.md
```diff
@@ -640,6 +640,30 @@ If you want the Mobile Center SDK to be disabled completely, use the `setEnabled
 ```swift
 MSMobileCenter.setEnabled(false)
 ```
+
+### Skip Distribute while developing
+ 
+The Distribute service does popup UI/browser at application start. While this is an expected behavior for your end users it could be disruptive for you during the development stage of your application. In this case you can simply skip the Distribute service from the SDK start:
```
TroubleMakerBen: "popup a UI/browser".

PR #484, opened 2017-04-18
fix jump to nx page

TroubleMakerBen:
This makes the app crash again for release builds. Which, oddly, is what we want.

### DIFF
--- a/CrashLibIOS/CrashLibIOS/MSCrashNXPage.m
+++ b/CrashLibIOS/CrashLibIOS/MSCrashNXPage.m
@@ -2,8 +2,9 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  */
 
-#import "MSCrashNXPage.h"
 #import <sys/mman.h>
+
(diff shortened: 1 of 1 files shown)


CrashLibIOS/CrashLibIOS/MSCrashNXPage.m
```diff
@@ -3,6 +3,7 @@
  */
 
 #import "MSCrashNXPage.h"
+#import <sys/mman.h>
```
jaelim-ms: Reorder? `import <>` then `import ""`, right?

CrashLibIOS/CrashLibIOS/MSCrashNXPage.m
```diff
@@ -19,7 +20,17 @@ - (NSString *)desc {
 }
 
 - (void)crash {
-  ((void (*)(void)) NULL)();
+  /**
```
jaelim-ms: Line break.

PR #487, opened 2017-04-20
Checking release notes URL as well

jaelim-ms:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -559,7 +559,7 @@ - (void)showConfirmationAlert:(MSReleaseDetails *)details {
                                          }];
     }
 
-    if ([details.releaseNotes length] > 0) {
+    if ([details.releaseNotes length] > 0 && details.releaseNotesUrl) {
```
clpolet: Does the release note will still be part of the release details. If not then this part will be skipped. Shouldn't we remove the details.releaseNotes?

jaelim-ms:
Rerun CI build.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -539,10 +539,13 @@ - (void)showConfirmationAlert:(MSReleaseDetails *)details {
     NSString *messageFormat = details.mandatoryUpdate
                                   ? MSDistributeLocalizedString(@"MSDistributeAppUpdateAvailableMandatoryUpdateMessage")
                                   : MSDistributeLocalizedString(@"MSDistributeAppUpdateAvailableOptionalUpdateMessage");
+    NSString *appName = [MS_APP_MAIN_BUNDLE objectForInfoDictionaryKey:@"CFBundleDisplayName"];
+    if (!appName) {
+      appName = [MS_APP_MAIN_BUNDLE objectForInfoDictionaryKey:@"CFBundleName"];
+    }
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wformat-nonliteral"
```
TroubleMakerBen: This is more of a question. Why are we ignoring the warning here? Is there no way to make this work. My proposal would be that we add the following to our coding "rules": Never ignore a warning by using `clang diagnostic ignored` unless there is a good reason. State that reason next to the macro so we remember why we are ignoring the warning.

jaelim-ms: `NSString::stringWithFormat` recommended in Apple documentation if string resources have formatted string but `stringWithFormat` doesn't like a string variable for format string, it complains the parameter is non string literal.
If you have any recommendation, please let me know.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
jaelim-ms: Yes, it will be a part of release details. ReleaseDetails is a public facing data model, customers might want to get release notes.

PR #509, opened 2017-04-28
Fixed compiler errors in error-attachment-changes branch

bmourat:


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -909,7 +909,8 @@ - (BOOL)shouldProcessErrorReport:(MSErrorReport *)errorReport {
 }
 
 - (BOOL)delegateImplementsAttachmentCallback {
-  return self.delegate && [self.delegate respondsToSelector:@selector(attachmentsWithCrashes:forErrorReport:)];
+  id<MSCrashesDelegate> strongDelegate = self.delegate;
```
TroubleMakerBen: Shouldn't we check if the delegate is `!= NULL`?

jaelim-ms: What should be covered by below line at the first condition.

TroubleMakerBen: ok, i didn't check that one. Sorry!

PR #517, opened 2017-05-03
Don't allow update check in testing mode

erychagov:


clpolet:
@erychagov Did you find anything to ease the customer experience here? It means customers will have to inject the environment constant on each of their UI tests. I don't think there is an easy way without taking dependency on XCTest which we try to avoid. Just let me know of your findings.

TroubleMakerBen:
I share the same concern as @clpolet. We should make our customer's lives easier not more complicated. If there is no way of doing this in a non-intrusive way, I'd recommend solving this by documentation instead and tell folks to not use Distribution in case they run a test target.

erychagov:
https://github.com/Microsoft/mobile-center-sdk-ios/pull/521

clpolet:
Closing this one for now, priority changed. Though I think we'll keep the environment variable option because some people are using this but we'll also suggest the creation of another configuration dedicated to UI Tests then customers will simple `#if MY_UI_TEST_CONF` to not start Distribute in case of UI Tests. We'll probably expose an API in the future to soft disable Distribute, the way it's done while debugger is attached.

PR #518, opened 2017-05-03
OpenURL now returns BOOL + update apps + update tests

clpolet:
OpenURL should return a boolean to determine wether the URL is handled by Mobile Center Distribute or not.

### DIFF
--- a/MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
+++ b/MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
@@ -160,8 +160,8 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   [[self sharedInstance] setInstallUrl:installUrl];
 }
 
-+ (BOOL)openUrl:(NSURL *)url {
-  return [[self sharedInstance] openUrl:url];
++ (BOOL)openURL:(NSURL *)url {
+  return [[self sharedInstance] openURL:url];
 }
 
 + (void)notifyUpdateAction:(MSUpdateAction)action {
@@ -679,7 +679,7 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   exit(0);
 }
 
-- (BOOL)openUrl:(NSURL *)url {
+- (BOOL)openURL:(NSURL *)url {
   
   /*
    * Ignore if app secret not set, can't test scheme.
--- a/MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
+++ b/MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
@@ -160,8 +160,8 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   [[self sharedInstance] setInstallUrl:installUrl];
 }
 
-+ (void)openUrl:(NSURL *)url {
-  [[self sharedInstance] openUrl:url];
++ (BOOL)openUrl:(NSURL *)url {
+  return [[self sharedInstance] openUrl:url];
 }
 
 + (void)notifyUpdateAction:(MSUpdateAction)action {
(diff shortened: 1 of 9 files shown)


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -679,14 +679,19 @@ - (void)closeApp __attribute__((noreturn)) {
...
 }
 
-- (void)openUrl:(NSURL *)url {
+- (BOOL)openUrl:(NSURL *)url {
+  
+  /*
+   * Ignore if app secret not set, can't test scheme.
+   * Also ignore if request is not for Mobile Center Distribute and this app.
+   */
+  if (!self.appSecret || ![[url scheme] isEqualToString:[NSString stringWithFormat:kMSDefaultCustomSchemeFormat, self.appSecret]]) {
```
clpolet: I am a bit divided  here. We could also just check for "mobilecenter" at the beginning of the scheme but in this case it just validate whereas the scheme is for mobile center but not specifically for this app.
Any suggestions?

jaelim-ms: We already know the app secret so the SDK can determine the exact URL scheme for the app. In this case, we can also log to console if developers don't specify it properly so it would be better than just checking the prefix.

TroubleMakerBen: I agree that we need to check for more than the beginning as there could be more than 1 app on the device that has mobile center integrated.

Puppet/Puppet/AppDelegate.m
```diff
@@ -36,19 +36,23 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-/**
- *  This addition is required in case apps support iOS 8. Apps that are iOS 9 and later don't need to implement this
- * as our SDK uses SFSafariViewController for MSDistribute.
- */
+// Open URL for iOS 8.
 - (BOOL)application:(UIApplication *)application
               openURL:(NSURL *)url
     sourceApplication:(NSString *)sourceApplication
            annotation:(id)annotation {
+  NSLog(@"%@ Got waken up via openURL:sourceApplication:annotation: %@", kPUPLogTag, url);
```
TroubleMakerBen: `Got woken up` instead of `waken up`.

Puppet/Puppet/AppDelegate.m
```diff
@@ -36,19 +36,23 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
   // Forward the URL to MSDistribute.
-  [MSDistribute openUrl:url];
-  NSLog(@"%@ Got waken up via openURL: %@", kPUPLogTag, url);
-  return YES;
+  return [MSDistribute openURL:url];
+}
+
+// Open URL for iOS 9+.
+- (BOOL)application:(UIApplication *)application openURL:(nonnull NSURL *)url options:(nonnull NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options{
+  NSLog(@"%@ Got waken up via openURL:options: %@", kPUPLogTag, url);
```
TroubleMakerBen: `Was woken up` instead of `Got waken up`?

Sasquatch/SasquatchSwift/AppDelegate.swift
```diff
@@ -54,15 +54,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
     sasquatchController.mobileCenter = MobileCenterDelegateSwift()
   }
   
-  /**
-   *  This addition is required in case apps support iOS 8. Apps that are iOS 9 and later don't need to implement this
-   * as our SDK uses SFSafariViewController for MSDistribute.
-   */
+  // Open URL for iOS 8.
```
jaelim-ms: This is doc. Keep /** */

PR #521, opened 2017-05-04
Base idea for disabling updates using category in ui tests case

erychagov:
@clpolet @TroubleMakerBen @jaelim-ms 
Please look at this.
It's a base idea, not for merge.

In this case customer has to create once a category with a special method which should return bool value. But customer needs to switch the return value each time when he switches between ui tests and developing.

I wanted to setup environment variables for tests using project scheme however I came across with a problem that `[[NSProcessInfo processInfo] environment]` doesn’t contain added variables when we run ui tests.

Another variant:
https://github.com/Microsoft/mobile-center-sdk-ios/pull/517


clpolet:
@erychagov how does customer will decide on switching the return value? Do you mean having a dedicated scheme of the app for UI tests where the return value for this category will be YES? In that case it might be even simpler for the customer to not start the Distribute module based on the scheme or we should introduce a pre-proc macro that they add to that testing scheme then the SDK would test it to dis/allow updates.

clpolet:
Closing this one for now, priority changed. Though I think we'll keep the environment variable option because some people are using this but we'll also suggest the creation of another configuration dedicated to UI Tests then customers will simple `#if MY_UI_TEST_CONF` to not start Distribute in case of UI Tests. We'll probably expose an API in the future to soft disable Distribute, the way it's done while debugger is attached.

PR #523, opened 2017-05-04
Fix failing test on iOS 10

MatkovIvan:
Before this change `testOpenURLInSafariApp` fails on iOS 10 Simulator


MatkovIvan:
Hmm.. `MSIngestionSenderTests testTasksRunningOnSenderResumed` failed.
This is strange, because this PR modify only 1 test in other module.

PR #545, opened 2017-05-11
Always use base64 for attachment data

MatkovIvan:


jaelim-ms:
Sorry to say this, backend decided to bring content-type back due portal. Please revert that change.

PR #546, opened 2017-05-11
Custom properties validation implemented

bmourat:


MobileCenter/MobileCenter/MSCustomProperties.m
```diff
@@ -59,10 +58,32 @@ - (BOOL) isValidKey:(NSString *)key {
     MSLogError([MSMobileCenter logTag], @"Custom property \"%@\" must match \"%@\"", key, kKeyPattern);
     return NO;
   }
+  if (key.length > maxPropertyKeyLength) {
+    MSLogError([MSMobileCenter logTag], @"Custom property \"%@\" length cannot be longer than \"%d\" characters.", key, maxPropertyKeyLength);
+    return NO;
+  }
   if ([self.properties objectForKey:key]) {
     MSLogWarning([MSMobileCenter logTag], @"Custom property \"%@\" is already set or cleared and will be overridden.", key);
+  } else if ([self properties].count >= maxPropertiesCount) {
```
clpolet: Shouldn't it be ">" instead of ">=" ? Looks like we allow 60 props.

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
```diff
@@ -47,17 +47,31 @@ - (void)testKeyValidate {
   
   // Then
   assertThat([customProperties properties], hasCountOf(0));
-  
+
+  // Too long key.
+  // When
+  NSString *tooLongKey = [@"" stringByPaddingToLength:129 withString: @"a" startingAtIndex:0];
```
clpolet: I think in general we should also test that exact max values (60, 128, 128) pass the validation.

MobileCenter/MobileCenter/MSCustomProperties.m
jaelim-ms: It is up to 60 props.

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
jaelim-ms: I think this is testing over the length. There is a max length test below. 😄

MobileCenter/MobileCenter/MSCustomProperties.m
```diff
@@ -59,10 +58,32 @@ - (BOOL) isValidKey:(NSString *)key {
...
+    MSLogError([MSMobileCenter logTag], @"Custom properties cannot contain more than \"%d\" items.", maxPropertiesCount);
+    return NO;
   }
   return YES;
 }
 
+- (BOOL) isValidValue:(NSObject *)value {
+  if (value) {
+    if ([value isKindOfClass:[NSString class]]) {
```
clpolet: What about the other types? I am not worried about booleans ^^. But we should add numbers definitely.

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
bmourat: Yep, here they are:
https://github.com/Microsoft/mobile-center-sdk-ios/pull/546/files#diff-ecf1144fa64f59bf3d0a30278bcaab2cR65
https://github.com/Microsoft/mobile-center-sdk-ios/pull/546/files#diff-ecf1144fa64f59bf3d0a30278bcaab2cR151

MobileCenter/MobileCenter/MSCustomProperties.m
bmourat: Sure, what are limitations for numbers?

clpolet: You should turn it to string and validate it as a string because the limitation is on the number of characters.

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
clpolet: I see them now :). But I don't see the one for the exact max property count validation.

MobileCenter/MobileCenter/MSCustomProperties.m
clpolet: Ah ok my bad, the property is not yet in the dictionary yet.

MobileCenter/MobileCenter/MSCustomProperties.m
bmourat: Do we really need this one? The only numeric type that can handle 128 positions is double, and it is represented in exponential format, so there is no way a user can pass anything to setNumber method, that could result in string with length over 128.

jaelim-ms: I think we only limit the values for strings. If there is a number value, it is just as is. Am I missing something?

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
jaelim-ms: That's also here. 😃 
https://github.com/Microsoft/mobile-center-sdk-ios/pull/546/files#diff-ecf1144fa64f59bf3d0a30278bcaab2cR89

clpolet: Yes now it's there.

MobileCenter/MobileCenter/MSCustomProperties.m
clpolet: Weird to only validate string to me but if you can guaranty that we'll never fall into the limit for other types then I am fine ;) .

MobileCenter/MobileCenterTests/MSCustomPropertiesTests.m
jaelim-ms: Yeah.. My browser was not updating all comments but referring to the latest code.

PR #549, opened 2017-05-11
Fix/remove channel delegates from services

jaelim-ms:


MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,75 @@
...
+ * log.
  */
 - (void)onFailedPersistingLog:(id<MSLog>)log withInternalId:(NSString *)internalId;
 
+/**
+ * Callback method that will be called before each log will be send to the server.
+ *
+ * @param log The log to be sent.
+ */
+- (void)willSendLog:(id<MSLog>)log;
```
clpolet: Usually delegates should also reference the sender for instance `- (void)logManager:(id<MSLogManager>) willSendLog:(id<MSLog>)log;`. That's because this protocol doesn't know about who will implement it. Here we are making the assumption that only object already having a logManager object will implement this protocol and that may not always be true. (The other delegates before this one will have to be revised too be that's another story). @TroubleMakerBen and @jaelim-ms any thoughts?

jaelim-ms: This comment is what I expected. 😄  There are other methods without the sender instance already, I had to follow the exist style.
Based on our code base, logManager always generates warnings since we don't actually use it. Instead of adding `(void)logManager` everywhere, probably it is better to remove it from the method signatures.

clpolet: :blush: I would say we should at list prefix the method so that we know who performs the action. So that would be `- (void)logManagerWillSendLog:(id<MSLog>)log;` for instance. Maybe in another PR because as you said it was already the case before.

jaelim-ms: The change also requires other methods that are not part of the PR. We can address this one in a separate PR after whatever we decide.

PR #551, opened 2017-05-12
Fix timing issue for mac os apps

erychagov:
Fixed the issue which described here https://github.com/Microsoft/mobile-center-sdk-ios/pull/544#pullrequestreview-37631057


jaelim-ms:
Thanks for addressing this one and sorry to say that I missed one basic thing, the SDK is storing enabled state in NSUserDefaults whenever it is enabled or disabled. I need to think more about it that which implementation is better.

SasquatchMac/SasquatchMac/ServiceStateStore.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation;
+
+@objc class ServiceStateStore : NSObject {
+  private static let AnalyticsKey : String = "MSAnalytics";
```
jaelim-ms: Let's do this. Instead of using other keys, can you read NSUserDefaults for the keys that SDK is using?
kMSAnalyticsIsEnabledKey is an example of Analytics.

SasquatchMac/SasquatchMac/ViewControllers/AnalyticsViewController.swift
```diff
@@ -44,6 +46,7 @@ class AnalyticsViewController : NSViewController, MobileCenterProtocol {
     guard let `mobileCenter` = mobileCenter else {
       return
     }
+    ServiceStateStore.AnalyticsState = sender.state == 1
```
jaelim-ms: We don't need this part I think. below line will update the value to storage since we are referencing the same key.

SasquatchMac/SasquatchMac/ViewControllers/CrashesViewController.swift
```diff
@@ -29,6 +31,7 @@ class CrashesViewController : NSViewController, MobileCenterProtocol {
     guard let `mobileCenter` = mobileCenter else {
       return
     }
+    ServiceStateStore.CrashesState = sender.state == 1
```
jaelim-ms: We don't need this part I think. below line will update the value to storage since we are referencing the same key.

SasquatchMac/SasquatchMacObjC/AppDelegate.m
```diff
@@ -12,8 +12,8 @@ @interface AppDelegate ()
 @implementation AppDelegate
 
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
-
-  // Insert code here to initialize your application
+  [MSAnalytics setEnabled:[ServiceStateStore AnalyticsState]];
```
jaelim-ms: The SDK will read the value from storage in each service to set enable/disable for initial state. You can remove setEnabled call in AppDelegate.

SasquatchMac/SasquatchMac/ServiceStateStore.swift
```diff
@@ -0,0 +1,26 @@
...
+
+@objc class ServiceStateStore : NSObject {
+  private static let AnalyticsKey : String = "kMSAnalyticsIsEnabledKey";
+  private static let CrashesKey : String = "kMSCrashesIsEnabledKey";
+
+  class var AnalyticsState : Bool {
+    get {
+      return UserDefaults.standard.bool(forKey: AnalyticsKey);
+    }
+    set {
```
jaelim-ms: We shouldn't set any values to NSUserDefaults for the key that is used by the SDK. Get should be enough.

SasquatchMac/SasquatchMac/ServiceStateStore.swift
```diff
@@ -0,0 +1,26 @@
...
+      UserDefaults.standard.set(newValue, forKey: AnalyticsKey);
+      UserDefaults.standard.synchronize();
+    }
+  }
+
+  class var CrashesState : Bool {
+    get {
+      return UserDefaults.standard.bool(forKey: CrashesKey);
+    }
+    set {
```
jaelim-ms: We shouldn't set any values to NSUserDefaults for the key that is used by the SDK. Get should be enough.

jaelim-ms:
FYI, build will always fail because it doesn't have right schemes to run. This PR needs to wait until #561 is completed.

PR #553, opened 2017-05-12
Swizzle a selector in all super classes

erychagov:


MobileCenter/MobileCenter/Internals/Application Delegate/MSAppDelegateForwarder.m
```diff
@@ -116,7 +116,15 @@ + (IMP)swizzleOriginalSelector:(SEL)originalSelector withCustomSelector:(SEL)cus
   IMP customImp = class_getMethodImplementation(self, customSelector);
   IMP originalImp = NULL;
   BOOL methodAdded = NO;
-  
+
+  // Replace implementation in super classes
+  Class baseClass = originalClass.superclass;
+  if( baseClass != [NSObject class]) {
```
jaelim-ms: I agree the first comment that you need to check if it is implemented or not.
BTW for the second part, isn't NSObject always be the root class of other classes in Objective C? Just curios.

TroubleMakerBen: @jaelim-ms in theory: no, in practice: yes. Check out everything that Mike Ash writes about ObjC/Swift, e.g. his [blog on NSobject](https://www.mikeash.com/pyblog/friday-qa-2013-10-25-nsobject-the-class-and-the-protocol.html).

MobileCenter/MobileCenter/Internals/Application Delegate/MSAppDelegateForwarder.m
```diff
@@ -116,7 +116,15 @@ + (IMP)swizzleOriginalSelector:(SEL)originalSelector withCustomSelector:(SEL)cus
   IMP customImp = class_getMethodImplementation(self, customSelector);
   IMP originalImp = NULL;
   BOOL methodAdded = NO;
-  
+
+  // Replace implementation in super classes
+  Class baseClass = originalClass.superclass;
```
TroubleMakerBen: Can we be sure that originalClass actually responds to superclass? In practice: yes, in theory: no. So we should check that originalClass actually responds to `superclass`.

Or am I too paranoid @jaelim-ms @clpolet ?

clpolet: I don't think I understand, @TroubleMakerBen do you have examples in mind?

MobileCenter/MobileCenter/Internals/Application Delegate/MSAppDelegateForwarder.m
clpolet: Right NSObject is not always the root class but it's most probably is, though it doesn't hurt the way it is right now I think. To me it's maybe more an optimization.

clpolet:
We figured out that those scenario were already mostly covered by the current implementation. Will do another PR.

PR #556, opened 2017-05-16
Feature/macos project conf

jaelim-ms:
Making incremental changes now. There are a few known issues;

1. MobileCenterMacTests is broken due to having swift framework for OHHTTPStubs.
2. 7 unit tests are disabled in Core for macOS.
3. Treating warnings as errors flag turned off due to non-SDK code.


MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -321,7 +325,13 @@ - (void)testEmptyLogBufferFiles {
   NSString *filePath = [[self.sut.logBufferDir path]
       stringByAppendingPathComponent:[testName stringByAppendingString:@".mscrasheslogbuffer"]];
 
+#if TARGET_OS_IPHONE
   [someData writeToFile:filePath options:NSDataWritingFileProtectionNone error:nil];
+#else
+
+  // TODO: Make sure this is a right replacement.
+  [someData writeToFile:filePath atomically:YES];
```
jaelim-ms: @TroubleMakerBen macOS doesn't take the option. Is this enough?

MobileCenter.xcworkspace/xcshareddata/xcschemes/AllLibraries for macOS.xcscheme
```diff
@@ -48,6 +48,48 @@
                ReferencedContainer = "container:MobileCenterCrashes/MobileCenterCrashes.xcodeproj">
```
clpolet: I think schemes should all following the same naming conventions, for instance this one should be `AllLibrariesMacOS` and the iOS one would be `AllLibrariesIOS`. I think there's a couple of other schemes like that.

jaelim-ms: Let's take scheme things separately as action items. We need to hide/rename some of them so let's discuss this offline.

clpolet: All right.

PR #560, opened 2017-05-17
Flush swizzling traces when first set the log level

clpolet:


clpolet:
@jaelim-ms I totally agree with you, this is a workaround to me. That's also why I added this TODO in code:
```
 /**
  * Flush debugging traces accumulated until now.
  * TODO: We should find a way for customers to set the log level in their configuration somehow so that it'll be set at
  * the time of the swizzling. This will allow having swizzling traces in real time, in that case we can remove the whole
  * trace buffer mechanism.
  */
```

PR #561, opened 2017-05-17
Feature/macos

jaelim-ms:


jaelim-ms:
Bitrise workflow also requires to be updated with new scheme name. Rebuild

Global.xcconfig
```diff
@@ -12,6 +10,9 @@ MS_SIM_ARCHS = x86_64 i386
 ARCHS = $(MS_ARM_ARCHS)
```
clpolet: We should define architectures directly in an xconfig file dedicated to all iOS targets and do the same for macOS. This will prevent having mac config variables showing up in iOS projects.

clpolet:
Just a note on the naming/organizing:
* Maybe to rename the iOS sasquatch and puppet apps to `puppetIOS` and `sasquatchIOSObjC/Swift`.
* We should regroup schemes by platforms
* Maybe prefix the scheme with the platform `iOS Push Library`, `iOS Push Framework`, `iOS Push Tests`... and do the same for targets.

jaelim-ms:
I addressed @clpolet's feedback. As discussed offline, I created a task for the follow up.

PR #563, opened 2017-05-18
Fix/completion handler not called swizzling

clpolet:
Fix didReceiveRemoteNotification completion handler not called while no more delegate in swizzling.

* Forward a completion handler block that collects `UIBackgroundFetchResult` values.
* Call the original completion handler in the end with the forwarded `UIBackgroundFetchResult` value.


clpolet:
I'll add unit tests tomorrow.

MobileCenter/MobileCenter/Internals/Application Delegate/MSAppDelegateForwarder.m
```diff
@@ -286,19 +287,28 @@ - (void)custom_application:(UIApplication *)app
     didReceiveRemoteNotification:(NSDictionary *)userInfo
           fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
 
+  // Collect `UIBackgroundFetchResult` from delegates in order to call the original completion handler later.
+  __block UIBackgroundFetchResult forwardedFetchResult = UIBackgroundFetchResultNoData;
+  void (^forwardedCompletionHandler)(UIBackgroundFetchResult) = ^(UIBackgroundFetchResult fetchResult) {
+    forwardedFetchResult = fetchResult;
```
jaelim-ms: Looks it overwrites whatever the previous fetch results are. In other words, if there are multiple methods and it returned different value, it will be replaced by the last return value. Is this possible?

clpolet: Yes exactly, that's why I added the FIXME just below :). Right now there only one delegate implementing this selector so it's fine but we will have to fix it yes.

PR #564, opened 2017-05-18
Add page with Mobile Center settings

erychagov:
I added fields with install id, app secret, log url and checkbox for enabling/disabling Mobile Center.


jaelim-ms:
Rebuild.

jaelim-ms:
This is merging into working branch which doesn't have workflow set up in Bitrise. Mark as don't merge until feature/macos branch is merged into multiple-platforms. Then we'll change the base branch and move forward.

PR #572, opened 2017-05-26
Implemented MSCrashes delegate to mac OS.

erychagov:
Using NSAlert we can't setup button for special action like default or cancel. These properties depend on order which we use when adding button. First button is default button, second button is no button, third button is cancel button.

If we display two alerts (or more) then last alert replaces previous one, but after hiding last alert previous alert showes automatically. It's default behaviour. We can change it if needed.

tvOS uses class UIAlertController as is iOS.


erychagov:
> BTW, can you write objective c version?
Sure :)

PR #573, opened 2017-05-26
Ignore warnings related to consuming PLCrashReporter for macOS.

TroubleMakerBen:
Some explanation on this:

None of the warnings are actually related to PLCrashReporter doing something different. Enabling `-Weverything` is what causes them to show up.
The way to "fix" them without **major** changes to PLCrashReporter is to put `#pragma clang diagnostic push' and `... ignore` everywhere the warnings show up.
I have a local branch for PLCrashReporter that I can push if you like that approach better. I don't really see the reason for this, so I added the ignore-flags to our `Global.xcconfig` and we're good. I also tried to add an explanation for the various flags. Let me know if you have questions.


jaelim-ms:
Let's get this in for now and eliminate all warnings later.

PR #575, opened 2017-06-01
Print warning to log if there more than 2 attachments per crash report

erychagov:


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -188,6 +190,9 @@ + (void)notifyWithUserConfirmation:(MSUserConfirmation)userConfirmation {
     // Get error attachments.
     if ([crashes delegateImplementsAttachmentCallback]) {
       attachments = [crashes.delegate attachmentsWithCrashes:crashes forErrorReport:report];
+      if (attachments.count > kMaxAttachmentsPerCrashReport) {
```
jaelim-ms: This logic should be done within below for loop. We are currently missing validation of attachments, we should call isValid for each attachment before sending. You did this for Android so you should understand. 😄

jaelim-ms: Call [attachement isValid] before [line 206](https://github.com/Microsoft/mobile-center-sdk-ios/pull/575/files#diff-434ca6481b9819743235b3b829b57787R206), only process attachment log if it is valid. Otherwise, please log the error.
This is the error message in Android.
`MobileCenterLog.error(LOG_TAG, "Not all required fields are present in ErrorAttachmentLog.");`

jaelim-ms:
As the feedback I gave you for Android, please add unit test for this one here.

PR #589, opened 2017-06-06
Feature/configurations

jaelim-ms:


Global Documentation.xcconfig
```diff
@@ -0,0 +1 @@
+USER_HEADER_SEARCH_PATHS = "$(SRCROOT)/../MobileCenter/MobileCenter"
```
clpolet: We can probably remove this config file.

jaelim-ms: Removed!

PR #609, opened 2017-06-13
Feature/xcconfig scheme naming

jaelim-ms:


jaelim-ms:
The build would be failed because I didn't change bitrise configuration due to scheme change. Let's do merge this first and then I'll check build result after changing bitrise workflow.

jaelim-ms:
Missed one configuration in Bitrise, running the build again.

PR #611, opened 2017-06-13
Fix a crash when response payload is null

jaelim-ms:


TroubleMakerBen:
Just one question: did you change the provisioning profile on purpose?

jaelim-ms:
Yeah.. I refreshed the provisioning profiles in developer account as well as bitrise.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -267,92 +267,100 @@ - (void)checkLatestRelease:(NSString *)updateToken releaseHash:(NSString *)relea
...
+
+             // Error instance for JSON parsing.
+             NSError *jsonError = nil;
+
+             // Success.
+             if (statusCode == MSHTTPCodesNo200OK) {
+               MSReleaseDetails *details = nil;
+               if (data) {
+                 id dictionary =
+                     [NSJSONSerialization JSONObjectWithData:data options:NSJSONReadingMutableContainers error:nil];
```
clpolet: Looks like this can produce an error, we should handle it.

clpolet: This will be part of another task as discussed off line.

TroubleMakerBen:
Let's merge this to develop. I need to update the changelog.

PR #622, opened 2017-06-15
Remove channel dependency from services

jaelim-ms:
MSAnalyticsDelegate has not been exposed so I don't need to hide the delegate.
Still we need to decouple the dependency of channel from services so I remove this dependency and all communication will go through log manager delegate.

This is re-implementation of #549 and reverting #566.

### DIFF
--- a/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
+++ b/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
@@ -6,7 +6,6 @@
 #import "MSLogManagerDefaultPrivate.h"
 #import "MSMobileCenterErrors.h"
 #import "MSMobileCenterInternal.h"
-#import "MobileCenter+Internal.h"
 
 static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecenter.LogManagerQueue";
 
@@ -14,11 +13,6 @@ static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecente
  * Private declaration of the log manager.
  */
 @interface MSLogManagerDefault ()
-
-/**
- * A boolean value set to YES if this instance is enabled or NO otherwise.
- */
-@property BOOL enabled;
 
 @end
 
@@ -72,28 +66,6 @@ static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecente
   }
 }
 
-#pragma mark - Channel Delegate
-
-- (void)addChannelDelegate:(id<MSChannelDelegate>)channelDelegate forGroupId:(NSString *)groupId {
-  if (channelDelegate) {
-    if (self.channels[groupId]) {
-      [self.channels[groupId] addDelegate:channelDelegate];
-    } else {
-      MSLogWarning([MSMobileCenter logTag], @"Channel has not been initialized for the group ID: %@", groupId);
-    }
-  }
-}
-
-- (void)removeChannelDelegate:(id<MSChannelDelegate>)channelDelegate forGroupId:(NSString *)groupId {
-  if (channelDelegate) {
-    if (self.channels[groupId]) {
-      [self.channels[groupId] removeDelegate:channelDelegate];
-    } else {
-      MSLogWarning([MSMobileCenter logTag], @"Channel has not been initialized for the group ID: %@", groupId);
-    }
-  }
-}
-
 - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLogManagerDelegate> delegate))block {
   @synchronized(self) {
     for (id<MSLogManagerDelegate> delegate in self.delegates) {
@@ -102,6 +74,35 @@ static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecente
       }
     }
   }
+}
+
+#pragma mark - Channel Delegate
+
+- (void)channel:(id<MSChannel>)channel willSendLog:(id<MSLog>)log {
+  [self enumerateDelegatesForSelector:@selector(willSendLog:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate willSendLog:log];
+                              }
+                            }];
+}
+
+- (void)channel:(id<MSChannel>)channel didSucceedSendingLog:(id<MSLog>)log {
+  [self enumerateDelegatesForSelector:@selector(didSucceedSendingLog:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate didSucceedSendingLog:log];
+                              }
+                            }];
+}
+
+- (void)channel:(id<MSChannel>)channel didFailSendingLog:(id<MSLog>)log withError:(NSError *)error {
+  [self enumerateDelegatesForSelector:@selector(didFailSendingLog:withError:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate didFailSendingLog:log withError:error];
+                              }
+                            }];
 }
 
 #pragma mark - Process items
--- a/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
+++ b/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
@@ -86,11 +86,9 @@ static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecente
 
                               /*
                                * If the delegate doesn't have groupId implementation, it assumes that the delegate is
-                               * interested in all kinds of logs.
+                               * interested in all kinds of logs. Otherwise, compare groupId.
                                */
                               if (![delegate respondsToSelector:@selector(groupId)] ||
-
-                                  // If the delegate has groupId implementation, compare groupId.
                                   [[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
                                 [delegate willSendLog:log];
                               }
@@ -103,11 +101,9 @@ static char *const MSlogsDispatchQueue = "com.microsoft.azure.mobile.mobilecente
 
                               /*
                                * If the delegate doesn't have groupId implementation, it assumes that the delegate is
-                               * interested in all kinds of logs.
+                               * interested in all kinds of logs. Otherwise, compare groupId.
                                */
                               if (![delegate respondsToSelector:@selector(groupId)] ||
-
-                                  // If the delegate has groupId implementation, compare groupId.
                                   [[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
                                 [delegate didSucceedSendingLog:log];
                               }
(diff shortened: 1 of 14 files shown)


jaelim-ms:
@guperrot @achocron I need to test Xamarin UI test with this change. Can you give me an instruction if you have?

guperrot:
@jaelim-ms Need to publish-internal this branch so you can pull it in Xamarin (modify version in build.cake).
Then update bindings project.
Then test the UI test project locally with a real iPhone/iPad.

guperrot:
@jaelim-ms Forms puppet also has the analytics delegate so you can look at that app first, easier than UI test projects.

jaelim-ms:
I tested on Xamarin and confirmed that it works.
I'll add unit tests for this implementation soon.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,78 @@
 #import <Foundation/Foundation.h>
+
 #import "MSConstants+Internal.h"
+#import "MSLog.h"
 
+// TODO: We need to pass a sender instance in delegate methods or changing method name with a prefix of sender.
```
clpolet: Agree, we will have to do some renaming at least.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,78 @@
...
+/**
+ * Callback method that will be called in case the SDK was unable to send a log.
+ *
+ * @param log The log to be sent.
+ * @param error The error that occured.
+ */
+- (void)didFailSendingLog:(id<MSLog>)log withError:(NSError *)error;
+
+/**
+ * Get service unique key for storage purpose.
```
clpolet: 'delegate' instead of 'service'

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,78 @@
...
+ * @discussion We had some discussion about the naming of the method. To match the
+ * onEnqueueingLog:withInternalId (@see onEnqueueingLog:withInternalId) callback, it should
+ * also use `enqueuing` in it's signature, yet, as of now, it indicates successful persistence of a log. This
+ * method's name might change in the future to make a distinction between a successfully enqueued log and a persisted
+ * log.
  */
 - (void)onFailedPersistingLog:(id<MSLog>)log withInternalId:(NSString *)internalId;
 
+/**
+ * Callback method that will be called before each log will be send to the server.
```
clpolet: 'sent' instead of 'send' :)

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,78 @@
...
+ *
+ * @param log The log to be sent.
+ * @param error The error that occured.
+ */
+- (void)didFailSendingLog:(id<MSLog>)log withError:(NSError *)error;
+
+/**
+ * Get service unique key for storage purpose.
+ *
+ * @return A group ID of the service.
```
clpolet: goup 'Id'

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -104,6 +78,35 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
   }
 }
 
+#pragma mark - Channel Delegate
+
+- (void)channel:(id<MSChannel>)channel willSendLog:(id<MSLog>)log {
+  [self enumerateDelegatesForSelector:@selector(willSendLog:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate willSendLog:log];
```
clpolet: This methods are optional so I would say you need to check if the delegate responds to the selector first.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -104,6 +78,35 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
...
+                              }
+                            }];
+}
+
+- (void)channel:(id<MSChannel>)channel didSucceedSendingLog:(id<MSLog>)log {
+  [self enumerateDelegatesForSelector:@selector(didSucceedSendingLog:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate didSucceedSendingLog:log];
```
clpolet: This methods are optional so I would say you need to check if the delegate responds to the selector first.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -104,6 +78,35 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
...
+                              }
+                            }];
+}
+
+- (void)channel:(id<MSChannel>)channel didFailSendingLog:(id<MSLog>)log withError:(NSError *)error {
+  [self enumerateDelegatesForSelector:@selector(didFailSendingLog:withError:)
+                            withBlock:^(id<MSLogManagerDelegate> delegate) {
+                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+                                [delegate didFailSendingLog:log withError:error];
```
clpolet: This methods are optional so I would say you need to check if the delegate responds to the selector first.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -1,46 +1,78 @@
...
+ */
+- (NSString *)groupId;
```
clpolet: This one can be optional too as it is now, if not implemented then it would mean that the delegate is interested in all logs from all group Ids. If so then you'll have to do the selector check before calling it in the logManager.
But if it's not intended to be optional then it should be an explicit required field.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
jaelim-ms: This is explaining about group ID. Group ID is a unique key of storage for each service.
Why should it be a delegate unique key?

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
jaelim-ms: Oops.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
jaelim-ms: Good point!

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
jaelim-ms: It is checked in `enumerateDelegatesForSelector`.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
jaelim-ms: It is checked in `enumerateDelegatesForSelector`.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
jaelim-ms: It is checked in `enumerateDelegatesForSelector`.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
jaelim-ms: I think we will need a discussion about this one. 😄 
There is no rule for ID vs Id and we already have both of them in a bunch of places.
I'll leave this as it and would like to touch all the occurrences together.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
clpolet: Because this class is about delegates not services so it feels weird to have service mentioned here. For instance when we refer to group Id in log manager or a channels we don't mention service because it's abstracted, we don't need to know who is using it, we just need to know the purpose of it.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
clpolet: I think we already had the discussion about ID vs Id and we chose Id. Or did I just invented it :sweat_smile:? I remember there was even a task to rename everything.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
clpolet: Ah right, the funny part is I introduced it :stuck_out_tongue_closed_eyes:.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
clpolet: Ah right, the funny part is I introduced it 😝.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
clpolet: Ah right, the funny part is I introduced it 😝.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
clpolet: We can always redefine the rules if we want.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
clpolet: We can always redefine the rules if we want.

jaelim-ms: I'm not aware of the discussion for ID vs Id. Probably it was before I started contributing iOS? Anyway, let's revisit there! I'll create a task for that.

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
clpolet: Works for me ;)

MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -83,7 +83,15 @@ - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSLog
...
                             withBlock:^(id<MSLogManagerDelegate> delegate) {
-                              if ([[delegate groupId] isEqualToString:[channel.configuration groupId]]) {
+
+                              /*
+                               * If the delegate doesn't have groupId implementation, it assumes that the delegate is
+                               * interested in all kinds of logs.
+                               */
+                              if (![delegate respondsToSelector:@selector(groupId)] ||
+
+                                  // If the delegate has groupId implementation, compare groupId.
```
clpolet: It's kind of hard to read having the 2 conditions from the same `if` separated by a new line and a comment. Could you merge this comment with the comment just above ?

jaelim-ms: Done.

PR #623, opened 2017-06-16
Feature/add batch id column to db

TroubleMakerBen:


MobileCenter/MobileCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -86,8 +85,18 @@ - (BOOL)loadLogsWithGroupId:(NSString *)groupId
   logsAvailable = logs.count > 0;
   if (logsAvailable) {
     batchId = MS_UUID_STRING;
-    [self.batches setObject:(NSArray<NSString *> * _Nonnull)[logs allKeys]
-                     forKey:[groupId stringByAppendingString:batchId]];
+    
+    // Update the logs in the DB with the batchId
+    NSString *logIdsForBatch = [[logs allKeys] componentsJoinedByString:@"','"];
+    
+    NSString *updateLogsQuery = [NSString stringWithFormat:@"UPDATE %@ SET %@ = '%@' WHERE %@ IN ('%@')", kMSLogTableName, kMSBatchIdColumnName, batchId, kMSIdColumnName, logIdsForBatch];
```
jaelim-ms: Don't we need retries here?
What will happen if update fails? Current implementation will resend logs.

TroubleMakerBen:
As discussed offline, we won't do this change as it comes with a few downsided:
- Android does not use this approach, so we should try to remain consistent.
- it's very hard to unit test having the batchID in the DB, the approach with having a dictionary of logIDs can be easily unit tested
- Multithreading will be an issue when moving the batchID to the DB.

PR #625, opened 2017-06-19
Fix/fix error attachment validation bump version

TroubleMakerBen:


MobileCenterCrashes/MobileCenterCrashes/Model/MSErrorAttachmentLog.m
```diff
@@ -84,7 +84,7 @@ - (BOOL)isEqual:(id)object {
 }
 
 - (BOOL)isValid {
-  return [super isValid] && self.errorId && self.attachmentId && self.data && self.contentType;
+  return self.errorId && self.attachmentId && self.data && self.contentType;
```
jaelim-ms: We should keep `[super isValid]`. Let's validate available attachment properties at that time.

PR #636, opened 2017-06-21
Anonymize user path in Crashes logs

jaelim-ms:


jaelim-ms:
The CI build is broken in this branch. There is a PR with the fix and it should be wait until that goes in.

PR #637, opened 2017-06-21
Change to use Mach exception handler by default

jaelim-ms:
There will be a separate doc PR in doc repo.


jaelim-ms:
Good point @TroubleMakerBen. I also updated the version.

PR #640, opened 2017-06-26
Added way to customize properties sent with trackEvent/trackPage methods in Mac demo

erychagov:


erychagov:
@clpolet 
Fixed

TroubleMakerBen:
Tested, works now.

PR #641, opened 2017-06-26
Feature/merge develop multiple platforms

jaelim-ms:


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.h
```diff
@@ -95,37 +95,23 @@ typedef NS_ENUM(NSUInteger, MSUserConfirmation) {
...
- *
- *  Default: _NO_
- *
- * @warning The Mach exception handler executes in-process, and will interfere with debuggers when
- *  they attempt to suspend all active threads (which will include the Mach exception handler).
- *  Mach-based handling should _NOT_ be used when a debugger is attached. The SDK will not
- *  enabled catching exceptions if the app is started with the debugger running. If you attach
- *  the debugger during runtime, this may cause issues if the Mach exception handler is enabled!
- * @see isDebuggerAttached
+ * Default: _YES_
```
TroubleMakerBen: There is no "default" anymore, I think. This was wrong before the PR, I think.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.h
```diff
@@ -95,37 +95,23 @@ typedef NS_ENUM(NSUInteger, MSUserConfirmation) {
 ///-----------------------------------------------------------------------------
 
 /**
- *  Trap fatal signals via a Mach exception server.
+ * Disable to trap fatal signals via a Mach exception server.
```
TroubleMakerBen: This is ambiguous.
How about "Disable the Mach exception handler. This can be useful when debugging." here and then add some more context. Further down.

```
The Mach exception handler executes in-process, and might interfere with debuggers when they attempt to suspend all active threads (which will include the Mach exception handler). Mach-based handling should _NOT_ be used when a debugger is attached. If you attach a debugger during runtime, this may cause issues if the Mach exception handler is enabled.

```

jaelim-ms: This is a merge PR and your feedback needs to be addressed separately. #642 will do that. Thanks!

PR #644, opened 2017-06-26
Feature/add tv os crashes

jaelim-ms:
In this PR
1. PLCrashReporter for tvOS has been added to Mobile Center.
2. MobileCenterCrashes for tvOS has been added to Mobile Center.
3. SasquatchTV application now imports MobileCenterCrashes.
4. The test application has not been verified. This will be a separate task.


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -242,7 +244,9 @@ - (instancetype)init {
     _logBufferDir = [MSCrashesUtil logBufferDir];
     _analyzerInProgressFile = [_crashesDir URLByAppendingPathComponent:kMSAnalyzerFilename];
     _didCrashInLastSession = NO;
+#if !TARGET_OS_TV
     _enableMachExceptionHandler = YES;
+#endif
```
clpolet: I would suggest
```
#if TARGET_OS_TV
     _enableMachExceptionHandler = NO;
#else
     _enableMachExceptionHandler = YES;
#endif
```
So you don't need the macro line 525.

MobileCenter.xcworkspace/xcshareddata/xcschemes/AllDoc+FrameworksTVOS.xcscheme
```diff
@@ -47,9 +61,9 @@
       <MacroExpansion>
          <BuildableReference
             BuildableIdentifier = "primary"
-            BlueprintIdentifier = "043121491EE0BFC7007054C5"
-            BuildableName = "MobileCenterTVOSFramework"
```
clpolet: Shouldn't this scheme also generate the framework?

jaelim-ms: This is a doc framework and it should generate tvOS framework as well.

clpolet: Right :)

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
jaelim-ms: That was my first implementation but we cannot eliminate the macro at line 525.
`PLCrashReporterSignalHandlerTypeMach` is not available on tvOS SDK so anyway we have to do this one.

jaelim-ms:
TARGET_OS_TV is set to 1 on tvOS SDK for device and simulator. Must be like this.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
clpolet: OK

PR #646, opened 2017-06-27
Fix sqlite dependencies

clpolet:
Sasquatch apps no longer directly linked to sqlite lib.


MobileCenter/MobileCenter/Support/module.modulemap
```diff
@@ -8,4 +8,5 @@ framework module MobileCenter {
   link framework "CoreTelephony"
   link framework "SystemConfiguration"
   link framework "UIKit"
+  link "sqlite3"
```
jaelim-ms: Don't you need to update podspec as well? Keep in mind that you also need to this in multiple-platforms branch.

clpolet: Yep I created another task for this specific one.

PR #647, opened 2017-06-27
Feature/tvos unit tests

jaelim-ms:
1. MobileCenter unit test doesn’t work.
- MobileCenter unit test is excluded from AllLibraryTVOS build
- There are bunch of file system access failures which are quite expected. TV OS only has cache folder that can be accessed from application. Find FIXME comments in tests.
- AppDelegate code for swizzling is included in tvOS build but openURL related method is not available on tvOS. MSAppDelegateForwarderTests.m file has some disabled unit tests.
- OHHTTPStubs doesn’t officially provide framework for tvOS. We need to manually build it and I put that in under Vendor folder. Seems this one doesn’t work.
- Because of Swift module version, macOS and tvOS tests require to use OHHTTPStubs version 5.2.3-swift3. Keep this in mind.

2. Crashes has disabled unit tests due to file access. Find MSCrashesTests.mm and MSWrapperExceptionManagerTests.m. There are FIXME comments for that.


jaelim-ms:
Closing this PR, #710 has been merged with all the implementation that was done in this PR.

PR #649, opened 2017-06-28
Workaroud for iOS 10 push notifications bug

MatkovIvan:
See https://forums.developer.apple.com/thread/54332

### DIFF
--- a/MobileCenterPush/MobileCenterPush/Internal/MSPushAppDelegate.m
+++ b/MobileCenterPush/MobileCenterPush/Internal/MSPushAppDelegate.m
@@ -17,13 +17,10 @@
 }
 
 // Workaroud for iOS 10 bug. See https://forums.developer.apple.com/thread/54332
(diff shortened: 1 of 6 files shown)


MobileCenterPush/MobileCenterPush/Internal/MSPushAppDelegate.m
```diff
@@ -16,6 +16,16 @@ - (void)application:(__attribute__((unused))UIApplication *)application
   [MSPush didFailToRegisterForRemoteNotificationsWithError:error];
 }
 
+// Workaroud for iOS 10 bug. See https://forums.developer.apple.com/thread/54332
+- (void)application:(UIApplication *)application
+    didReceiveRemoteNotification:(NSDictionary *)userInfo
+{
+  [self application:application
```
clpolet: Here you can simply do: `[MSPush didReceiveRemoteNotification:userInfo]`.

PR #669, opened 2017-07-12
Remove SFSafariViewController

bmourat:


TroubleMakerBen:
Holding off the merge, as the latest iOS beta brought a new API, we might be able to utilize SFAuthenticationSession for the feature. We need to do some investigation and decide if we want to have, essentially, 3 different implementations for Distribute.

TroubleMakerBen:
We're using SFAuthenticationSession for iOS 11, use SFSafariViewController for iOS 10 and open Safari for older iOS versions.

PR #679, opened 2017-07-26
Fix log limits in database

jaelim-ms:


jaelim-ms:
OK. However if we have that per groupId, the size of database will become over 3 ~ 4MB.

PR #686, opened 2017-08-01
Fix sending start_services log with empty array

MatkovIvan:


MobileCenter/MobileCenter/MSMobileCenter.m
```diff
@@ -211,7 +211,9 @@ - (void)start:(NSString *)appSecret withServices:(NSArray<Class> *)services {
           [servicesNames addObject:[service serviceName]];
         }
       }
-      [self sendStartServiceLog:servicesNames];
+      if ([servicesNames count] > 0) {
+        [self sendStartServiceLog:servicesNames];
+      }
```
jaelim-ms: Can we add a debug log here if no services have been started?

MobileCenter/MobileCenter/MSMobileCenter.m
```diff
@@ -202,7 +202,7 @@ - (BOOL)configure:(NSString *)appSecret {
 - (void)start:(NSString *)appSecret withServices:(NSArray<Class> *)services {
   @synchronized(self) {
     BOOL configured = [self configure:appSecret];
-    if (configured) {
+    if (configured && services) {
```
jaelim-ms: This is not the right fix for the issue because the method is not supposed to be used for empty service list.
However, I like this fix better and just working as `configure` is harmless.

MobileCenter/MobileCenterTests/MSMobileCenterTests.m
```diff
@@ -186,6 +186,23 @@ - (void)testStartServiceWithInvalidValues {
...
 
+- (void)testStartWithoutServices {
+  
+  // If
+  id logManager = OCMClassMock([MSLogManagerDefault class]);
+  OCMStub([logManager alloc]).andReturn(logManager);
+  OCMStub([logManager initWithAppSecret:[OCMArg any] installId:[OCMArg any] logUrl:[OCMArg any]]).andReturn(logManager);
+  
+  // Not allow processLog
+  OCMReject([logManager processLog:[OCMArg isKindOfClass:[MSStartServiceLog class]] forGroupId:[OCMArg any]]);
```
jaelim-ms: I wonder if you forgot to add OCMVerify or something else to verify this OCMReject. I'm not familiar with this thing. 😄

MatkovIvan: Actually it no need here, but I added `OCMVerifyAll` to be sure.

PR #689, opened 2017-08-02
Fix OCMReject usage

MatkovIvan:


MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
```diff
@@ -724,18 +740,12 @@ - (void)testDontShowConfirmationAlertIfNoMandatoryReleaseWhileNoNetwork {
     [expection fulfill];
   });
   [self waitForExpectationsWithTimeout:1
-                               handler:^(__attribute__((unused)) NSError *error) {
-
-                                 /*
-                                  * Then
-                                  */
-                                 OCMReject(
```
jaelim-ms: Any reasons of removing OCMReject here?

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
```diff
@@ -783,21 +796,22 @@ - (void)testDontPersistLastestReleaseIfNotMandatory {
...
-  [self mockMSPackageHash];
+  id utilityMock = [self mockMSPackageHash];
   
   // When
   NSURL *url = [NSURL URLWithString:[NSString stringWithFormat:@"%@://?", scheme]];
   BOOL result = [MSDistribute openURL:url];
   
   // Then
   assertThatBool(result, isFalse());
-  OCMReject([distributeMock checkLatestRelease:[OCMArg any] releaseHash:[OCMArg any]]);
```
jaelim-ms: There are 6 conditions that `checkLatestRelease:releaseHash:` is not being called and this test is verifying that. Can you tell me why you remove all OCMReject here?

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
jaelim-ms: Oh.. I didn't see that you added that at the beginning of the method.

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
jaelim-ms: Oh.. I didn't see that you added that at the beginning of the method.

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeTests.m
MatkovIvan: `OCMReject` needs to call _before_ testing method.
Also verify call needed when:
> In fail-fast mode an exception might not cause the test to fail. This can happen when the call stack for the method does not end in the test. Fail fast exceptions will be re-thrown when `OCMVerifyAll` is called.

PR #690, opened 2017-08-02
Fix/ios bug fix

jaelim-ms:


MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -248,7 +248,7 @@ - (instancetype)init {
                                                                    priority:MSPriorityHigh
                                                               flushInterval:1.0
                                                              batchSizeLimit:10
-                                                        pendingBatchesLimit:6];
+                                                        pendingBatchesLimit:1];
```
TroubleMakerBen: How did a batch limit of 6 even end up here?!?! 🤔

jaelim-ms: We forgot to change this to 1 during the database merge. This was a workaround to keep more logs and now we have to change this back to the original number since it can cause an issue if error_attachment log is too big.

PR #693, opened 2017-08-04
Feature/merge develop to multiple platforms

jaelim-ms:


MobileCenter/MobileCenter/Internals/Util/MSUtility+File.h
```diff
@@ -33,4 +33,16 @@ extern NSString *MSUtilityFileCategory;
  */
 + (BOOL)removeItemAtURL:(NSURL *)itemURL;
 
+/**
+* Creates a directory at the given location, intermediate directories are also created if nonexistant.
```
TroubleMakerBen: It should be `nonexistent` but let's fix this in another PR

Sasquatch/Sasquatch.xcodeproj/project.pbxproj
```diff
@@ -1326,6 +1336,7 @@
 		D31EF9741E9517C600450162 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ARCHS = "$(ARCHS_STANDARD)";
```
TroubleMakerBen: I think we should remove the `ARCHS = "$(ARCHS_STANDARD)";` lines.

jaelim-ms: As discussed offline, we will keep this one and add this line to the other two watch targets.

PR #694, opened 2017-08-07
Detecting environment for macOS

bmourat:


jaelim-ms:
Looks good to me but we need to update OSS since this implementation contains 3rd party implementation.

TroubleMakerBen:
@bmourat please add the copyright information to the 3rd party license document.

TroubleMakerBen:
We need to register the component, I'm creating a task for that.

TroubleMakerBen:
We've discussed this issue. I don't think we need to detect a MAS environment (HA didn't). And the detection mechanism in this PR doesn't look like it's gonna work reliably.
To detect a MAS environment, we'd need to 
a) check for a sandbox environment
b) check if the app was signed with a cert by apple (an app can be outside of the MAS but be sandboxed)
c) possibly do more?

In the end, we're closing this PR. If the need arises to detect the MAS as an environment, we'll create a new task and look into how we can detect the MAS reliably (and of course verify).

PR #688, opened 2017-08-02
Test frameworks for tvOS added.

bmourat:
OCHamcrest and OCMockito for tvOS 10.0 added to Vendor folder.
Instructions and scripts for building added to Scripts folder.

PR #698, opened 2017-08-07
Feature/tvos test setup

jaelim-ms:
Targets for tvOS unit test will be added later to avoid merge conflicts of project configurations.


jaelim-ms:
This PR will replace #647 and #688.

Config/Tests tvOS.xcconfig
```diff
@@ -0,0 +1,12 @@
...
+#include "./tvOS.xcconfig"
+
+SEARCH_PATHS = "$(SRCROOT)/../Vendor/tvOS"
+
+OTHER_CFLAGS = $(inherited) -iframework "$(SRCROOT)/../Vendor/tvOS/OCMock" -iframework "$(SRCROOT)/../Vendor/tvOS/OCHamcrest" -iframework "$(SRCROOT)/../Vendor/tvOS/OHHTTPStubs" -iframework "$(PLATFORM_DIR)/Developer/Library/Frameworks"
+OTHER_LDFLAGS = $(inherited) -framework UIKit -ObjC
+
+LD_RUNPATH_SEARCH_PATHS = @executable_path/Frameworks @loader_path/Frameworks
+
+// OCMock and OCHamcrest have a bunch of warnings so we just diable this only for testing.
```
TroubleMakerBen: -> `disable`

I also think that we can enable `TREAT_WARNINGS_AS_ERRORS` again as we're using builds for tvOS of OCMock and OCHamcrest now.

jaelim-ms: The typo is fixed.
For the second part, we need to fix all warnings from test frameworks by changing their code. This should be done prior to enable the flag so I don't think this PR can address the feedback.
In addition to that, fixing warnings would take time so we should take care of it in a separate task.

PR #710, opened 2017-08-11
Feature/add unit tests tvos

jaelim-ms:


Config/Global.xcconfig
```diff
@@ -27,6 +27,9 @@ GCC_C_LANGUAGE_STANDARD = gnu99
 ALWAYS_SEARCH_USER_PATHS = NO
 USER_HEADER_SEARCH_PATHS = "$(SRCROOT)/../MobileCenter/MobileCenter"/**
 
+// :Mark: Code signing
+CODE_SIGN_IDENTITY =
```
TroubleMakerBen: Just being curious..do you want to move this here?

MobileCenter/MobileCenterTests/MSAppDelegateForwarderTests.m
```diff
@@ -606,7 +610,9 @@ - (void)testDontForwardSelectorsNotToOverrideIfAlreadyImplementedByOriginalDeleg
 }
 #endif
 
-#if !TARGET_OS_OSX
+#if TARGET_OS_IOS
+
+// TODO: tvOS doesn't support application:openURL:sourceApplication:annotation:. Temporaily disable the test.
```
TroubleMakerBen: There is no indication that tvOS will ever have a proper "Safari", I don't think we need those around.

Config/Global.xcconfig
jaelim-ms: We haven't set any code signing configurations for all the projects and just use default settings. However, somehow, all tvOS targets have signing settings unlike iOS and macOS targets so I wanted to set this value.
The reason of having this in Global is that we need to take care of signing soon so we can address everything together.

MobileCenter/MobileCenterTests/MSAppDelegateForwarderTests.m
jaelim-ms: Removed all of them.

PR #715, opened 2017-08-17
Do not receive any push notifications when an application is not running

jaelim-ms:


MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -272,8 +271,9 @@ - (BOOL)didReceiveRemoteNotification:(NSDictionary *)userInfo fromUserNotificati
 #if TARGET_OS_OSX
 
     /*
-     * Only call the push delegate if the app is in topmost foreground and the notification is a remote notification. If
-     * the notification is a user notification, it should be consumed now.
+     * Only call the push delegate if the app is in topmost foreground and the notification is a remote notification or
+     * it is a user notification. Otherwise, convert a remote notification to a user notification and handle the
+     * notification when a user clicks it fron notification center.
```
dhei: typo: "fron" -> "from"

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -287,6 +287,13 @@ - (BOOL)didReceiveRemoteNotification:(NSDictionary *)userInfo fromUserNotificati
...
 #if TARGET_OS_OSX
+    } else {
+      NSUserNotification *notification = [[NSUserNotification alloc] init];
+      notification.title = title;
+      notification.informativeText = message;
+      notification.userInfo = userInfo;
+      NSUserNotificationCenter *center = [NSUserNotificationCenter defaultUserNotificationCenter];
+      [center deliverNotification:notification];
     }
 #endif
```
dhei: I have a feedback on readability of existing code, not directly on your new code in this PR. I found [Line 271](https://github.com/Microsoft/mobile-center-sdk-ios/pull/715/files?diff=unified#diff-8a6dda2e220279e67126fd356f3f5381R271) to [Line 298](https://github.com/Microsoft/mobile-center-sdk-ios/pull/715/files?diff=unified#diff-8a6dda2e220279e67126fd356f3f5381R298) are not slightly less readable than I expected because of nested compiler directives `#if TARGET_OS_OSX` inside `if-else` conditions. To be more specific, two pairs of `#if TARGET_OS_OSX` `#endif` makes it hard to understand the logic. To improve readability I hope there could be some refactoring here. At least you can refactor out the shared logic between different platforms to avoid this.

jaelim-ms: Thanks for the feedback. I hope we can revisit and refactor/cleanup all of those `#if` things. 😄

dhei: @jaelim-ms sounds good, we should.

PR #717, opened 2017-08-17
Remove TODOs

jaelim-ms:


dhei:
What about [Line 175](https://github.com/Microsoft/mobile-center-sdk-ios/pull/717/files#diff-8a6dda2e220279e67126fd356f3f5381R175)? Should we remove it?

jaelim-ms:
That line needs to be reviewed by @TroubleMakerBen and @clpolet to make sure that is not necessary.

PR #726, opened 2017-08-23
Mobile Center initialisation moved to applicationDidFinishLaunching

bmourat:


SasquatchMac/SasquatchMacObjC/AppDelegate.m
```diff
@@ -22,11 +21,24 @@ - (instancetype)init {
   [MSMobileCenter start:@"8649b73e-6df0-4985-a039-8ab1453d44f3"
            withServices:@[ [MSAnalytics class], [MSCrashes class], [MSPush class] ]];
   [MobileCenterProvider shared].mobileCenter = [[MobileCenterDelegateObjC alloc] init];
-  return self;
+  [self updateMobileCenterState];
 }
 
 #pragma mark - Private
 
+- (void)updateMobileCenterState{
```
MatkovIvan: Missing space before curly bracket

SasquatchMac/SasquatchMac/ViewControllers/CrashesViewController.swift
```diff
@@ -7,8 +7,12 @@ class CrashesViewController : NSViewController {
   
   override func viewDidLoad() {
     super.viewDidLoad()
+    updateMCState()
+  }
+
+  func updateMCState(){
```
MatkovIvan: Missing space before curly bracket

SasquatchMac/SasquatchMacObjC/AppDelegate.m
```diff
@@ -22,11 +21,24 @@ - (instancetype)init {
   [MSMobileCenter start:@"8649b73e-6df0-4985-a039-8ab1453d44f3"
            withServices:@[ [MSAnalytics class], [MSCrashes class], [MSPush class] ]];
   [MobileCenterProvider shared].mobileCenter = [[MobileCenterDelegateObjC alloc] init];
-  return self;
+  [self updateMobileCenterState];
```
bmourat: Well, i doubt that users will be using MC components in a views, it is just this particular case, testing MC, but never the less, there is a problem with timing: applicationDidFinishLaunching is called after viewDidLoad. What if users want to track some event in viewDidLoad and MC has not been started yet. One possible solution is to introduce on macOS some kind of delegate, but that is not one liner either.

clpolet: I am not very familiar with macOS. Is this a norm on macOS to configure libraries in the `applicationDidFinishLaunching` delegate? Otherwise, do we know where macOS devs are expecting libraries to be configured/started?

jaelim-ms: This is definitely a timing issue between `applicationDidFinishLaunching` and `viewDidLoad`.
I really wonder why this doesn't happen in iOS apps. If this is just a macOS issue then we can probably document to start MobileCenter in `init` but we need to make sure this doesn't break anything because MobileCenter should start after application is ready for use.

jaelim-ms:
Confirmed, it shows the window.

PR #728, opened 2017-08-23
Fix/project configurations

jaelim-ms:


jaelim-ms:
I already tried that but some configurations don't like "config" filter. That's why I separated config files.

PR #730, opened 2017-08-25
Added unit tests to push and distribute modules

bmourat:


MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeAppDelegateTests.m
```diff
@@ -0,0 +1,37 @@
...
+
+@end
+
+@implementation MSDistributeAppDelegateTests
+
+- (void)testOpenURLIsCalled {
+
+  // If.
+  MSDistributeAppDelegate *appDelegate = [[MSDistributeAppDelegate alloc] init];
+  id distributeMock = OCMPartialMock([MSDistribute sharedInstance]);
```
MatkovIvan: All `OCMPartialMock` *should* `stopMocking` then to avoid impact to other tests

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeSenderTests.m
```diff
@@ -1,6 +1,8 @@
 #import "MSDistributeSender.h"
 #import "MSHttpSenderPrivate.h"
 #import "MSTestFrameworks.h"
+#import "MSLogger.h"
+#import "MSDistributePrivate.h"
```
MatkovIvan: Nitpick: please sort imports

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeAppDelegateTests.m
bmourat: > The real object can be returned to its original state by calling stopMocking. This is only necessary if the original state must be restored before the end of the test. The partial mock automatically calls stopMocking during its own deallocation.

From docs: http://ocmock.org/reference/#partial-mocks

MatkovIvan: True for local objects, but I guess `sharedInstance` not deallocated here.

bmourat: distributeMock is local variable, so stopMocking method will be called after its deallocation.

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeAppDelegateTests.m
```diff
@@ -0,0 +1,37 @@
...
+
+  // When.
+  [appDelegate application:[UIApplication sharedApplication] openURL:url sourceApplication:@"" annotation:@"" returnedValue:YES];
+
+  // Then.
+  OCMVerify([distributeMock openURL:url]);
+  XCTAssertEqual(count, 2);
+}
+
+
```
jaelim-ms: Remove empty line. Keep just one.

MobileCenterDistribute/MobileCenterDistributeTests/MSDistributeSenderTests.m
```diff
@@ -11,6 +13,8 @@ @implementation MSDistributeSenderTests
 #pragma mark - Tests
 
 - (void)testCreateRequest {
+
+  // If.
```
jaelim-ms: No dots (.) for `If`/`When`/`Then`

PR #736, opened 2017-08-29
Fix missing sid when logs are sent from crashes buffer

jaelim-ms:
Please review the name of the method.

### DIFF
--- a/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
+++ b/MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
@@ -9,6 +9,14 @@
 @optional
 
 /**
+ * A callback that is called when a log is ready to enqueue.
+ *
+ * @param log The log.
+ * @param internalId An internal Id that can be used to keep track of logs.
+ */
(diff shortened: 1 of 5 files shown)


MobileCenter/MobileCenter/Internals/Channel/MSLogManagerDelegate.h
```diff
@@ -9,6 +9,14 @@
 @optional
 
 /**
+ * A callback that is called when a log is ready to enqueue.
+ *
+ * @param log The log.
+ * @param internalId An internal Id that can be used to keep track of logs.
+ */
+- (void)onPreparedLog:(id<MSLog>)log withInternalId:(NSString *)internalId;
```
clpolet: I would call it "onPreparingLog" since the log is missing the sid so not yet ready to be enqueued.

jaelim-ms: Good point. BTW `sid` is only a concept of Analytics. It doesn't need to be set in Crashes logs if not working with Analytics. Can be considered as extra. Any thoughts @TroubleMakerBen?

PR #740, opened 2017-08-29
Force exit the app while updating on iOS >= 11

clpolet:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -672,11 +672,13 @@ - (void)startDownload:(nullable MSReleaseDetails *)details {
         }
 
         /*
-         * We've seen the behavior on iOS 8.x devices in HockeyApp that it doesn't download until the application
-         * goes in background by pressing home button. Simply exit the app to start the update process.
-         * For iOS version >= 9.0, we still need to exit the app if it is a mandatory update.
+         * On iOS 8.x and >= iOS 11.0 devices the update download doesn't start until the application goes
+         * in background by pressing home button. Simply exit the app to start the update process.
+         * For iOS version >= 9.0 and < iOS 11.0, we still need to exit the app if it is a mandatory update.
```
jaelim-ms: `Only iOS 9.x and 10.x will download the update after users click the "Install" button. We need to force-exit the application for other versions or any versions if it is a mandatory update.`

PR #747, opened 2017-08-30
Exclude disablemachExceptionHandler from header for tvOS and make sur…

TroubleMakerBen:
…e it’s not enabled when initializing MSCrashes on tvOS.

The API won't be available in the header, but I chose to leave the internal methods in there for tvOS.

### DIFF
--- a/MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
+++ b/MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
@@ -70,10 +70,7 @@ typedef struct MSCrashesCallbacks {
   MSCrashesPostCrashSignalCallback handleSignal;
 } MSCrashesCallbacks;
 
-// TODO: Mach exception handler is not supported on tvOS.
-#if !TARGET_OS_TV
 @property(nonatomic, assign, getter=isMachExceptionHandlerEnabled) BOOL enableMachExceptionHandler;
-#endif
 
 /**
  * A list containing all crash files that currently stored on disk for this app.
--- a/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
+++ b/MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
@@ -232,7 +232,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
 /*
  * This can never be bound to Xamarin.
  *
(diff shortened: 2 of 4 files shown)


MobileCenterCrashes/MobileCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -518,6 +518,14 @@ - (void)testDisableMachExceptionWorks {
 }
 #endif
 
+#if TARGET_OS_TV
```
jaelim-ms: You can change the order with previous test method and do 
`#if TARGET_OS_TV`
`#else`
`#endif`

MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
```diff
@@ -70,10 +70,7 @@ typedef struct MSCrashesCallbacks {
   MSCrashesPostCrashSignalCallback handleSignal;
 } MSCrashesCallbacks;
 
-// TODO: Mach exception handler is not supported on tvOS.
-#if !TARGET_OS_TV
```
jaelim-ms: I still don't know this is a right thing to have this property on tvOS. If it is not supported by tvOS, better to remove instead of keeping it. However if you are fine with change, I'm also okay.

MobileCenterCrashes/MobileCenterCrashes/MSCrashes.mm
```diff
@@ -229,13 +229,15 @@ + (MSErrorReport *_Nullable)lastSessionCrashReport {
   return [[self sharedInstance] getLastSessionCrashReport];
 }
 
-/* This can never be binded to Xamarin */
-// TODO: Mach exception handler is not supported on tvOS.
-#if !TARGET_OS_TV
+/*
+ * This can never be bound to Xamarin.
+ *
+ * This method is not part of the publicly available on tvOS as Mach exception handling is not possible on tvOS.
```
clpolet: "This method is not part of the publicly available **APIs** on tvOS" :)

MobileCenterCrashes/MobileCenterCrashes/Internals/MSCrashesPrivate.h
clpolet: I am fine as soon as it's not public.

TroubleMakerBen:
I just addressed the feedback. =)

giphy-downsized

PR #748, opened 2017-08-31
Change date format for timestamp with milliseconds

jaelim-ms:


jaelim-ms:
As @clpolet pointed, at least "timestamp" should have accurate value.

TroubleMakerBen:
As discussed today, this means we need to change PLCR. I have set the process in motion. How about we merge this PR and have a sep. one for PLCR?

PR #757, opened 2017-09-11
Fix/swizzling completion handler chaining

clpolet:
App delegate forwarder now supports application:didReceiveRemoteNotification:fetchCompletionHandler: while already implemented
    
    * Support application:didReceiveRemoteNotification:fetchCompletionHandler: even when already implemented by original delegate
    * Fetch completion handers can run from asynchroneous delegates
    * Fetch completion handlers execution order doesn't matter
    * Add unit tests to cover fetch completion handler use cases


clpolet:
I was thinking: people will have the SDK call already in the application:didReceiveRemoteNotification:fetchCompletionHandler:. This will lead to duplicate. We should prevent duplicate SDK calls before merging this one. But that's another story ;)

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateUtil.h
```diff
@@ -0,0 +1,18 @@
+#if TARGET_OS_OSX
+#import <AppKit/AppKit.h>
+#ifndef MSOriginalAppDelegate
+#define MSOriginalAppDelegate NSApplicationDelegate
+#endif
+#ifndef MSOriginalApplication
```
jaelim-ms: We should check MSOriginalAppDelegate and define all of them if it is not defined.
Even though UIApplication and NSApplication cannot exist together, we have to make sure Xcode can also give us an error (by warnings as errors) for redefining issue.

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateForwarder.m
```diff
@@ -446,29 +403,88 @@ - (void)custom_application:(UIApplication *)application didReceiveRemoteNotifica
...
+         * one
+         * `UIBackgroundFetchResult` in the end so there is a need for triage while comparing results from delegates.
+         *
+         * Priorities are:
+         * `UIBackgroundFetchResultNewData`>`UIBackgroundFetchResultFailed`>`UIBackgroundFetchResultNoData`.
+         *  - `UIBackgroundFetchResultNewData` means at least one of the delegates did download something successfully.
+         *  - `UIBackgroundFetchResultFailed` means there was one/several downloads among the delegates but they failed.
+         *  - `UIBackgroundFetchResultNoData` means that none of the delegates had anything to download.
+         */
+        if (fetchResult != actualFetchResult &&
```
jaelim-ms: I don't think you need this `if` statement. With or without the `if` statement the result will be same if you just do `actualFetchResult = fetchResult` and this means you can just use `fetchResult`.
I doubt if it is not working as your expectation.

clpolet: That's right we can remove `fetchResult != actualFetchResult`, it just spares some unnecessary assignments to `actualFetchResult` (that's why I added in the first place) but removing it will probably improve the readiness.

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateUtil.h
```diff
@@ -0,0 +1,18 @@
+#if TARGET_OS_OSX
+#import <AppKit/AppKit.h>
+#ifndef MSOriginalAppDelegate
```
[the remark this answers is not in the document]
clpolet: Thing is it's then not easy to differentiate `MSApplicationDelegate` and `MSAppDelegate`. It's not obvious that the `MSApplicationDelegate` is the original one since it's prefixed by our prefix ^^. We can probably rename `MSApplication` though since there is no custom application.

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateUtil.h
clpolet: Actually `MSOriginalApplication` should be defined in the utility class since it's not only used by the app delegate forwarder.

jaelim-ms: If that's the case, you can check the definition of Application then just define AppDelegate based on platform?

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateUtil.h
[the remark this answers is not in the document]
jaelim-ms: Yes, I agree that it will be hard to distinguish between `MSApplicationDelegate` and `MSAppDelegate`.
I would recommend renaming `MSAppDelegate` to something else. 😄  Something like `MSCustomApplicationDelegate` and `MSApplicationDelegateForwarder`?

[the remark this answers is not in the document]
clpolet: Renaming MSApplicationDelegateForwarder will be a breaking change because of the `isAppDelegateForwarderEnabled` API. Although the term `AppDelegate` is quite known in th iOS world.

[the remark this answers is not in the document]
clpolet: I think the choice is either `MSCustomApplicationDelegate` and `MSApplicationDelegate` or `MSCustomAppDelegate` and `MSAppDelegate`. Knowing that we should keep `MSAppDelegateForwarder`. @TroubleMakerBen  what's your thoughts.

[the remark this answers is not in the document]
clpolet: Naming will now be: `MSApplicationDelegate`, `MSCustomApplicationDelegate` and `MSAppDelegateForwarder`

MobileCenter/MobileCenter/Internals/AppDelegate/MSAppDelegateUtil.h
clpolet: We'll play it safe and keep it as it is.

clpolet:
We can merge this one but we will have to document that people using push should remove the AppCenter call from their `:didReceiveRemoteNotification:fetchCompletionHandler:` implementation as it's now handled by the swizzling.

clpolet:
Still it's better to wait for the auto dedup or manual selective disabling of the app forwarder.

AppCenter/AppCenter/Internals/AppDelegate/MSAppDelegateForwarder.m
```diff
@@ -446,29 +402,87 @@ - (void)custom_application:(UIApplication *)application didReceiveRemoteNotifica
...
+        }
 
-  // Collect `UIBackgroundFetchResult` from delegates in order to call the original completion handler later.
-  __block UIBackgroundFetchResult forwardedFetchResult = UIBackgroundFetchResultNoData;
-  void (^forwardedCompletionHandler)(UIBackgroundFetchResult) = ^(UIBackgroundFetchResult fetchResult) {
-    forwardedFetchResult = fetchResult;
+        // Be sure original delegate and/or custom delegates and/or the app forwarder executed their completion handler.
+        if ((customHandlerCalledCount == customDelegateToCallCount &&
+             (executors & MSCompletionExecutorOriginal || !originalImp)) ||
+            (executor == MSCompletionExecutorForwarder)) {
```
jaelim-ms: Let's reorder conditions, place `executor == MSCompletionExecutorForwarder` first.

clpolet: Do as you want, as soon as the logic remains the same. There are unit tests to backup this one anyways.

jaelim-ms: I'll change.

TroubleMakerBen:
For completeness sake and in case anyone is interested:
we chose not to implement deduplication as it comes with a big maintenance cost. We'll document the requirement to remove the AppCenter call from their `didReceiveRemoteNotification:fetchCompletionHandler:` implementation as best as we can.

AppCenter/AppCenter/Internals/AppDelegate/MSAppDelegateForwarder.m
```diff
@@ -446,29 +402,86 @@ - (void)custom_application:(UIApplication *)application didReceiveRemoteNotifica
...
+  __block UIBackgroundFetchResult actualFetchResult = UIBackgroundFetchResultNoData;
+  __block short customHandlerCalledCount = 0;
+  __block short customDelegateToCallCount = 0;
+  __block MSCompletionExecutor executors = MSCompletionExecutorNone;
 
-  // Collect `UIBackgroundFetchResult` from delegates in order to call the original completion handler later.
-  __block UIBackgroundFetchResult forwardedFetchResult = UIBackgroundFetchResultNoData;
-  void (^forwardedCompletionHandler)(UIBackgroundFetchResult) = ^(UIBackgroundFetchResult fetchResult) {
-    forwardedFetchResult = fetchResult;
+  // This handler will be used by all the delegates, it unifies the results and execte the real handler at the end.
```
TroubleMakerBen: typo `execute`

TroubleMakerBen:
I really like how NSApplicationDelegate and UIApplicationDelegate is wrapped. 👍

jaelim-ms:
I marked dontmerge for another purpose. PR itself is not blocked at all.

jaelim-ms:
Trigger CI build again

PR #752, opened 2017-09-06
Feature/use sf authentication session

TroubleMakerBen:

PR #760, opened 2017-09-14
Fixes MSDistributeTests compilation error

bmourat:


TroubleMakerBen:
Looks good. What's really odd is that I swear that I tried this earlier and it didn't work. I've cherrypicked that change for #752 to make it pass the unit tests.

TroubleMakerBen:
The tests fail because of an error that was present in #752. I updated #752 with this change here, so I'm closing this PR.

PR #762, opened 2017-09-14
Feature/swizzle user notification center delegate

jaelim-ms:


MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -55,10 +64,63 @@ - (instancetype)init {
...
+
+- (void)observeValueForKeyPath:(NSString *)keyPath
+                      ofObject:(id)object
+                        change:(NSDictionary *)change
+                       context:(void *)context {
+  if (context == UserNotificationCenterDelegateContext &&
+      [keyPath isEqualToString:kMSUserNotificationCenterDelegateKey]) {
+    userNotificationCenterDelegate = [change objectForKey:NSKeyValueChangeNewKey];
+    NSUserNotificationCenter *center = [NSUserNotificationCenter defaultUserNotificationCenter];
+    [center removeObserver:self forKeyPath:keyPath];
```
clpolet: What is the reason behind removing and re-adding self as observer?

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -38,10 +38,19 @@
...
+static NSString *const kMSUserNotificationCenterDelegateKey = @"delegate";
+
+/**
  * Singleton
  */
 static MSPush *sharedInstance = nil;
 static dispatch_once_t onceToken;
+#if TARGET_OS_OSX
+static id<NSUserNotificationCenterDelegate> userNotificationCenterDelegate;
+static void *UserNotificationCenterDelegateContext = &UserNotificationCenterDelegateContext;
```
clpolet: Do we need this? Looks like push doesn't have a super or child class to conflict to.

MobileCenterPush/MobileCenterPush/MSPush.m
clpolet: Ah got it, it's to avoid re-triggering the event for ourself. I would suggest testing `userNotificationCenterDelegate` for simplicity, if the new delegate is us just ignore?

jaelim-ms: That's also possible but I don't think that can simplify the process. By removing/adding observer there can prevent going thru observer process that will eventually just ignore our delegate.

MobileCenterPush/MobileCenterPush/MSPush.m
jaelim-ms: A kind of safeguard. We only intercept if it is our context.

MobileCenterPush/MobileCenterPush/MSPush.m
clpolet: I think it will at least improve readability.

MobileCenterPush/MobileCenterPush/MSPush.m
clpolet: All right.

MobileCenterPush/MobileCenterPush/MSPush.m
TroubleMakerBen: I agree with Clement as it improves readability.

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -258,6 +318,15 @@ - (BOOL)didReceiveRemoteNotification:(NSDictionary *)userInfo fromUserNotificati
   // The notification is not for Mobile Center if customData is nil. Ignore the notification.
   NSDictionary *customData = [userInfo objectForKey:kMSPushNotificationCustomDataKey];
   if (customData) {
+
+    // If Push is disabled, discard the notification.
+    if (![[self class] isEnabled]) {
+      MSLogVerbose(
+          [MSPush logTag],
+          @"Notification received while Push was enabled but Push is not disabled now, discard the notification.");
```
clpolet: `Notification received while Push was enabled but it is disabled now, discard the notification.`

jaelim-ms:
`didDeliverNotification` is called when a user notification is delivered. We don't actually do on this method since `didActivateNotification` already handled the user notification.
`shouldPresentNotification` is only called if User Notification Center couldn't determine whether it needs to be presented or not. Our scenario doesn't have the case that the method is called. It is the only case if the Alert type is enabled.

jaelim-ms:
OK I got your point. Those methods in custom delegate won't be called, adding a fix for that.

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -38,10 +38,19 @@
...
+ */
+static NSString *const kMSUserNotificationCenterDelegateKey = @"delegate";
+
+/**
  * Singleton
  */
 static MSPush *sharedInstance = nil;
 static dispatch_once_t onceToken;
+#if TARGET_OS_OSX
+static id<NSUserNotificationCenterDelegate> userNotificationCenterDelegate;
```
clpolet: Can this one be a property instead? And probably renaming it to make it explicit that it's the original userNotificationCenterDelegate not ours.

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -219,9 +276,30 @@ - (void)applicationDidFinishLaunching:(NSNotification *)notification {
   [self didReceiveUserNotification:[notification.userInfo objectForKey:NSApplicationLaunchUserNotificationKey]];
 }
 
-- (void)userNotificationCenter:(NSUserNotificationCenter *)__unused center
+- (void)userNotificationCenter:(NSUserNotificationCenter *)center
```
clpolet: As a note, we can use message forwarding to dynamically forward to the original delegate without explicitly implement all the methods. It's particularly useful for instance whenever a new method is added to the delegate, then there is no need for updating the SDK. But that can be an improvement in another story.

TroubleMakerBen: +1 for doing this in a separate story.

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -219,9 +277,21 @@ - (void)applicationDidFinishLaunching:(NSNotification *)notification {
...
        didActivateNotification:(NSUserNotification *)notification {
   [self didReceiveUserNotification:notification];
+  if ([self.originalUserNotificationCenterDelegate
+          respondsToSelector:@selector(userNotificationCenter:didActivateNotification:)]) {
+    [self.originalUserNotificationCenterDelegate userNotificationCenter:center didActivateNotification:notification];
+  }
+}
+
+- (void)forwardInvocation:(NSInvocation *)anInvocation {
+  if ([self.originalUserNotificationCenterDelegate respondsToSelector:[anInvocation selector]]) {
```
clpolet: Shouldn't we make sure the selector is part of the  NSUserNotificationDelegate protocol in addition to the responsiveness? Just to be sure we don't forward anything to the delegate.

jaelim-ms: Good point.

MobileCenterPush/MobileCenterPush/MSPush.m
```diff
@@ -287,7 +288,10 @@ - (void)userNotificationCenter:(NSUserNotificationCenter *)center
 }
 
 - (void)forwardInvocation:(NSInvocation *)anInvocation {
-  if ([self.originalUserNotificationCenterDelegate respondsToSelector:[anInvocation selector]]) {
+
+  // Testing if the selector is defined in NSUserNotificationCenterDelegate or not.
+  struct objc_method_description hasMethod = protocol_getMethodDescription(@protocol(NSUserNotificationCenterDelegate), [anInvocation selector], NO, YES);
+  if (hasMethod.name != nil && [self.originalUserNotificationCenterDelegate respondsToSelector:[anInvocation selector]]) {
```
clpolet: I would just use `NULL` instead of `nil` since we are not comparing an objc object: `hasMethod.name != NULL`.

jaelim-ms: Done.

PR #763, opened 2017-09-14
Add silent notifications to apps + fix sasquatch ios swift not calling fetch handler

clpolet:


Puppet/Puppet/AppDelegate.m
```diff
@@ -226,23 +222,32 @@ - (BOOL)distribute:(MSDistribute *)distribute releaseAvailableWithDetails:(MSRel
...
   for (NSString *key in pushNotification.customData) {
-    message = [NSString stringWithFormat:@"%@\n%@: %@", message, key, [pushNotification.customData objectForKey:key]];
+    ([customData length] == 0) ? customData = [NSMutableString new] : [customData appendString:@", "];
+    [customData appendFormat:@"%@: %@", key, [pushNotification.customData objectForKey:key]];
+  }
+  if (UIApplication.sharedApplication.applicationState == UIApplicationStateBackground) {
+    NSLog(@"%@ Notification received in background, title: \"%@\", message: \"%@\", custom data: \"%@\"", kPUPLogTag,
+          title, message, customData);
+  } else {
+    message = [NSString stringWithFormat:@"%@%@%@", (message ? message : @""), (message && customData ? @"\n" : @""),
```
jaelim-ms: You don't need to check the `message` since it is a mandatory field.

Puppet/Puppet.xcodeproj/project.pbxproj
```diff
@@ -788,6 +812,9 @@
 						DevelopmentTeam = L4RARDNZ2Y;
 						ProvisioningStyle = Manual;
 						SystemCapabilities = {
+							com.apple.BackgroundModes = {
```
TroubleMakerBen: I get why we are doing this, but we need to remember during testing that we can't assume our users set this in their apps.

Puppet/Puppet/AppDelegate.m
clpolet: A mandatory field yes but not for long ;). Can send silent notifications without title and messages and the backend is getting prepared to that.

jaelim-ms: OK. I've heard about not having title and message in payload. 😄

PR #765, opened 2017-09-20
Release 0.12.2

jaelim-ms:


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
 
+## Version 0.12.2
+
+This version contains improvements. **Verified all functionalities against iOS 11 GM.**
+
+### MobileCenterCrashes
+
+* **[Improvement]** Added a millisecond precision to crash logs for more accurate log time.
+
+### MobileCenterDistribute
```
clpolet: Also the behavior has changed on iOS 11 now that we are using the new feature a popup is displayed.

clpolet: There was a workaround before and the behavior noticeably changed between the two versions.

jaelim-ms: OK. We are bringing it back so I added that to the change log.

jaelim-ms: For popup thing, it is a new behavior on iOS 11. Not mentioning details here.

TroubleMakerBen: Maybe say "When authenticating for in-app-updates, iOS 11 now displays a confirmation dialog to access login credentials"? I would mention it.

PR #768, opened 2017-09-22
Fix/fix compatibility with xcode8

TroubleMakerBen:


MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -457,36 +449,57 @@ - (nullable NSURL *)buildTokenRequestURLWithAppSecret:(NSString *)appSecret rele
...
+      if (error.code == 1) {
+        MSLogError([MSDistribute logTag], @"Authentication session was cancelled by user or failed.");
+      }
+      if (callbackUrl) {
+        [self openURL:callbackUrl];
+      }
+    };
+
+    // Initialize the SFAuthenticationsession.
+    typedef void (*MSInitSFAuthenticationSession)(id, SEL, NSURL *, NSString *, MSCompletionBlockForAuthSession);
```
clpolet: I love it!

Config/Global.xcconfig
```diff
@@ -17,6 +17,7 @@ PRODUCT_NAME = $(PROJECT_NAME)
 PUBLIC_HEADERS_FOLDER_PATH = include/$(PRODUCT_NAME)
 
 // :Mark: Language configurations
+
```
clpolet: Is it normal not to have the `CLANG_MODULES_AUTOLINK = NO` here?

jaelim-ms: We don't need that because we stay to build the SDK on Xcode 8.

jaelim-ms: @TroubleMakerBen Probably you can remove the empty line. 😄

clpolet: OK so we Bitrise is back to XCode 8 I guess.

PR #773, opened 2017-09-25
Only show in-app update failure message once per release

JessicaYeh:
Currently, if the in-app update setup process fails, the app will continue to try to set it up on each fresh app launch. That's a pretty annoying user experience, so this PR is an attempt to make that better. In the new experience, if it's detected that the update setup failed (through a new query parameter, `update_setup_failed={error message}` that is sent back to the SDK), and the user chooses to ignore the failure message, then the SDK will make note of which package/release hash the failure was encountered on, and won't repeat the update setup process until the hash is different (i.e., the user installs a new version of the app). There is also a `Re-install` option on the error dialog (that now shows from the SDK as opposed to the page in Safari), and if the user clicks that, it will take them to the Mobile Center install portal to try re-installing the app. A future improvement would be to take them to the exact app install page so they don't need to navigate back to it.

Apps using prior versions of the SDK will still show the error message in Safari (since on the update-setup page, we'll check whether the SDK sends a new `enable_failure_redirect` query parameter).

To do:
- [ ] Get localized translations for the new strings

img_4633


pniko:
@JessicaYeh please update the screengrab when you can to reflect the latest messaging we dropped in.

JessicaYeh:
@pniko updated!

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -449,6 +441,37 @@ - (nullable NSURL *)buildTokenRequestURLWithAppSecret:(NSString *)appSecret rele
   return components.URL;
 }
 
+- (void)openUrlInSafari:(NSURL *)url {
```
TroubleMakerBen: Can we rename this to something like `startAuthenticationForUrl:(NSURL *)url`? The method name is misleading as only iOS 8 will actually open Safari. The method name confused me for a bit as I thought "where the heck did the `SFAuthenticationSession`/`SFSafariViewController` logic go?!". But definitely 👍 for extracting the logic into it's own method.

JessicaYeh: 👍 Good point! `openUrlInSafari` is kind of confusing. But I wonder if having `Authentication` in the function name would not make sense for the non-iOS 11 cases where it doesn't actually pull up that authentication message before going into the SafariViewController?

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -204,41 +204,32 @@ - (void)requestInstallInformationWith:(NSString *)releaseHash {
           @"The device lost its internet connection. The SDK will retry to get an update token in the next launch.");
       return;
     }
+    
+    // If failed to enable in-app updates on the same app build before, don't try again.
```
jaelim-ms: Multi line comment should use `/* */`.

elamalani:
@jaelim-ms @TroubleMakerBen Localization work is in progress in our team. Can you make sure these new strings are localized as well?

deskfolio:
Let's see, information wise, I think we can do better with the copy/message:

1. I wouldn't treat or refer to this as an "error" dialog. Let's treat it as offering the user the possibility of "enabling" an amazing feature that for some reason and at some point wasn't enabled.

2. Title looks good.

3. I have NO idea what a "side-loaded" release is. Also "... A browser using private mode" is kinda misleading since is not an accurate requirement, because it cannot any browser but a specific one. On iOS apps can ONLY be installed using Safari (on non private mode), no? Anyway, do we actually have to explain so much? I would rather let the user know that this feature is disable, and they can enable it by reinstalling the app. That's all.

4. It would be great is I, as a user, don't have to re-install the app to get this feature enable. To me, features are either enabled or disabled, and the install requirement seems a little-bit-too-much. But I guess we can't do much about this.

5. As soon as the user taps on the "Ignore" button, I would recommend not showing the dialog again as you suggested.

6. As soon as the user taps on the "Reinstall app" button, I would take them to the app install page. If the user happens to be on "private mode" again, and this feature won't get enabled once again due to that, I would try to show a message to the user specifying the actual requirement for the installation to be successful.

Here's a revisited mockup:

screen shot 2017-09-27 at 1 05 10 pm

Maybe someone can review and help with the copy a bit just to make sure that it's what we want.

JessicaYeh:
@deskfolio Thanks for your review!

3. Yeah, I had the same concern that testers might not know what side-loaded means. And yes that's true that it must be a specific browser. FYI: recently we completely removed the private browsing check (and also the restriction to not allow installs in private browsing mode) at the request of Thomas. Which makes sense because majority of the installed apps aren't using the distribution SDK. I kind of worry that if we just tell them that reinstalling the app should fix the issue, they won't realize that they might need to disable private browsing mode first, and then they get stuck.

4. Agreed. The reinstall shouldn't be necessary. In the future, maybe we could have the SDK open up a page in Mobile Center that just re-attempts saving the cookies that are needed for enabling in-app updates, so that they don't need to completely reinstall the app.

5. Yup! That's exactly how it will work 😄 Sorry I didn't explain what the buttons did exactly. If someone presses `Ignore` it will not try to enable in-app updates again, until they install another version of the app where the package hash has changed. If they press `Reinstall`, it will take them to the install portal where they can reinstall the app, and the next time that they open up the app (can be same version), it will still try to enable in-app updates.

6. Yea, a help page with some common solutions to the problem would be really nice. I'll actually make a small change now to append a query parameter to the reinstall link so that the page will know that it should show some help info.

pniko:
@deskfolio our main goal with this PR was to resolve an outstanding issue users are having today and make sure they can get back to testing as quickly without future interruptions from the page telling them in-app updates require them to reinstall. 

After speaking with Thomas about this last week we had settled on making this process as quick to get out of as possible while why work on better UI not only on this site, but for the dialog as well.

I agree we should have a nicer landing page in the future to explain the issues when they click "Reinstall" (which should become something like "Learn More" in the future to make it friendlier). 
Right now I would like to get this one in so users can get the immediate benefit of not being hassled while we work on improving the UX overall.

Right after this PR we should discuss and execute:
- Creating a better landing page 
- Creating a nicer popup modal in the users app to replace this temporary one
- Adding in the ability to re-login and get the appropriate checks done from the landing page so users don't need to reinstall the application

Looping in Garrett right now.

MobileCenterDistribute/MobileCenterDistribute/MSDistribute.m
```diff
@@ -707,6 +732,42 @@ - (void)showDistributeDisabledAlert {
   });
 }
 
+- (void)showUpdateSetupFailedAlert:(NSString *)errorMessage {
```
clpolet: Just one very picky comment, there is a convention across the SDKs where empty lines within code blocs should always be followed by a comment. For instance:
```
  int a = 1;

  // Awesome example.
  a++; 
```

JessicaYeh: ahh I see, thanks for the info! I like that convention 😄

JessicaYeh:
@deskfolio 

The default behavior for the iOS alert dialog buttons is to have them side by side horizontally, if both of the texts inside of the buttons can fit in there. And the button texts “Reinstall App” and “Ignore” can fit fine in there side by side. The only way to make it be vertical buttons is to either make one of those button texts longer, or we can hack it and add extra spaces to the beginning/end of “Reinstall App” or something to make the text slightly longer. Adding the extra spaces feels weird, so I think it would be better to just leave it as is and keep the default iOS behavior

img_4668

jaelim-ms:
Please do not merge this PR. I'll do reformat all files in the PR before merging it to develop.

deskfolio:
@JessicaYeh Sorry if I'm late with the following comments, but I just saw the update on this issue:

- "Reinstall app" should we sentence case and not "Re-install App".
- The "Ignore" button should be on the left side.

On this dialog when trying to install an app, we use "Install" on the right and "Cancel" (Ignore) on the left. We should keep that consistent unless we have a good excuse not to.

bbuttons

JessicaYeh:
@deskfolio Sure thing! I've updated it so that it says "Reinstall app" instead of "Reinstall App" and the "Reinstall app" button is now on the right side (and still bolded).

@jaelim-ms @clpolet @TroubleMakerBen FYI: I made a small change to [MSAlertController's addPreferredActionWithTitle](https://github.com/Microsoft/mobile-center-sdk-ios/pull/773/commits/1c4889598b509391fa49a8e6559fa7168d87594e) in order to change the order of the buttons. If the `cancel` style is used, then the button will always be on the left. If we change that to `default`, the button can be moved to the right by just having that action added after the other default action. And it will still appear bolded, because the effect of `setPreferredAction` is to bold the button.

jaelim-ms:
@JessicaYeh Have you tested the alert in iOS 11?

JessicaYeh:
@jaelim-ms Nope, don't have access to an iOS 8 device. :(

JessicaYeh:
@jaelim-ms Here's what the alert might look like in an iOS 8 device. I used an 8.4 simulator on Xcode. Note that the preferred action button is correctly bolded, but will be on the left side because iOS 8 doesn't support `setPreferredAction`

jaelim-ms:
Yeah, I also tested on iOS 8 device and confirmed that your change doesn't affect the behavior in iOS 8.

MobileCenterDistribute/MobileCenterDistribute/MSDistributePrivate.h
```diff
@@ -82,6 +84,16 @@ static NSString *const kMSUpdateTokenKey = @"MSUpdateToken";
  */
 static NSString *const kMSDistributionGroupIdKey = @"MSDistributionGroupId";
 
+/**
+ * The storage key for update setup failure error message.
+ */
+static NSString *const kMSUpdateSetupFailedMessageKey = @"MSUpdateSetupFailedMessage";
```
jaelim-ms: Looks this is not necessary.

PR #776, opened 2017-09-27
Add getSdkVersion API

jaelim-ms:


MobileCenter/MobileCenter/MSMobileCenter.h
```diff
@@ -138,4 +138,9 @@
  */
 + (BOOL)isDebuggerAttached;
 
+/**
+ * Get the current version of MobileCenter SDK.
+ */
++ (NSString *)getSdkVersion;
```
TroubleMakerBen: The naming of this violates ObjC convention. It should be `(NSString *)sdkVersion`.

jaelim-ms: Yes. But compare to other public APIs such as `setLogLevel` and `setLogUrl`, it should be like that. Isn't it?

TroubleMakerBen: It should not be called `getSdkVersion` in my opinion. We are not truly exposing a `@property` but almost. So we should follow the convention. From [Apple's documentation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Conventions/Conventions.html):
```
When you use the @property syntax to declare properties on an object, as described in Encapsulating Data, the compiler automatically synthesizes the relevant getter and setter methods (unless you indicate otherwise). If you need to provide your own accessor method implementations for any reason, it’s important to make sure that you use the right method names for a property in order for your methods to be called through dot syntax, for example.

Unless specified otherwise, a getter method should use the same name as the property. For a property called firstName, the accessor method should also be called firstName. The exception to this rule is for Boolean properties, for which the getter method should start with is. For a property called paused, for example, the getter method should be called isPaused.

The setter method for a property should use the form setPropertyName:. For a property called firstName, the setter method should be called setFirstName:; for a Boolean property called paused, the setter method should be called setPaused:.

Although the @property syntax allows you to specify different accessor method names, you should only do so for situations like a Boolean property. It’s essential to follow the conventions described here, otherwise techniques like Key Value Coding (the ability to get or set a property using valueForKey: and setValue:forKey:) won’t work. For more information on KVC, see Key-Value Coding Programming Guide.```

Comment (jaelim-ms) [reply MobileCenter/MobileCenter/MSMobileCenter.h]: Hmm. I get confused. `getSdkVersion` is actually not getting a value in property. I mean the `sdkVersion` is not a property of the class. Still?

Comment (clpolet) [reply MobileCenter/MobileCenter/MSMobileCenter.h]: Yep this rules of not adding 'get' to the method name applies to all method names not only properties I think.

File: MobileCenter/MobileCenter/MSMobileCenter.h [hunk 2]
```diff
@@ -138,4 +138,9 @@
  */
 + (BOOL)isDebuggerAttached;
 
+/**
+ * Get the current version of MobileCenter SDK.
+ */
```
clpolet: Missing an '@return' doc.

jaelim-ms: 😢

PR #784, opened 2017-09-29
Check silent push flag and not forward to Notification Center

jaelim-ms:
Requires a discussion with PMs.


jaelim-ms:
Closing this one by PM decision. macOS will not support silent push.

PR #801, opened 2017-10-13
Refactor folder structure for SasquatchMac app

jaelim-ms:


clpolet:
Right so we need this other project file for the SDK, maybe we can just rename the scheme from the TestApps projects

PR #821, opened 2017-10-26
Run user confirmation handler from main thread

clpolet:
This is a hot fix. 
Run crash user confirmation handler from main thread


MobileCenterCrashes/MobileCenterCrashes.xcodeproj/project.pbxproj
```diff
@@ -839,17 +839,24 @@
 			name = iOS;
 			sourceTree = "<group>";
 		};
-		350B29E91F192A5C009B91CF /* WrapperSdkUtilities */ = {
+		350B29E91F192A5C009B91CF /* WrapperSDKUtilities */ = {
```
jaelim-ms: Why do we have two groups for WrapperSDKUtilities?

clpolet: One is internal, this other one private ;)

PR #841, opened 2017-11-14
Add unit tests to cover the check lastest release request with failure

MatkovIvan:


AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -77,10 +86,14 @@ @implementation MSDistributeTests
 
 - (void)setUp {
   [super setUp];
+  [MSLogger setCurrentLogLevel:MSLogLevelVerbose];
```
TroubleMakerBen: I think this can be removed.

MatkovIvan: Log level also set in other tests and we have [conditions dependent on it](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/9449f274db22f9d5c3777a3c749b4b1ff0afdf53/AppCenterDistribute/AppCenterDistribute/Internals/Sender/MSDistributeSender.m#L58-L65). So, I faced to exception only if I ran all tests, but one test or all tests in this file passed fine. This's why it's here - to get the same behavior.

TroubleMakerBen: Gotcha!

PR #843, opened 2017-11-14
Support sending logs in the background

MatkovIvan:


AppCenter/AppCenter/Internals/Channel/MSChannel.h
```diff
@@ -71,6 +73,19 @@ typedef void (^enqueueCompletionBlock)(BOOL);
...
+/**
+ * Stop flushing logs.
+ *
+ * @param completion Completion block to be called when stopped flushing logs.
+ */
+- (void)stopFlushingWithCompletion:(MSStopFlushingCompletionBlock)completion;
+
+/**
+ * Resume the channel. It will resume it's normal activity. A channel can't be resumed if it's disabled or the sender is
+ * still supsended.
```
TroubleMakerBen: There's a cute typo here, it should say `suspended`.

AppCenter/AppCenter/Internals/Channel/MSLogManagerDefault.m
```diff
@@ -184,6 +192,21 @@ - (void)processLog:(id<MSLog>)log forGroupId:(NSString *)groupId {
                                       }];
           }
         }];
+  
+  /*
+   * If the app in the background we need:
```
TroubleMakerBen: There needs to be `is` somewhere in the sentence.

AppCenter/AppCenter/Internals/Channel/MSChannel.h
```diff
@@ -9,13 +9,15 @@
 #import "MSStorage.h"
 
 @protocol MSChannelDelegate;
+@protocol MSChannel;
```
jaelim-ms: Order

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -338,18 +338,7 @@ - (void)applyPipelineEnabledState:(BOOL)isEnabled {
   [MS_NOTIFICATION_CENTER removeObserver:self];
 
   // Hookup to application life-cycle events
```
jaelim-ms: Remove the comment.

AppCenter/AppCenter/Internals/Channel/MSChannelDefault.m
```diff
@@ -16,8 +17,9 @@ - (instancetype)init {
     _pendingBatchQueueFull = NO;
     _availableBatchFromStorage = NO;
     _enabled = YES;
-
+    _suspended = NO;
     _delegates = [NSHashTable weakObjectsHashTable];
+    _stopFlushingCompletion = [NSMutableArray new];
```
clpolet: What was the motivation behind having an array of completion handlers instead of just a single completion handler?

clpolet: Do you know how a channel can get multiple completion handlers?

MatkovIvan: Pay attention that we immediately stop flushing the channel on processing log if the app is in the background.
Also, see `testStopFlushingWhenItsAlreadyGoing` test.

clpolet: To me the only object the calling the `stopFlushingWithCompletion` in each channels is the logManager, the problem is if two completion handlers are called for the same channel on the logManager then the `remainedChannelsCount` will be corrupted because decrease twice for the same channel.

clpolet: I see, `remainedChannelsCount` is actually increased twice too so not a problem

jaelim-ms:
After a discussion, we decide to expose an API to enable sending logs in background.

PR #844, opened 2017-11-14
Fix flaky unit test failures

MatkovIvan:


AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
```diff
@@ -164,12 +164,15 @@ - (void)testProcessIdAndExceptionForObjectiveCExceptionCrash {
 }
 
 - (void)testSelectorForRegisterWithName {
-  NSData *crashData = [[[MSCrashes sharedInstance] plCrashReporter] generateLiveReport];
+
```
TroubleMakerBen: Does switching this to a fixture help with improving the flakiness of this test? This is interesting!

TroubleMakerBen: On the other hand, the `generateLiveReport`-API doesn't generate a "full" report but a "live crash report" that's non-crashing, so probably not the way to unit test the selectors anyway. ^^

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -112,6 +116,7 @@ - (void)tearDown {
   [self waitForExpectations:@[ expection ] timeout:1];
 
   // Clear
+  [OHHTTPStubs removeAllStubs];
```
TroubleMakerBen: Just nit-picking. If we have `MSHttpTestUtil`, why not move this line over there, too? I know that it is just one line, but it seems to be more concise to me.

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
MatkovIvan: I've faced that `generateLiveReport` may return `nil`, and it happens sometimes on CI.

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -351,6 +351,7 @@ - (NSURLSession *)session {
     NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration defaultSessionConfiguration];
     sessionConfiguration.timeoutIntervalForRequest = kRequestTimeout;
     _session = [NSURLSession sessionWithConfiguration:sessionConfiguration];
+    _session.delegateQueue.maxConcurrentOperationCount = 1;
```
TroubleMakerBen: I'm sceptical here. We're limiting to one concurrent operation and we really shouldn't do that.

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
TroubleMakerBen: ok

MatkovIvan:
Catch this! Should fail all time before, but [filename of crash file is time-based](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/1.0.1/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm#L898), so if multiple crashes were written in the same second, the file was overwritten.

MatkovIvan:
🤞

PR #846, opened 2017-11-14
Fix issues with manual crash processing for react native

achocron:


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -245,7 +245,6 @@ - (instancetype)init {
 #pragma mark - MSServiceAbstract
 
 - (void)applyEnabledState:(BOOL)isEnabled {
-  [super applyEnabledState:isEnabled];
```
jaelim-ms: Why are you removing this line?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -607,21 +608,24 @@ - (void)startDelayedCrashProcessing {
...
+      [self startCrashProcessing];
+    }
+    @finally {
+      dispatch_semaphore_signal(self.delayedProcessingSemaphore);
+    }
   });
 }
 
 - (void)startCrashProcessing {
-
```
jaelim-ms: Add this back.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -595,7 +597,6 @@ - (void)configureCrashReporterWithUncaughtExceptionHandlerEnabled:(BOOL)enableUn
 #pragma mark - Crash processing
 
 - (void)startDelayedCrashProcessing {
-
```
jaelim-ms: Add this back.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -25,7 +25,7 @@
 static NSString *const kMSTestAppSecret = @"TestAppSecret";
 static NSString *const kMSCrashesServiceName = @"Crashes";
 static NSString *const kMSFatal = @"fatal";
-static NSString *const kMSTypeHandledError = @"handledError";
+static NSString *const kMSTypeHandledError = @"handled_error";
```
clpolet: Isn't that Camel case now?

clpolet: Oh that's just a test

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -833,35 +822,35 @@ - (void)testStartingCrashesWithoutAutomaticProcessing {
...
   }
-  [crashes startWithLogManager:logManager appSecret:kMSTestAppSecret];
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [crashes startWithLogManager:logManager appSecret:kMSTestAppSecret];
+  });
+
+  //  [crashes startWithLogManager:logManager appSecret:kMSTestAppSecret];
   if (startWithReports) {
-    assertThat(crashes.crashFiles, hasCountOf(1));
+    //assertThat(crashes.crashFiles, hasCountOf(1));
```
clpolet: There's some commented code here, I don't think it's on purpose.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -611,9 +618,15 @@ - (void)startDelayedCrashProcessing {
   // This must be performed asynchronously to prevent a deadlock with 'unprocessedCrashReports'.
   dispatch_time_t delay = dispatch_time(DISPATCH_TIME_NOW, (1 * NSEC_PER_SEC));
   dispatch_after(delay, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-      [self startCrashProcessing];
+    [self startCrashProcessing];
 
-      dispatch_semaphore_signal(self.delayedProcessingSemaphore);
+    // Only release once to avoid releasing an unbounded number of times.
+    @synchronized(self) {
```
clpolet: A dispatch once should do the work nicely here instead of the synchronized + property:
```
  dispatch_once(&onceToken, ^{
    // Do something only once
  });
```

TroubleMakerBen: Now that I think about it, that'd be better.

achocron: I hesitate to use a `static` `onceToken`. The problem is that since this is an instance method with its own semaphore, any other instances will be deadlocked, which can be a problem for tests. So I guess I will need to `malloc` and `free` space for a `dispatch_token_t`? (If I make it a `@property`, then I run into the problem of needing to access its address, which can't be done.

TroubleMakerBen: I'm hesitant to use `malloc`/`free`. I'm fine if we keep the variable around. It's easy to understand, which I like a lot.

clpolet: OK I am fine with that.

PR #855, opened 2017-11-15
Fix warnings related to location of license file in pod spec and pres…

TroubleMakerBen:
…erve the path to the readme.md


TroubleMakerBen:
FYI: Linking to the ios license file as we can't specify license per platform.

PR #871, opened 2017-12-04
Feature/environment variables for test cloud

achocron:
Spoke with karl and instead of proposed names for services in the env. var. we will use service name. Also, I'm worried that some tests will become flakey due to concurrency and environment variable usage


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -448,4 +454,24 @@ - (void)applicationDidEnterBackground {
...
+ *
+ * @return YES if the service should be disabled.
+ */
+- (BOOL)shouldDisable:(NSString*)serviceName {
+  NSDictionary *environmentVariables = [[NSProcessInfo processInfo] environment];
+  NSString *disabledServices = environmentVariables[@"APP_CENTER_DISABLE"];
+  if (!disabledServices) {
+    return NO;
+  }
+  NSArray* disabledServicesList = [disabledServices componentsSeparatedByString:@","];
```
jaelim-ms: Can you also trim the items in the list? Just adding one more protection if the environment variable has spaces.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -448,4 +454,24 @@ - (void)applicationDidEnterBackground {
...
+  return  [disabledServicesList containsObject:serviceName] ||
+  [disabledServicesList containsObject:@"All"];
```
jaelim-ms: Formatting.

AppCenter/AppCenterTests/Util/MSMockSecondService.m
```diff
@@ -0,0 +1,50 @@
+#import "MSMockSecondService.h"
+#import "MSChannelConfiguration.h"
+
+static NSString *const kMSServiceName = @"MSMockSecondService";
```
jaelim-ms: Why don't you just use MSMockService by adding setters for serviceName and groupId in `init`?

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -152,6 +153,48 @@ - (void)testSdkVersion {
...
+
+  // When
+  [MSAppCenter start:@"AppSecret" withServices:@[MSMockService.class, MSMockSecondService.class]];
+
+  // Then
+  XCTAssertFalse([[MSMockService sharedInstance] started]);
+  XCTAssertFalse([[MSMockSecondService sharedInstance] started]);
+
+  // If
+  setenv("APP_CENTER_DISABLE", "MSMockService", 1);
```
jaelim-ms: Can you read class name from `[MSMockService class]`?

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -152,6 +153,48 @@ - (void)testSdkVersion {
   XCTAssertTrue([[MSAppCenter sdkVersion] isEqualToString:version]);
 }
 
+- (void)testDisableServicesWithEnvironmentVariable {
+  // If
```
jaelim-ms: Please review coding style in your files.

AppCenter/AppCenter/MSAppCenter.m
jaelim-ms: And constant for `All`.

AppCenter/AppCenterTests/Util/MSMockSecondService.m
achocron: I would, but `serviceName` is class-level, so if I changed it then it would return the wrong value for the other one

achocron: I can make this better by making a base class for MSMockService, which would allow easy creation of mock services with different names for scenarios like this one

jaelim-ms: Then why don't you extend MSMockService class instead of writing all methods in the second class?

achocron: Because class level methods can't be overridden, so it would require one base class and two derived classes

achocron: which I can do

AppCenter/AppCenter/MSAppCenter.m
achocron: Do you mean whitespace at the beginning/end? or even in the middle?

AppCenter/AppCenterTests/Util/MSMockSecondService.m
jaelim-ms: You can create member methods and call this from class level methods. Then you just need to override methods that you created. Whenever you call class level methods, `sharedInstance` will always be ready to use.

AppCenter/AppCenter/MSAppCenter.m
jaelim-ms: My concern is that if the value is provided like `Something, Something, Something`.
In this case, you will have `Something`, `<space>Something` and `<space>Something`.
I don't think you need to take care of trimming all spaces at the end or in the middle of strings but beginning should be.

AppCenter/AppCenterTests/Util/MSMockSecondService.m
jaelim-ms: I thought something wrong there. `sharedInstance` is also a class level method. 😮 
Don't mind.

PR #873, opened 2017-12-06
Support processing of handled exceptions coming from wrapper SDKs

chrwend:
## Context

While the currently exposed methods to enable crash reporting for wrapper SDKs only accepts one exception (that gets correlated to the actual iOS crash), for Unity we need a way to enqueue and send out multiple handled exceptions per session.

## Changes

Expose a function that directly adds an exception object to the channel queue.

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -139,6 +139,11 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
  */
 @property dispatch_semaphore_t delayedProcessingSemaphore;
 
+/**
+ * Track handled exception directly as model form. This API is not public and is used by wrapper SDKs.
+ */
++ (void)trackModelException:(MSException *)exception;
+
 @end
 
 @implementation MSCrashes


AppCenterCrashes/AppCenterCrashes/Internals/MSWrapperExceptionManager.m
```diff
@@ -32,6 +32,13 @@ + (void)saveWrapperException:(MSWrapperException *)wrapperException {
   [self saveWrapperException:wrapperException withBaseFilename:kMSLastWrapperExceptionFileName];
 }
 
+/**
+ * Enqueue and send out handled exceptions coming from wrapper SDK.
+ */
++ (void)trackHandledWrapperException:(MSWrapperException *)wrapperException;
+  [MSCrashes sharedInstance] trackModelException:wrapperException;
```
MatkovIvan: `[]` are missed

jaelim-ms:
Close and open it again to trigger VSTS build

jaelim-ms:
Close and open it again to trigger VSTS build

jaelim-ms:
Final try with VSTS Build

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -139,6 +139,11 @@ @interface MSCrashes ()
  */
 @property dispatch_semaphore_t delayedProcessingSemaphore;
 
+/**
+ * Track handled exception directly as model form. This API is not public and is used by wrapper SDKs.
+ */
```
clpolet: You also need to add the `exception` parameter in the method documentation :).

PR #885, opened 2017-12-19
Fix state of services and internal classes when AppCenter started as disabled

MatkovIvan:
There is a bug when `AppCenter` was disabled in the previous session.
In this case, log manager, seder, channels, services are enabled after start.
This PR harmonized behavior when `AppCenter` disabled in current and previous session.


AppCenter/AppCenter/Internals/Channel/MSChannelDefault.m
```diff
@@ -94,6 +96,11 @@ - (void)enqueueItem:(id<MSLog>)item withCompletion:(enqueueCompletionBlock)compl
     if (completion) {
       completion(success);
     }
+    
+    // Don't flush while disabled.
+    if (!self.enabled) {
```
bmourat: Do we need this when we have the same condition in flushQueue and startTimer?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -285,6 +280,13 @@ - (BOOL)startService:(Class)clazz andSendLog:(BOOL)sendLog {
 
     // Start service with log manager.
     [service startWithLogManager:self.logManager appSecret:self.appSecret];
+    
+    // Disable service if AppCenter is disabled.
+    if ([clazz isEnabled] && !self.isEnabled) {
+      self.enabledStateUpdating = YES;
```
jaelim-ms: @TroubleMakerBen @clpolet Do we really need this flag? I wonder if we synchronize `setEnabled` and get rid of this flag.

jaelim-ms: @MatkovIvan Do you need to change this flag? You are not disabling the core services here.

jaelim-ms:
@MatkovIvan Please resolve the conflict.

AppCenter/AppCenter/MSAppCenter.m
MatkovIvan: @jaelim-ms This is caused by https://github.com/Microsoft/AppCenter-SDK-Apple/blob/9449f274db22f9d5c3777a3c749b4b1ff0afdf53/AppCenter/AppCenter/MSServiceAbstract.m#L88

jaelim-ms: Gotcha

PR #886, opened 2017-12-19
Sending start services log after enable

MatkovIvan:


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -16,13 +16,11 @@
 #import "MSCustomPropertiesPrivate.h"
 #endif
 
-// Singleton
+// Singleton.
```
jaelim-ms: This is a comment for **definition** of variables. Should use `/** */`

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -16,13 +16,11 @@
...
 static MSAppCenter *sharedInstance = nil;
 static dispatch_once_t onceToken;
 
-/**
- * Base URL for HTTP Ingestion backend API calls.
- */
+// Base URL for HTTP Ingestion backend API calls.
```
jaelim-ms: Revert the change due to above comment.

AppCenter/AppCenter/MSAppCenter.m
jaelim-ms: Please change the other two below. The scope for the change is remaining only on this file. We can do others with any changes of files.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -313,11 +311,11 @@ - (void)setEnabled:(BOOL)isEnabled {
   self.enabledStateUpdating = YES;
   if ([self isEnabled] != isEnabled) {
 
-    // Enable/disable pipeline.
```
jaelim-ms: Any intentions of this change?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -418,15 +422,18 @@ - (void)sendStartServiceLog:(NSArray<NSString *> *)servicesNames {
...
+    }
+    [self.startedServicesNamesToLog addObjectsFromArray:servicesNames];
   }
 }
 
 #if !TARGET_OS_TV
 - (void)sendCustomPropertiesLog:(NSDictionary<NSString *, NSObject *> *)properties {
   MSCustomPropertiesLog *customPropertiesLog = [MSCustomPropertiesLog new];
   customPropertiesLog.properties = properties;
-
```
jaelim-ms: Thanks for the cleanup!

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -418,15 +422,18 @@ - (void)sendStartServiceLog:(NSArray<NSString *> *)servicesNames {
     MSStartServiceLog *serviceLog = [MSStartServiceLog new];
     serviceLog.services = servicesNames;
     [self.logManager processLog:serviceLog forGroupId:kMSGroupId];
+  } else {
+    if (self.startedServicesNamesToLog == nil) {
+      self.startedServicesNamesToLog = [NSMutableArray new];
+    }
+    [self.startedServicesNamesToLog addObjectsFromArray:servicesNames];
```
jaelim-ms: Can we just point servicesNames instead of creating an array?

AppCenter/AppCenter/Internals/MSAppCenterInternal.h
```diff
@@ -19,6 +19,7 @@ static NSString *const kMSDisableAll = @"All";
 
 @property(nonatomic) id<MSLogManager> logManager;
 @property(nonatomic) NSMutableArray<NSObject<MSServiceInternal> *> *services;
+@property(nonatomic) NSMutableArray<NSString *> * startedServicesNamesToLog;
```
jaelim-ms: Just `startedServiceNames`?

AppCenter/AppCenter/MSAppCenter.m
MatkovIvan: Nope. I've added a test for this case. See https://github.com/Microsoft/AppCenter-SDK-Apple/pull/886/commits/1986a03397e878983645079b4d9f22e8d27b775f

AppCenter/AppCenter/MSAppCenter.m
MatkovIvan: Because `isEnabled` is checked in `applyPipelineEnabledState` (`sendStartServiceLog`)

jaelim-ms: The `isEnabled` is a parameter passed to `applyPipelineEnabledState`. It doesn't read anything in NSUserDefaults. I think we should make sure the pipeline is properly changed before setting the value to NSUserDefaults.

jaelim-ms: Changing this is not a big deal but I would like to keep the previous code to get rid of a chance for side effect. Please revert unless it has to be changed.

jaelim-ms: OK I got it. `isEnabled` from `sendStartServiceLog`.

PR #888, opened 2018-01-02
Fix/session

jaelim-ms:


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -202,6 +203,10 @@ - (BOOL)configure:(NSString *)appSecret {
       if ((![MSLogger isUserDefinedLogLevel]) && ([MSUtility currentAppEnvironment] == MSEnvironmentOther)) {
         [MSAppCenter setLogLevel:MSLogLevelWarning];
       }
+
+      // Initialize session.
+      [MSSessionContext setSessionId:nil];
```
TroubleMakerBen: I find this initialization rather unintuitive. I'd never have guessed without the comment that this actually initialized the `MSSessionContext`. It's not a merge blocker in my opinion, but I'd prefer more obvious way to init the session(context).

jaelim-ms: Yeah.. Good point. I'll add FIXME comment there.

jaelim-ms:
Flaky unit test failures. Trigger a CI build again.

PR #891, opened 2018-01-05
Fix/recoverable errors suspend but not fatal

TroubleMakerBen:
Previously, we would consider recoverable errors "fatal" (which would delete them) if we used all retries on them. Now, we simply suspend the sender. On the next launch, the SDK will try to send the events.


AppCenter/AppCenter/Internals/Sender/MSSenderCall.m
```diff
@@ -81,10 +82,20 @@ - (void)sender:(id<MSSender>)sender
...
-  // Retry.
-  else if ([MSSenderUtil isRecoverableError:statusCode] && ![self hasReachedMaxRetries]) {
-    [self startRetryTimerWithStatusCode:statusCode];
+  else if ([MSSenderUtil isRecoverableError:statusCode]) {
+    
+    // Reset the retry count, will retry once app goes to foreground again.
+    if ([self hasReachedMaxRetries]) {
+      [self resetRetry];
+      NSString *logMessage = @"Received recoverable error but reached max. number of retries.";
+      MSLogInfo([MSAppCenter logTag], logMessage);
```
jaelim-ms: Any reason of using variable instead of string constant itself?

AppCenter/AppCenter/Internals/Sender/MSSenderCall.m
```diff
@@ -81,10 +82,20 @@ - (void)sender:(id<MSSender>)sender
...
+      [sender suspend];
```
jaelim-ms: I think we should do more than this. I think we should remove pending batch ID and pending call ID, disable it instead of just suspending by calling `[self setEnabled:NO addDeleteDataOnDisabled: NO]`

jaelim-ms:
Trigger a CI build again to get a right code coverage report.

TroubleMakerBen:
@jaelim-ms we should talk about 586cc3886e8bd177e353c1fb31f8672c4be5e7c5 and if we want this in this PR or not.

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -276,7 +267,27 @@ - (void)call:(MSSenderCall *)call completedWithFatalError:(BOOL)fatalError {
...
+
+      // Disable and do not delete data. Do not notify the delegates as this will cause data to be deleted.
+      [self setEnabled:NO andDeleteDataOnDisabled:NO];
+      break;
+    case MSSenderCallResultSuccess:
+      break;
     }
+
+    // Remove call from pending call. This needs to happen after calling setEnabled:andDeleteDataOnDisabled:
+    // FIXME: Refactor ependency between calling setEnabled:andDeleteDataOnDisabled: and suspending the sender.
```
jaelim-ms: typo? `ependency`

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -254,19 +254,10 @@ - (void)sendCallAsync:(MSSenderCall *)call {
   }
 }
 
-- (void)call:(MSSenderCall *)call completedWithFatalError:(BOOL)fatalError {
+- (void)call:(MSSenderCall *)call completedWithResult:(MSSenderCallResult)result {
   @synchronized(self) {
-    NSString *callId = call.callId;
```
jaelim-ms: The deleted block was executed even for fatal error but your change doesn't. Any intentions?

AppCenter/AppCenter/Internals/Sender/MSSenderCallDelegate.h
```diff
@@ -1,4 +1,5 @@
 #import <Foundation/Foundation.h>
+#import "MSConstants+Internal.h"
```
jaelim-ms: New line between `#import <>` and `#import ""`

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
TroubleMakerBen: The deleted call gets executed for all calls but at the end of this method. See [here](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/891/files/ffee0191010ee051b126548a49726e26a518a61f#diff-7de131038a5e31d3fa3c7bf3557d1ef7R281).

PR #895, opened 2018-01-09
Fix flaky unit tests

jaelim-ms:


AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
```diff
@@ -248,7 +248,7 @@ - (void)testAnonymizedPathWorks {
              @"Containers/Bundle/Application/8CC7B5B5-7841-45C4-BAC2-6AA1B944A5E1/Puppet.app/Puppet";
   actual = [MSErrorLogFormatter anonymizedPathFromPath:testPath];
   assertThat(actual, equalTo(expected));
-  XCTAssertFalse([actual containsString:@"someone"]);
+  XCTAssertFalse([actual hasPrefix:@"/Users/someone"]);
```
TroubleMakerBen: way better!

AppCenter/AppCenterTests/MSIngestionSenderTests.m
```diff
@@ -298,28 +291,24 @@ - (void)testTasksRunningOnSenderResumed {
...
-          if (!testFinished) {
-            XCTFail(@"Completion handler shouldn't be called as test will finish before the response timeout.");
-          }
-        }
-      }];
+        }];
+  }
+
+  // Make sure all log containers are enqueued before suspending sender.
+  [NSThread sleepForTimeInterval:0.5];
```
TroubleMakerBen: Kinda ugly but also more readable. The things we have to do to have more reliable unit tests. ;)

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
jaelim-ms: Yeah.. `hasPrefix:` is available macOS 10.10 and above. This will eliminate compile warning.

AppCenter/AppCenterTests/MSIngestionSenderTests.m
jaelim-ms: As mentioned offline, this is not a real fix for the issue. I doubt it might be caused by at least one log that was enqueued right before suspend which wasn't actually processed yet. Based on Apple document, initial state value is NSURLSessionTaskStateSuspended and it will be changed upon task process. I just want to make sure all logs are enqueued before suspending. Will check if the test still fails in the future.

PR #905, opened 2018-01-16
Add tests for cxx exception handler

MatkovIvan:


AppCenterCrashes/AppCenterCrashesTests/MSCrashesCXXExceptionTests.mm
```diff
@@ -1,11 +1,30 @@
 #import "MSCrashesCXXExceptionHandler.h"
 #import "MSCrashesCXXExceptionWrapperException.h"
 #import "MSTestFrameworks.h"
+#import <exception>
```
jaelim-ms: Please reorder. `#import <>` andthen `#import ""` with empty line in between them.

jaelim-ms: Will merge without my change request since it is a formatting issue and we can address it later.

PR #912, opened 2018-01-26
Fix application state check for crashes processing

MatkovIvan:


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -685,11 +685,10 @@ - (void)startCrashProcessing {
    * counterintuitive, this is important because there are scenarios in some wrappers (i.e. RN) where
    * the application state is not ready by the time crash processing needs to happen.
    */
-  if (self.automaticProcessing &&
-      ([MSUtility applicationState] != MSApplicationStateActive &&
-       [MSUtility applicationState] != MSApplicationStateUnknown)) {
-        return;
-      }
+  if (self.automaticProcessing && [MSUtility applicationState] == MSApplicationStateBackground) {
+    MSLogWarning([MSCrashes logTag], @"Crashes processing is closed because the application is in the background");
```
achocron: Maybe change message to "Crashes will not be processed because the application is in the background."

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -685,11 +685,10 @@ - (void)startCrashProcessing {
    * counterintuitive, this is important because there are scenarios in some wrappers (i.e. RN) where
    * the application state is not ready by the time crash processing needs to happen.
    */
-  if (self.automaticProcessing &&
-      ([MSUtility applicationState] != MSApplicationStateActive &&
-       [MSUtility applicationState] != MSApplicationStateUnknown)) {
-        return;
-      }
+  if (self.automaticProcessing && [MSUtility applicationState] == MSApplicationStateBackground) {
```
achocron: The comment indicates that there is no lifecycle for app extensions, so i guess if we remove the check for the "unknown" state, crash processing will not work for app extensions.

MatkovIvan: I changed `!=` to `==`

achocron: Yes disregard my comment

PR #922, opened 2018-01-30
Fix UI tests

MatkovIvan:
- Actualize existing tests (broken after UI changes);
- Simplify tests code using helpers and setting identifiers;
- Added test for settings persistence (failing on iOS 10 and later);
- Added simple test for custom properties;

PS I faced with [XCode 8.3 bug](https://forums.developer.apple.com/thread/75546), on other versions should work fine.


TroubleMakerBen:
I'm perfectly fine making these changes.
`Added test for settings persistence (failing on iOS 10 and later);` gives me a really uneasy feeling. I know it doesn't happen on actual devices, but it's not cool. =( Did we file a rdar for this (or dupe existing ones)?

PR #928, opened 2018-02-07
Feature/tester app in app updates

JessicaYeh:
The related Android PR is here: https://github.com/Microsoft/AppCenter-SDK-Android/pull/613

> Apps installed using the native tester app currently aren't able to successfully enable in-app updates, because the cookies stored on install aren't there. This is part of the work to fix that.
> 
> When enabling in-app updates, the Android SDK should check if an app with the ms-actesterapp:// scheme (belongs to the App Center tester app) is installed. If it is, then it should try to enable in-app updates by redirecting to that scheme instead of the update-setup page on the install portal. If for some reason enabling in-app updates fails, then the SDK should fall back to attempting to enable in-app updates using the update-setup page. This could happen if someone has the native tester app installed, but for some reason they still install an app from the install portal. The user will see two attempts to enable in-app updates in this case, but hopefully this is rare. We will track telemetry for how often this happens.

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
@@ -248,7 +248,7 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
     dispatch_async(dispatch_get_main_queue(), ^{
       NSURL *url;
       
-      BOOL shouldUseTesterAppForUpdateSetup = [MS_USER_DEFAULTS objectForKey:kMSTesterAppUpdateSetupFailedKey] != NULL;
+      BOOL shouldUseTesterAppForUpdateSetup = [MS_USER_DEFAULTS objectForKey:kMSTesterAppUpdateSetupFailedKey] == NULL;
       BOOL testerAppOpened = NO;
       if (shouldUseTesterAppForUpdateSetup) {
         // Attempt to open the native iOS tester app to enable in-app updates
--- a/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
@@ -244,10 +244,6 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
 
     MSLogInfo([MSDistribute logTag], @"Request information of initial installation.");
     
-    // Use swizzling to get [UIApplication sharedApplication]
-    SEL sharedAppSel = NSSelectorFromString(@"sharedApplication");
-    UIApplication* sharedApp = ((UIApplication * (*)(id, SEL))[[UIApplication class] methodForSelector:sharedAppSel])([UIApplication class], sharedAppSel);
-    
     // Don't run on the UI thread, or else the app may be slow to startup
     NSURL *testerAppUrl = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash isTesterApp:true];
     NSURL *installUrl = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash isTesterApp:false];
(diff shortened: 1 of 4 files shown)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -228,17 +234,41 @@ - (void)requestInstallInformationWith:(NSString *)releaseHash {
...
     MSLogInfo([MSDistribute logTag], @"Request information of initial installation.");
-
-    // Most failures here require an app update. Thus, it will be retried only on next App instance.
-    url = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash];
-    if (url) {
-      [self openUrlInAuthenticationSessionOrSafari:url];
-    }
+    
+    // Use swizzling to get [UIApplication sharedApplication]
+    SEL sharedAppSel = NSSelectorFromString(@"sharedApplication");
```
TroubleMakerBen: `MSUtility+Application` already contains logic for this. Why don't we expose the method so we can use that across the app?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -228,17 +234,41 @@ - (void)requestInstallInformationWith:(NSString *)releaseHash {
...
-    NSURL *url;
     MSLogInfo([MSDistribute logTag], @"Request information of initial installation.");
-
-    // Most failures here require an app update. Thus, it will be retried only on next App instance.
-    url = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash];
-    if (url) {
-      [self openUrlInAuthenticationSessionOrSafari:url];
-    }
+    
+    // Use swizzling to get [UIApplication sharedApplication]
```
TroubleMakerBen: That's actually not swizzling, can you please update the comment?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -228,17 +234,41 @@ - (void)requestInstallInformationWith:(NSString *)releaseHash {
...
+    
+    // Don't run on the UI thread, or else the app may be slow to startup
+    NSURL *testerAppUrl = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash isTesterApp:true];
+    NSURL *installUrl = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash isTesterApp:false];
+    dispatch_async(dispatch_get_main_queue(), ^{
+      
+      BOOL shouldUseTesterAppForUpdateSetup = [MS_USER_DEFAULTS objectForKey:kMSTesterAppUpdateSetupFailedKey] == NULL;
+      BOOL testerAppOpened = NO;
+      if (shouldUseTesterAppForUpdateSetup) {
+        // Attempt to open the native iOS tester app to enable in-app updates
```
TroubleMakerBen: Please add an empty line before the comment.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -228,17 +234,36 @@ - (void)requestInstallInformationWith:(NSString *)releaseHash {
...
+    NSURL *installUrl = [self buildTokenRequestURLWithAppSecret:self.appSecret releaseHash:releaseHash isTesterApp:false];
+    dispatch_async(dispatch_get_main_queue(), ^{
+      BOOL shouldUseTesterAppForUpdateSetup = [MS_USER_DEFAULTS objectForKey:kMSTesterAppUpdateSetupFailedKey] == NULL;
+      BOOL testerAppOpened = NO;
+      if (shouldUseTesterAppForUpdateSetup) {
+        MSLogInfo([MSDistribute logTag], @"Attempting to use tester app for update setup.");
+        
+        // Attempt to open the native iOS tester app to enable in-app updates
+        if (testerAppUrl) {
+          testerAppOpened = [self openUrlUsingSharedApp:testerAppUrl];
```
jaelim-ms: Let's log `testerAppOpened` value as Info log in the `if` statement.
We should be able to know what happened in Distribute process from logs, this will say that it found tester app, tried open the app, and succeeded or failed to do it.

TroubleMakerBen: Very good point!

TroubleMakerBen: I just saw that you addressed this, Yess!

PR #932, opened 2018-02-13
Fix reporting of in-app update for IPA with several architectures

nevalenny:
The PR addresses two issues:
1. When there are several architectures in one IPA, backend reports an array with release hashes (one for each architecture). We were saving only the first one, and in case it doesn't match the one installed, no reporting for in-app update took place.
2. The release hashes for downloaded release were saved too late so sometimes the app was closed before the NSUserDefaults storage was updated, which also leads to missing reports.

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
@@ -34,6 +34,11 @@ static NSString *const kMSGroupId = @"Distribute";
  */
 static NSString *const kMSUpdateTokenApiPathFormat = @"/apps/%@/update-setup";
 
+/**
+ * The tester app path for update token request.
+ */
+static NSString *const kMSTesterAppUpdateTokenPath = @"ms-actesterapp://update-setup";
+
 #pragma mark - Error constants
 
 static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid update token URL:%@";
@@ -74,6 +79,10 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
                                    name:UIApplicationWillEnterForegroundNotification
                                  object:nil];
   }
+
+  // Init the distribute info tracker.
+  _distributeInfoTracker = [[MSDistributeInfoTracker alloc] init];
+
   return self;
 }
 
@@ -100,6 +109,12 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   return kMSGroupId;
 }
 
+- (MSInitializationPriority)initializationPriority {
+
+  // Initialize Distribute before Analytics to add distributionGroupId field to the first startSession event after app start.
+  return MSInitializationPriorityHigh;
+}
+
 #pragma mark - MSServiceAbstract
 
 - (void)applyEnabledState:(BOOL)isEnabled {
@@ -111,13 +126,20 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
     self.releaseDetails = nil;
     [self startUpdate];
     [MSAppDelegateForwarder addDelegate:self.appDelegate];
+
+    // Enable the distribute info tracker.
+    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+    [self.distributeInfoTracker updateDistributionGroupId:distributionGroupId];
+    [self.channelGroup addDelegate:self.distributeInfoTracker];
   } else {
     [self dismissEmbeddedSafari];
+    [self.channelGroup removeDelegate:self.distributeInfoTracker];
     [MSAppDelegateForwarder removeDelegate:self.appDelegate];
     [MS_USER_DEFAULTS removeObjectForKey:kMSUpdateTokenRequestIdKey];
     [MS_USER_DEFAULTS removeObjectForKey:kMSPostponedTimestampKey];
     [MS_USER_DEFAULTS removeObjectForKey:kMSMandatoryReleaseKey];
     [MS_USER_DEFAULTS removeObjectForKey:kMSUpdateSetupFailedPackageHashKey];
+    [MS_USER_DEFAULTS removeObjectForKey:kMSTesterAppUpdateSetupFailedKey];
     MSLogInfo([MSDistribute logTag], @"Distribute service has been disabled.");
   }
 }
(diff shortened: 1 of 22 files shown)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -151,26 +148,28 @@ - (void)notifyUpdateAction:(MSUpdateAction)action {
   }
   switch (action) {
   case MSUpdateActionUpdate:
-#if TARGET_OS_SIMULATOR
 
-    /*
+    if ([self isEnabled]) {
+      MSLogDebug([MSDistribute logTag], @"'Update now' is selected. Start download and install the update.");
+      // Store details to report new download after restart if this release is installed.
```
MatkovIvan: Nitpick: Empty line before the comment

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -151,26 +148,28 @@ - (void)notifyUpdateAction:(MSUpdateAction)action {
...
+      [self storeDownloadedReleaseDetails:self.releaseDetails];
+#if TARGET_OS_SIMULATOR
+     /*
```
MatkovIvan: Nitpick: Empty line before the comment

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
nevalenny: Thanks, fixed.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
nevalenny: Here also, and several other places.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -749,13 +759,13 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
 
   // Check if we need to report release installation.
   NSString *lastDownloadedReleaseHash = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
-  if (!lastDownloadedReleaseHash) {
+  if (lastDownloadedReleaseHash == nil) {
```
TroubleMakerBen: I like that. `!` is so easy to miss.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -151,26 +148,30 @@ - (void)notifyUpdateAction:(MSUpdateAction)action {
...
+
+    if ([self isEnabled]) {
+      MSLogDebug([MSDistribute logTag], @"'Update now' is selected. Start download and install the update.");
+
+      // Store details to report new download after restart if this release is installed.
+      [self storeDownloadedReleaseDetails:self.releaseDetails];
 #if TARGET_OS_SIMULATOR
 
-    /*
+     /*
```
jaelim-ms: Remove a space.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -720,25 +722,33 @@ - (void)storeDownloadedReleaseDetails:(nullable MSReleaseDetails *)details {
     return;
   }
 
-  MSLogDebug([MSDistribute logTag], @"Store downloaded release hash and id for later reporting.");
   NSNumber *releaseId = details.id;
-  NSString *releaseHash = [details.packageHashes objectAtIndex:0];
+
+  // IPA can contain several hashes, each for different architecture and we can't predict which will be installed,
```
jaelim-ms: Use /* */ for multi line comment.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
jaelim-ms: I like this one better. It is a trade-off but it will reduce one bit operation.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -749,13 +759,13 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
   NSString *lastDownloadedReleaseHash = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
-  if (!lastDownloadedReleaseHash) {
+  if (lastDownloadedReleaseHash == nil) {
     MSLogDebug([MSDistribute logTag], @"Current release was already reported, skip reporting.");
     return nil;
   }
 
   // Skip if downloaded release not installed yet.
-  if (![lastDownloadedReleaseHash isEqualToString:currentInstalledReleaseHash]) {
+  if (![lastDownloadedReleaseHash containsString:currentInstalledReleaseHash]) {
```
jaelim-ms: Not a big deal but any reasons for the change?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -720,25 +722,33 @@ - (void)storeDownloadedReleaseDetails:(nullable MSReleaseDetails *)details {
...
 }
 
 - (void)removeDownloadedReleaseDetailsIfUpdated:(NSString *)currentInstalledReleaseHash {
    NSString *lastDownloadedReleaseHash = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
    if (lastDownloadedReleaseHash == nil) {
      return;
    }
-   if (![lastDownloadedReleaseHash isEqualToString:currentInstalledReleaseHash]) {
-     MSLogDebug([MSDistribute logTag], @"Stored release hash doesn't match current installation, probably downloaded but not installed yet, keep in store.");
+   if (![lastDownloadedReleaseHash containsString:currentInstalledReleaseHash]) {
```
MatkovIvan: Available on macOS only 10.10+
See https://developer.apple.com/documentation/foundation/nsstring/1414563-containsstring

Use `rangeOfString` instead

MatkovIvan: Available on macOS only 10.10+
See https://developer.apple.com/documentation/foundation/nsstring/1414563-containsstring

Use `rangeOfString` instead

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
nevalenny: The reason is that `lastDownloadedReleaseHash` can be a comma separated string of several hashes. Will rename the variable to reflect it.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
nevalenny: Nice catch, thanks!

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
nevalenny: Thanks for the review, fixed

PR #933, opened 2018-02-14
Update version number to 1.3.1

nevalenny:


Config/Version.xcconfig
```diff
@@ -1,2 +1,2 @@
 BUILD_NUMBER = 1
-VERSION_STRING = 1.3.0
+VERSION_STRING = 1.3.1
```
TroubleMakerBen: Please don't bump this. If we were to do that, it'd be 2 (and not 1.3.1).

jaelim-ms: Looks like we missed this version bump after the previous release. Valid one.
Don't forget to bump up this version and all verisons in .jazzy.xml after release.

PR #939, opened 2018-02-23
Add distribution tracker to public channel even  if distribution Id is not stored

meiravf:
Distribution tracker should always be added to the group channel even if the distribution group Id is empty. 

If not doing so when installing the app from scratch (removing the previous version or installing on the first time) the stored  distribution Id is empty which results in not adding the tracker into the public channels . Afterwards the browser is opened and we retrieve the distribution Id but its too late the tracker is not part of  the the public channels and the group Id is not logged.

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
@@ -130,10 +130,11 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
     // Store distributionGroupId in distributeInfoTracker
     NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
     if (distributionGroupId) {
-      MSLogDebug([MSDistribute logTag],@"Successfully retrieved distribution group Id setting it in distributeInfoTracker.");
+      MSLogDebug([MSDistribute logTag],
+                 @"Successfully retrieved distribution group Id setting it in distributeInfoTracker.");
       [self.distributeInfoTracker updateDistributionGroupId:distributionGroupId];
-      }
-    } else {
+    }
+  } else {
     [self dismissEmbeddedSafari];
     [self.channelGroup removeDelegate:self.distributeInfoTracker];
     [MSAppDelegateForwarder removeDelegate:self.appDelegate];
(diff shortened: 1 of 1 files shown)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -108,7 +108,8 @@ - (NSString *)groupId {
 
 - (MSInitializationPriority)initializationPriority {
 
-  // Initialize Distribute before Analytics to add distributionGroupId field to the first startSession event after app start.
+  // Initialize Distribute before Analytics to add distributionGroupId field to the first startSession event after app
```
jaelim-ms: Can you make it single line? Or use /* */ for multiline comment.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -125,10 +126,15 @@ - (void)applyEnabledState:(BOOL)isEnabled {
...
 
     // Enable the distribute info tracker.
-    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
-    [self.distributeInfoTracker updateDistributionGroupId:distributionGroupId];
     [self.channelGroup addDelegate:self.distributeInfoTracker];
-  } else {
+
+    // Update distribute info tracker with stored distributionGroupId
+    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+    if (distributionGroupId) {
```
jaelim-ms: Please reformat this if statement. Hard to read it.

TroubleMakerBen: I agree.

meiravf:
@jaelim-ms I have updated the title of the PR, you are right the `term` enable is wrong .
Its actually adding the `distributeInfoTracker` to the `publicChannel`.

meiravf:
@jaelim-ms Yes we have done e2e on iOS, I thought that macOS is not supported with in-app updates.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -125,10 +125,15 @@ - (void)applyEnabledState:(BOOL)isEnabled {
...
 
     // Enable the distribute info tracker.
-    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
-    [self.distributeInfoTracker updateDistributionGroupId:distributionGroupId];
     [self.channelGroup addDelegate:self.distributeInfoTracker];
-  } else {
+
+    // Store distributionGroupId in distributeInfoTracker
+    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+    if (distributionGroupId) {
```
jaelim-ms: Please format this again. The indentation for `}` seems wrong in this `if` statement.

meiravf:
@jaelim-ms done I have re-formatted the file.

meiravf:
@jaelim-ms I see that the formatting didn't work well. fixing..

PR #940, opened 2018-02-23
Update distribution group ID if a new release with different group ID was in-app updated

roland-chernov-akvelon:
This PR implements updating of distribution group ID if in-app updated release has different distribution group.

Group ID may change if one user is added to different distribution groups and a new release was distributed to another group.

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSDistribute.m
@@ -120,7 +120,6 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   // Enabling
   if (isEnabled) {
     MSLogInfo([MSDistribute logTag], @"Distribute service has been enabled.");
-    [self changeDistributionGroupIdAfterAppUpdateIfNeeded];
     self.releaseDetails = nil;
     [self startUpdate];
     [MSAppDelegateForwarder addDelegate:self.appDelegate];
@@ -211,10 +210,11 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
 #pragma mark - Private
 
 - (void)startUpdate {
-  NSString *updateToken = [MSKeychainUtil stringForKey:kMSUpdateTokenKey];
-  NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
   NSString *releaseHash = MSPackageHash();
   if (releaseHash) {
+    [self changeDistributionGroupIdAfterAppUpdateIfNeeded:releaseHash];
+    NSString *updateToken = [MSKeychainUtil stringForKey:kMSUpdateTokenKey];
+    NSString *distributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
     if (updateToken || distributionGroupId) {
       [self checkLatestRelease:updateToken distributionGroupId:distributionGroupId releaseHash:releaseHash];
     } else {
@@ -319,7 +319,7 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
     if (self.sender == nil) {
       NSMutableDictionary *queryStrings = [[NSMutableDictionary alloc] init];
       NSMutableDictionary *reportingParametersForUpdatedRelease =
-          [self getReportingParametersForUpdatedRelease:updateToken distributionGroupId:distributionGroupId];
+          [self getReportingParametersForUpdatedRelease:updateToken currentInstalledReleaseHash:releaseHash distributionGroupId:distributionGroupId];
       if (reportingParametersForUpdatedRelease != nil) {
         [queryStrings addEntriesFromDictionary:reportingParametersForUpdatedRelease];
       }
@@ -366,7 +366,7 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
                } else {
 
                  // Check if downloaded release was installed and remove stored release details.
-                 [self removeDownloadedReleaseDetailsIfUpdated];
+                 [self removeDownloadedReleaseDetailsIfUpdated:releaseHash];
 
                  /*
                   * Handle this update.
@@ -739,24 +739,26 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   [MS_USER_DEFAULTS setObject:releaseHashes forKey:kMSDownloadedReleaseHashKey];
 }
 
-- (void)removeDownloadedReleaseDetailsIfUpdated {
+- (void)removeDownloadedReleaseDetailsIfUpdated:(NSString *)currentInstalledReleaseHash {
   NSString *lastDownloadedReleaseHashes = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
   if (lastDownloadedReleaseHashes == nil) {
     return;
   }
-  if ([self isCurrentReleaseWasUpdated:lastDownloadedReleaseHashes] == NO) {
-    MSLogDebug([MSDistribute logTag], @"Stored release hash(es) doesn't match current installation hash, probably downloaded but not installed yet, keep in store.");
+  if ([lastDownloadedReleaseHashes rangeOfString:currentInstalledReleaseHash].location == NSNotFound) {
+    MSLogDebug([MSDistribute logTag], @"Stored release hash(es) (%@) doesn't match current installation hash (%@), probably downloaded but not installed yet, keep in store.",
+            lastDownloadedReleaseHashes, currentInstalledReleaseHash);
     return;
   }
 
   // Successfully reported, remove downloaded release details.
-  MSLogDebug([MSDistribute logTag], @"Successfully reported app update for downloaded release hash(es) (%@), removing from store.",
-             lastDownloadedReleaseHashes);
+  MSLogDebug([MSDistribute logTag], @"Successfully reported app update for downloaded release hash (%@), removing from store.",
+          currentInstalledReleaseHash);
   [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedReleaseIdKey];
   [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedReleaseHashKey];
 }
 
 - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)updateToken
+                                     currentInstalledReleaseHash:(NSString *)currentInstalledReleaseHash
                                              distributionGroupId:(NSString *)distributionGroupId {
 
   // Check if we need to report release installation.
@@ -767,7 +769,7 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   }
 
   // Skip if downloaded release not installed yet.
-  if ([self isCurrentReleaseWasUpdated:lastDownloadedReleaseHashes] == NO) {
+  if ([lastDownloadedReleaseHashes rangeOfString:currentInstalledReleaseHash].location == NSNotFound) {
     MSLogDebug([MSDistribute logTag], @"New release was downloaded but not installed yet, skip reporting.");
     return nil;
   }
@@ -785,12 +787,16 @@ static NSString *const kMSUpdateTokenURLInvalidErrorDescFormat = @"Invalid updat
   return reportingParameters;
 }
 
-- (void)changeDistributionGroupIdAfterAppUpdateIfNeeded {
+- (void)changeDistributionGroupIdAfterAppUpdateIfNeeded:(NSString *)currentInstalledReleaseHash {
   NSString *lastDownloadedReleaseHashes = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
   NSString *lastDownloadedDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDownloadedDistributionGroupIdKey];
-  if ((lastDownloadedDistributionGroupId == nil) || ([self isCurrentReleaseWasUpdated:lastDownloadedReleaseHashes] == NO)) {
+
+  // Skip if the current release was not updated.
+  if ((lastDownloadedDistributionGroupId == nil) || ([lastDownloadedReleaseHashes rangeOfString:currentInstalledReleaseHash].location == NSNotFound)) {
     return;
   }
+
+  // Skip if the group ID of an updated release is the same as the current one.
   NSString *currentDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
   if ((currentDistributionGroupId != nil) && ([lastDownloadedDistributionGroupId isEqualToString:currentDistributionGroupId])) {
     return;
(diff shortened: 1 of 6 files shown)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -717,46 +718,45 @@ - (BOOL)isNewerVersion:(MSReleaseDetails *)details {
...
-             lastDownloadedReleaseHashes, currentInstalledReleaseHash);
-     return;
-   }
-
-   // Successfully reported, remove downloaded release details.
-   MSLogDebug([MSDistribute logTag], @"Successfully reported app update for downloaded release hash (%@), removing from store.",
-           currentInstalledReleaseHash);
-   [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedReleaseIdKey];
-   [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedReleaseHashKey];
+- (void)removeDownloadedReleaseDetailsIfUpdated {
```
roland-chernov-akvelon: Fixed indentation (3 spaces was used).

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -767,7 +767,7 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
   }
 
   // Skip if downloaded release not installed yet.
-  if ([lastDownloadedReleaseHashes rangeOfString:currentInstalledReleaseHash].location == NSNotFound) {
+  if (![self isCurrentReleaseWasUpdated:lastDownloadedReleaseHashes]) {
```
TroubleMakerBen: I'd like us moving away from `if(!...)` as it's easy to overlook the bang. cc @jaelim-ms

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -785,6 +785,33 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
 }
 
+- (void)changeDistributionGroupIdAfterAppUpdateIfNeeded {
+  NSString *lastDownloadedReleaseHashes = [MS_USER_DEFAULTS objectForKey:kMSDownloadedReleaseHashKey];
+  NSString *lastDownloadedDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDownloadedDistributionGroupIdKey];
+  if (lastDownloadedDistributionGroupId == nil || ![self isCurrentReleaseWasUpdated:lastDownloadedReleaseHashes]) {
+    return;
+  }
+  NSString *currentDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+  if (currentDistributionGroupId != nil && [lastDownloadedDistributionGroupId isEqualToString:currentDistributionGroupId]) {
```
TroubleMakerBen: Please surround chained conditions with `()`. I know this seams like a detail but as we have more people working on the code base, I'd like us to be more careful about clean code and formatting, etc.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -225,32 +230,45 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
 - (BOOL)handleUpdate:(MSReleaseDetails *)details;
 
 /**
- * Store details about downloaded release to report it after installation.
+ * Store details about downloaded release.
```
TroubleMakerBen: Just voicing some thoughts here: Maybe it's just me but my expectation for a comment in a header is that it's a correct sentence with articles, etc.
Do we want to do that in the future?

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -876,7 +876,7 @@ - (void)testCheckLatestReleaseOnRecoverableError {
   [self.settingsMock setObject:@1 forKey:kMSUpdateTokenRequestIdKey];
   [self.settingsMock setObject:@1 forKey:kMSPostponedTimestampKey];
   [self.settingsMock setObject:@1 forKey:kMSDistributionGroupIdKey];
-  [self.sut checkLatestRelease:@"token" distributionGroupId:@"groupId" releaseHash:@"releaseHash"];
+  [self.sut checkLatestRelease:@"token" distributionGroupId:@"groupId" releaseHash:kMSTestReleaseHash];
```
TroubleMakerBen: why stop at introducing the constant for the release hash string, let's add a constant for `groupdID`, too?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: Sure. Does this looks good to you?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: Fixed. Thank you.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
roland-chernov-akvelon: Sorry, my grammar isn't good. Does this look good now?

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
roland-chernov-akvelon: Added. Thank you.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -318,7 +319,7 @@ - (void)checkLatestRelease:(NSString *)updateToken
     if (self.sender == nil) {
       NSMutableDictionary *queryStrings = [[NSMutableDictionary alloc] init];
       NSMutableDictionary *reportingParametersForUpdatedRelease =
-          [self getReportingParametersForUpdatedRelease:updateToken currentInstalledReleaseHash:releaseHash distributionGroupId:distributionGroupId];
```
jaelim-ms: I can see that you remove current release hash from some methods parameters. Can you tell me why?
You are calling `MSPackageHash()` instead of passing it around via method parameter?
Just curious.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -717,46 +718,45 @@ - (BOOL)isNewerVersion:(MSReleaseDetails *)details {
...
   /*
    * IPA can contain several hashes, each for different architecture and we can't predict which will be installed,
    * so save all hashes as comma separated string.
    */
-  NSString *releaseHashes = [details.packageHashes count] > 1
+  NSString *releaseHashes = [details.packageHashes count] > 0
           ? [details.packageHashes componentsJoinedByString:@","]
-          : details.packageHashes[0];
+          : nil;
+  MSLogDebug([MSDistribute logTag], @"Store downloaded release details, group id: %@, release hashes: %@, release id: %@.",
```
jaelim-ms: Can you move this log back to the original place? This log will not tell us that all values are properly stored to `NSUserDefaults`.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -785,6 +785,33 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
+    return;
+  }
+  NSString *currentDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+  if ((currentDistributionGroupId != nil) && ([lastDownloadedDistributionGroupId isEqualToString:currentDistributionGroupId])) {
+    return;
+  }
+
+  // Set group ID from downloaded release details.
+  MSLogDebug([MSDistribute logTag], @"Current group ID doesn't match the group ID of downloaded release, updating current group id: %@", lastDownloadedDistributionGroupId);
+  [MS_USER_DEFAULTS setObject:lastDownloadedDistributionGroupId forKey:kMSDistributionGroupIdKey];
```
jaelim-ms: Why do you set `setObject:lastDownloadedDistributionGroupId` here instead of `currentDistributionGroupId`?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -785,6 +785,33 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
+
+  // Remove saved downloaded group ID.
+  [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedDistributionGroupIdKey];
```
jaelim-ms: Is this temporary used in a same session and remove it? I don't understand why you remove this object.
I feel like you put a value to `NSUserDefaults` and use it in other methods. If this is the case, please do not rely on `NSUserDefaults` and pass the value as a parameter.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -785,6 +785,33 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
+}
+
+- (bool)isCurrentReleaseWasUpdated:(NSString *)lastDownloadedReleaseHashes {
+  if (lastDownloadedReleaseHashes == nil) {
+    return NO;
+  }
+  NSString *currentInstalledReleaseHash = MSPackageHash();
```
jaelim-ms: We need to check `MSPackageHash` in terms of performance. You removed all currentReleaseHash from methods' parameter that seems like you call this method whenever you need it.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: You are right, fixed. Thank you.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: One user can be added to different distribution groups and the user may download and update an release from another group. So after update and restart, we need to use another distribution group ID and update it in storage.

I added some code comments so I hope it should be clearer.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: This value is saved after a new release is downloaded and removed after an app update and restart. After the restart, we check if group ID of downloaded release is the same as current one and update it if not.

I added some code comments so I hope it should be clearer.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -256,6 +261,14 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
                                      currentInstalledReleaseHash:(NSString *)currentInstalledReleaseHash
                                              distributionGroupId:(NSString *)distributionGroupId;
 
+/**
+ * After an app is updated and restarted, check if an updated release has different group ID and update current
+ * group ID if needed.
+ * Group ID may change if one user is added to different distribution groups and a new release
+ * was updated from another group.
+*/
+- (void)changeDistributionGroupIdAfterAppUpdateIfNeeded:(NSString *)currentInstalledReleaseHash;
```
jaelim-ms: Missing `@param`

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -225,32 +230,46 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
 - (BOOL)handleUpdate:(MSReleaseDetails *)details;
 
 /**
- * Store details about downloaded release to report it after installation.
+ * Save details about a downloaded release.
+ * After an app is updated and restarted, this info will be used to report a download and to update the group ID
+ * (if it was changed).
```
jaelim-ms: Missing `@param`.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -230,16 +235,16 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
...
  *
  * @param details Release details.
  */
 - (void)storeDownloadedReleaseDetails:(nullable MSReleaseDetails *)details;
 
 /**
  * Remove details about downloaded release after it was installed.
- *
- * @param currentInstalledReleaseHash The release hash of the current version.
```
jaelim-ms: Missing `@param`

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -807,6 +809,31 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
+    return;
+  }
+
+  // Skip if the group ID of an updated release is the same as the current one.
+  NSString *currentDistributionGroupId = [MS_USER_DEFAULTS objectForKey:kMSDistributionGroupIdKey];
+  if ((currentDistributionGroupId == nil) || ([lastDownloadedDistributionGroupId isEqualToString:currentDistributionGroupId] == NO)) {
+
+    // Set group ID from downloaded release details if an updated release was downloaded from another distribution group.
+    MSLogDebug([MSDistribute logTag], @"Current group ID doesn't match the group ID of downloaded release, updating current group id: %@", lastDownloadedDistributionGroupId);
+    [MS_USER_DEFAULTS setObject:lastDownloadedDistributionGroupId forKey:kMSDistributionGroupIdKey];
```
jaelim-ms: What I mentioned is that it seems `currentDistributionGroupId` is the new value and `lastDownloadedDistributionGroupId` is the latest value stored in `NSUserDefaults`.
The value doesn't match which means it needs to be updated but you are storing the old value, not the new one.
Please correct me if I'm wrong.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -807,6 +809,31 @@ - (NSMutableDictionary *)getReportingParametersForUpdatedRelease:(NSString *)upd
...
+  }
+
+  // Remove saved downloaded group ID.
+  [MS_USER_DEFAULTS removeObjectForKey:kMSDownloadedDistributionGroupIdKey];
```
jaelim-ms: Sounds like `kMSDownloadedDistributionGroupIdKey` will be used in a same session. Why don't you pass this value as a parameter?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: `currentDistributionGroupId` is used to store the group ID of currently running release (before the update). `lastDownloadedDistributionGroupId` to store the ID of a release that was downloaded before app update and restart. So the workflow is:
- In in-ap update dialog a user press "Download". SDK saves details about the downloaded release (`lastDownloadedDistributionGroupId`).
- After an app restart, SDK checks if currently running release was actually updated from downloaded IPA.
- If an updated release was downloaded from another distribution group then we need to update the stored ID (move the user to another group).

I updated the variables names so I hope it's more clear now. Thank you.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
roland-chernov-akvelon: It will be used after restart of an app. If a downloaded release was actually installed then SDK will use group ID of downloaded release to update stored group ID.

SDK checks if the hash of currently running app matches the hash of the release that was recently downloaded. This logic is executed only if they match. If hashes match then it means that downloaded release was installed and we need to check the group ID. If the user downloaded a release from another group ID then we need to update stored group ID.

Please let me know if you have any questions.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
roland-chernov-akvelon: Fixed. Thank you.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
roland-chernov-akvelon: Sorry, I'm not sure what you mean. Here is the only `details` parameter and it's documented. Maybe I didn't get something. Maybe I don't understand something.

jaelim-ms: Oops.. I didn't see that. It's weird, GitHub didn't show me that in my previous screen. 😄

PR #941, opened 2018-02-26
Bump minor version to 1.5.0

meiravf:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # App Center SDK for iOS and macOS Change Log
 
+## Version 1.5.0
```
jaelim-ms: Handled exception is not part of this release?

meiravf: No but the changes we made in distribution applies to analytics as well, right?

jaelim-ms: The reason why I asked is that if there are more changes than Distribute in this release, we should make sure all changes are listed in the CHANGELOG.
That's Distribute feature so you don't need to have Analytics section here.

PR #949, opened 2018-03-22
Feature/finalize naming for transmission

TroubleMakerBen:
tenant -> transmissionTarget
tenantId -> transmissionTargetToken


AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -101,17 +101,17 @@ - (NSString *)serializeLogWithPrettyPrinting:(BOOL)prettyPrint {
   return jsonString;
 }
 
-#pragma mark - Tenant
+#pragma mark - Transmission Target logic
 
-- (NSSet *)getTenants {
+- (NSSet *)getTransmissionTargets {
```
clpolet: I think we get the `getTransmissionTargetTokens` here not the target objects, we should add tokens to the name?

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -101,17 +101,17 @@ - (NSString *)serializeLogWithPrettyPrinting:(BOOL)prettyPrint {
...
   @synchronized(self) {
-    return self.tenants;
+    return self.transmissionTargetTokens;
   }
 }
 
-- (void)addTenant:(NSString *)tenant {
+- (void)addTransmissionTargetFor:(NSString *)token {
```
clpolet: Now that I think of it, it should probably be `addTransmissionTargetToken`.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
clpolet: I wonder if this one is already specified or not.

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsPrivate.h
```diff
@@ -1,6 +1,6 @@
 #import "MSAnalytics.h"
 #import "MSAnalyticsDelegate.h"
-#import "MSAnalyticsTenant.h"
+#import "MSTransmissionTarget.h"
```
clpolet: Import should be in the end.

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsPrivate.h
```diff
@@ -19,23 +19,23 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
-@property(nonatomic) NSMutableDictionary *tenants;
+@property(nonatomic) NSMutableDictionary *transmissionTargets;
 
 /**
- * Default tenant.
+ * Default transmission target.
  */
-@property(nonatomic) MSAnalyticsTenant *defaultTenant;
+@property(nonatomic) MSTransmissionTarget *defaultTransmissionTarget;
```
clpolet: I think it should stay an analytics target, probably a discussion we need with other folks.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -22,7 +22,7 @@ @implementation MSAbstractLog
 - (instancetype)init {
   self = [super init];
   if (self) {
-    self.tenants = [NSMutableSet new];
+    self.transmissionTargetTokens = [NSSet new];
```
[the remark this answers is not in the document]
TroubleMakerBen: The reason is that because we are not using `getTransmissionTargetTokens` but `transmissionTargetTokens`, the compiler now complains about the type mismatch.
I was thinking that we don't want the property mutable so that's why.
Does that make sense?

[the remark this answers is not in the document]
clpolet: I see it's now a classic getter on the property, it's fine.

PR #950, opened 2018-03-29
Fix/parsing of target token

TroubleMakerBen:


TroubleMakerBen:
@clpolet let me know what you think about lazy init for targetTokens. It's not necessary to fix the bug.

AppCenter/AppCenter/Internals/Util/MSUtility+StringFormatting.m
```diff
@@ -67,4 +75,17 @@ + (NSString *)transmissionTargetTokenFrom:(NSString *)string {
   }
 }
 
++ (BOOL)isValidAppSecret:(NSString *)appSecretString {
+  if(appSecretString == nil) {
```
clpolet: You can probably merge the two tests, it pretty common to have a nil or empty test.

PR #952, opened 2018-04-04
Make MSLogger public again

MatkovIvan:


AppCenter/AppCenter/MSLogger.h
```diff
@@ -1,7 +1,6 @@
 #import <Foundation/Foundation.h>
 
 #import "MSConstants.h"
-#import "MSLogger.h"
```
MatkovIvan: OMG it was recursive. How did it work?

clpolet: Interesting...

AppCenter/AppCenter/MSLogger.h
```diff
@@ -1,7 +1,6 @@
 #import <Foundation/Foundation.h>
```
clpolet: You can probably move the declaration of `msDefaultLogHandler` to the internal header.

clpolet: This header should also be part of the umbrella header.

PR #955, opened 2018-04-05
 Increased event and handled error max property count and item length

guperrot:


TroubleMakerBen:
I fixed the tests & cleaned up the validation logic for properties. It's now shared among modules.

PR #957, opened 2018-04-06
Don't start services which requires app secret when no app secret

clpolet:
Services requiring an app secret won't start if none is provided on start.


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -288,19 +288,30 @@ - (BOOL)startService:(Class)clazz andSendLog:(BOOL)sendLog {
       // Service already works, we shouldn't send log with this service name
       return NO;
     }
+    if (service.isAppSecretRequired && ![self.appSecret length]) {
```
TroubleMakerBen: can we be sure that `self.appSecret` is not null?

PR #971, opened 2018-05-08
Add OneCollector listener on AppCenter start

bmourat:


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -423,6 +424,7 @@ - (void)initializeChannelGroup {
     // If there is no app secret, create a channel group without sender or storage.
     self.channelGroup = [[MSChannelGroupDefault alloc] initWithSender:nil storage:nil];
   }
+  [self.channelGroup addDelegate: [MSOneCollectorChannelDelegate alloc]];
```
clpolet: The OneCollector delegate was not initialized. You can replace alloc by new.

clpolet: You will need to retain the delegate in a private property because the delegates are stored in a weak NSHashTable.

PR #972, opened 2018-05-08
Fix project for XCode 9.3

guperrot:
NULL is not allowed on xcode 9.3, need to replace with nullptr


jwallra:
Please remember to put a description or link to the issue

PR #985, opened 2018-05-16
Move AppCenter Analytics validation to shouldFilter callback

MatkovIvan:


AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -187,6 +187,25 @@ + (BOOL)isAutoPageTrackingEnabled {
 
 #pragma mark - Private methods
 
+- (BOOL)validateLog:(MSLogWithNameAndProperties *)log {
```
clpolet: I would say we can get rid of this method and do everything in the shouldFilter.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
```diff
@@ -4,7 +4,6 @@
 static NSString *const kMSTypeEvent = @"event";
 
 static NSString *const kMSId = @"id";
-static NSString *const kMSName = @"name";
```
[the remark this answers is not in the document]
clpolet: We'll update the schema later.

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalytics+Validation.m
```diff
@@ -0,0 +1,58 @@
...
+#import "MSPageLog.h"
+
+// Events values limitations
+static const int minEventNameLength = 1;
+static const int maxEventNameLength = 256;
+
+@implementation MSAnalytics (Validation)
+
+- (BOOL)shouldFilterLog:(id<MSLog>)log {
+  NSObject *logObject = (NSObject *)log;
```
clpolet: I tried to put a break point here but it never got hit. Is the delegate registered?

MatkovIvan: Yeah, this is linker issue in the test app. I've added a workaround such as we use in other categories.

PR #1004, opened 2018-06-12
Fix/transmission bugs

TroubleMakerBen:
- makes sure `channel:didCompleteEnqueueingLog:withInternalId` gets called even if a sender is nil or a log gets filtered. This makes sure events are removed from the logbuffer correctly
- Don't re-enqueue the CommonSchemaLog from the main thread but from a background thread to avoid blocking the UI thread*

* the side effect of this is that the logbuffer is less effective for common schema logs

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -11,6 +11,7 @@
 @implementation MSChannelUnitDefault
 
 @synthesize configuration = _configuration;
+@synthesize logsDispatchQueue = _logsDispatchQueue;
 
 #pragma mark - Initialization
 
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -133,12 +133,20 @@
     // If sender is nil, there is nothing to do at this point.
     if (shouldFilter) {
       MSLogDebug([MSAppCenter logTag], @"Log of type '%@' was filtered out by delegate(s)", item.type);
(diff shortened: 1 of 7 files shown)


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -133,12 +134,20 @@ - (void)enqueueItem:(id<MSLog>)item {
     // If sender is nil, there is nothing to do at this point.
     if (shouldFilter) {
       MSLogDebug([MSAppCenter logTag], @"Log of type '%@' was filtered out by delegate(s)", item.type);
+      [self enumerateDelegatesForSelector:@selector(channel:didCompleteEnqueueingLog:withInternalId:)
```
jaelim-ms: Not a merge blocker but we need to think about not repeating this in many places in the method.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -420,6 +420,16 @@ - (void)channel:(id<MSChannelProtocol>)__unused channel
     return;
   }
 
+  // FIXME: Rejecting MSCommonSchemaLog in the buffer for now.
+  /*
+   * Don't buffer logs that are MSCommonSchemaLogs or contain one or more transmissiontargetTokens for now. The reason
+   * is that the crash buffer currently cannot handle MSCommonSchemaLogs. It needs significant work, e.g. for
+   * encrypting/decrypting the target tokens. Putting this in a separate paragraph for visibility.
+   */
+  if ([logObject isKindOfClass:[MSCommonSchemaLog class]] || log.transmissionTargetTokens) {
```
clpolet: You can also remove the `isKindOfClass:[MSCommonSchemaLog class]` test, testing existence of transmission token should be enough, AC only logs won't set it.

PR #1012, opened 2018-06-15
Fix .a files having MSCSModelConstants.o twice with a suffix

guperrot:
screen shot 2018-06-14 at 5 20 44 pm

Because the .m file was twice in the project, it generated 2 o files with a suffix causing xamarin to failed linking with duplicate objects.

With this fix, the numeric suffix is gone and there is only 1 .o file now.


TroubleMakerBen:
For context:
MSCSModelConstants.m was added to the Compile Source Build Phase twice.

PR #1015, opened 2018-06-15
Don’t use checkResourceIsReachableAndReturnError in unnecessary places

TroubleMakerBen:
On testing, I found that MSUtility+File.m was polluting the console with log statements about failing to delete files.
My first approach was to just change the log level - most cases where deletion fails is about non-existing files in the first place.
So first, I revisited the logging statements in the class and changed a few from the error to the debug level. So far, so good.
Then I noticed that we're checking for file existence quite a lot. This is expensive (as it's synchronous) and not necessary, so I've changed that.

### DIFF
--- a/AppCenter/AppCenter/Internals/Util/MSUtility+File.m
+++ b/AppCenter/AppCenter/Internals/Util/MSUtility+File.m
@@ -31,7 +31,10 @@ static NSString *const kMSAppCenterBundleIdentifier = @"com.microsoft.appcenter"
 
       // Create parent directories as needed.
       NSURL *directoryURL = [fileURL URLByDeletingLastPathComponent];
-      [self createDirectoryAtURL:directoryURL];
+      BOOL success = [self createDirectoryAtURL:directoryURL];
+      if(!success) {
+        return nil;
+      }
 
       // Create the file.
       NSData *theData = (data != nil) ? data : [NSData data];
@@ -171,6 +174,10 @@ static NSString *const kMSAppCenterBundleIdentifier = @"com.microsoft.appcenter"
     dirURL = [baseDirUrl URLByAppendingPathComponent:kMSAppCenterBundleIdentifier];
 #endif
     [self createDirectoryAtURL:dirURL];
+    BOOL success = [self createDirectoryAtURL:dirURL];
+    if(!success) {
+      MSLogError([MSAppCenter logTag], @"Unable to create directory for AppCenter SDK.");
+    }
   });
 
   return dirURL;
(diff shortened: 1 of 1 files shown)


AppCenter/AppCenter/Internals/Util/MSUtility+File.m
```diff
@@ -166,21 +170,24 @@ + (NSURL *)appCenterDirectoryURL {
 #else
     dirURL = [baseDirUrl URLByAppendingPathComponent:kMSAppCenterBundleIdentifier];
 #endif
-    if (![dirURL checkResourceIsReachableAndReturnError:nil]) {
-      [self createDirectoryAtURL:dirURL];
-    }
+    [self createDirectoryAtURL:dirURL];
```
jaelim-ms: Should we check the return value and log any errors here?

AppCenter/AppCenter/Internals/Util/MSUtility+File.m
```diff
@@ -24,16 +24,14 @@ + (NSURL *)createFileAtPathComponent:(NSString *)filePathComponent
...
       if (!forceOverwrite && [fileURL checkResourceIsReachableAndReturnError:nil]) {
         return fileURL;
       }
 
       // Create parent directories as needed.
       NSURL *directoryURL = [fileURL URLByDeletingLastPathComponent];
-      if (![directoryURL checkResourceIsReachableAndReturnError:nil]) {
-        [self createDirectoryAtURL:directoryURL];
-      }
+      [self createDirectoryAtURL:directoryURL];
```
jaelim-ms: Should we check the return value and log any errors here? Your comment mentions that **it is more efficient to attempt the operation and handle any failure that may occur.** which means we should handle errors in each operation.

AppCenter/AppCenter/Internals/Util/MSUtility+File.m
TroubleMakerBen: About logging errors: We are logging the error inside `createDirectoryAtURL:`, do you think we need more logging there?

AppCenter/AppCenter/Internals/Util/MSUtility+File.m
TroubleMakerBen: That's true, the issue here is that we can't really do anything except log an error message.
I've committed a change that handles the return value of createDirectoryAtURL and that tries to be more verbose in code as to what's going on.
Now about the actual error message of `createDiretoryAtURL:withIntermediateDirectories:attributes:error:`: we log the error message.
I can't come up with a more useful logic for errors. I was considering shutting down the SDK as we are useless in case the SDK cannot create any directories but shutting down the SDK seemed too radical.
Opinions?

TroubleMakerBen: @jaelim-ms no idea why github is hiding my comment.

clpolet: Probably don't need to shutdown the SDK, it will add complexity for not that much since the SDK won't work anyway. It should just not crash if such case happens.

AppCenter/AppCenter/Internals/Util/MSUtility+File.m
jaelim-ms: Oh.. I see. Then probably we don't need it.

TroubleMakerBen:
Reverted the last commit.

PR #1022, opened 2018-06-20
Fix/codecov fix

TroubleMakerBen:


TroubleMakerBen:
This was for testing purposes only.

PR #1027, opened 2018-06-26
Adds nested transmissionTargets feature

clpolet:
Adds nested transmissionTargets feature. One can call the `transmissionTargetForToken:` method on a  transmission target to create a nested transmission target.


AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -1,12 +1,14 @@
 #import "MSAnalyticsTransmissionTargetInternal.h"
 #import "MSAnalyticsInternal.h"
```
dhei: nit - please sort the import statements.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -2,9 +2,11 @@
```
jaelim-ms: Can you remove XCTest import even though it isn't your change? XCTest is imported in MSTestFrameworks.h.

PR #1029, opened 2018-06-28
Feature/dont require appsecret in sender

TroubleMakerBen:
Some context on this:
- the way `setAppSecret:` is implemented is a little unituitive. It should be a property on `MSChannelProtocol` with `@synthesize` statements in `MSDefaultChannelGroup` and `MSDefaultChannelUnit` with `MSDefaultChannelGroup` having the custom setter (`setAppSecret:`). The problem is that the compiler somehow doesn't "understand" that `setAppSecret:` is a setter, and it throws a warning because we're accessing the `ivar` directly. This doesn't make any sense but we spent about 2hrs yesterday trying different solutions and this PR is what we came up with.
- we experimented with passing around the created `NSURLRequest`. Unfortunately, this will make tracking `batchId` really hard as a `NSURLRequest` contains the `MSLogContainer` as it's body but serialized to an `NSString`, which is then converted and compressed to NSData. So accessing the actual content of an the request object becomes very expensive, so we decided to not pass around `NSURLRequest` objects.
- we also looked at removing the `appSecret` property from the `ChannelUnitDefault` and only have it in `ChannelGroupDefault` - this would be more similar to Android. The problem is that Android has most of the batching, etc. logic in `ChannelGroup`/`Channel`, while iOS/Apple has lots of it one level down in `ChannnelUnitDefault`. We think that it's best to keep the `appSecret` in both `...Group` and `...Unit` for now.

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupProtocol.h
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupProtocol.h
@@ -39,11 +39,10 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Attach an App Center sender to a channel group if a channel group started without an app secret.
  *
- * @param appSecret A unique and secret key used to identify the application.
  * @param installId A unique installation identifier.
  * @param logUrl A base URL to use for backend communication.
  */
-- (void)attachSenderWithAppSecret:(NSString *)appSecret installId:(NSUUID *)installId logUrl:(NSString *)logUrl;
+- (void)attachSenderWithInstallId:(NSUUID *)installId logUrl:(NSString *)logUrl;
 
 /**
  * Change the base URL (schema + authority + port only) used to communicate with the backend.
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -8,7 +8,7 @@
 #import "MSSender.h"
 #import "MSStorage.h"
 
-@interface MSChannelUnitDefault()
+@interface MSChannelUnitDefault ()
 
 @property(nonatomic, copy) NSString *appSecret;
 
@@ -259,6 +259,7 @@
 
                // Forward logs to the sender.
                [self.sender sendAsync:container
+                            appSecret:self.appSecret
                     completionHandler:^(NSString *senderBatchId, NSUInteger statusCode,
                                         __attribute__((unused)) NSData *data, NSError *error) {
                       dispatch_async(self.logsDispatchQueue, ^{
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -8,12 +8,6 @@
 #import "MSSender.h"
 #import "MSStorage.h"
 
-@interface MSChannelUnitDefault ()
-
-@property(nonatomic, copy) NSString *appSecret;
-
-@end
-
 @implementation MSChannelUnitDefault
 
 @synthesize configuration = _configuration;
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -8,6 +8,12 @@
 #import "MSSender.h"
 #import "MSStorage.h"
 
+@interface MSChannelUnitDefault()
+
+@property(nonatomic, copy) NSString *appSecret;
+
+@end
+
 @implementation MSChannelUnitDefault
 
 @synthesize configuration = _configuration;
@@ -48,6 +54,10 @@
     }
   }
   return self;
+}
+
+- (void)setAppSecret:(NSString *)appSecret {
+  _appSecret = appSecret;
 }
 
 #pragma mark - MSChannelDelegate
--- a/AppCenter/AppCenter/Internals/Sender/MSAppCenterIngestion.m
+++ b/AppCenter/AppCenter/Internals/Sender/MSAppCenterIngestion.m
@@ -13,12 +13,11 @@ static NSString *const kMSAPIVersion = @"1.0.0";
 static NSString *const kMSAPIVersionKey = @"api-version";
 static NSString *const kMSApiPath = @"/logs";
 
-- (id)initWithBaseUrl:(NSString *)baseUrl appSecret:(NSString *)appSecret installId:(NSString *)installId {
+- (id)initWithBaseUrl:(NSString *)baseUrl installId:(NSString *)installId {
   self = [super initWithBaseUrl:baseUrl
       apiPath:kMSApiPath
       headers:@{
         kMSHeaderContentTypeKey : kMSAppCenterContentType,
-        kMSHeaderAppSecretKey : appSecret,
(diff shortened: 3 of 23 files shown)


AppCenter/AppCenter/Internals/Sender/MSAppCenterIngestion.m
```diff
@@ -29,7 +28,9 @@ - (id)initWithBaseUrl:(NSString *)baseUrl appSecret:(NSString *)appSecret instal
   return self;
 }
 
-- (void)sendAsync:(NSObject *)data completionHandler:(MSSendAsyncCompletionHandler)handler {
+- (void)sendAsync:(NSObject *)data
```
dhei: nit - `:` not aligned correctly

AppCenter/AppCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -34,10 +34,14 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Send data to backend
  * @param data A data instance that will be transformed request body.
+ * @param appSecret The appSecret.
  * @param callId A unique ID that identify a request.
  * @param handler Completion handler
  */
-- (void)sendAsync:(NSObject *)data callId:(NSString *)callId completionHandler:(MSSendAsyncCompletionHandler)handler;
+- (void)sendAsync:(NSObject *)data
+            appSecret:(NSString *)appSecret
```
dhei: nit - `:` not aligned

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -407,7 +415,10 @@ - (NSString *)prettyPrintHeaders:(NSDictionary<NSString *, NSString *> *)headers
   return [flattenedHeaders componentsJoinedByString:@", "];
 }
 
-- (void)sendAsync:(NSObject *)data callId:(NSString *)callId completionHandler:(MSSendAsyncCompletionHandler)handler {
+- (void)sendAsync:(NSObject *)data
+            appSecret:(NSString *)appSecret
```
dhei: nit - same formatting issue

AppCenter/AppCenter/Internals/Channel/MSChannelGroupProtocol.h
```diff
@@ -39,11 +39,10 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Attach an App Center sender to a channel group if a channel group started without an app secret.
  *
- * @param appSecret A unique and secret key used to identify the application.
  * @param installId A unique installation identifier.
  * @param logUrl A base URL to use for backend communication.
  */
-- (void)attachSenderWithAppSecret:(NSString *)appSecret installId:(NSUUID *)installId logUrl:(NSString *)logUrl;
+- (void)attachSenderWithInstallId:(NSUUID *)installId logUrl:(NSString *)logUrl;
```
jaelim-ms: Are you going to remove installId later?

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -249,6 +253,7 @@ - (void)flushQueue {
 
                // Forward logs to the sender.
                [self.sender sendAsync:container
+                            appSecret:self.appSecret
```
jaelim-ms: Not a change request but we need to consider moving this method and others to channel group some time later.

AppCenter/AppCenter/Internals/Sender/MSAppCenterIngestion.m
jaelim-ms: I think it is because of `completionHandler` is too long.

AppCenter/AppCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -17,12 +17,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, copy) NSString *apiPath;
 
 /**
- *	Send Url.
+ *  Send Url.
```
jaelim-ms: One space between  `*` and `S`

AppCenter/AppCenter/Internals/Sender/MSHttpSender.h
```diff
@@ -17,12 +17,12 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 @property(nonatomic) NSURL *sendURL;
 
 /**
- *	Request header parameters.
+ *  Request header parameters.
```
jaelim-ms: Here too.

AppCenter/AppCenter/Internals/Sender/MSAppCenterIngestion.m
TroubleMakerBen: Jae is right, nothing I can do except rename the method which doesn't seem right to me

AppCenter/AppCenter/Internals/Sender/MSHttpSender.h
TroubleMakerBen: same as above

AppCenter/AppCenter/Internals/Sender/MSHttpSender.m
```diff
@@ -362,8 +369,9 @@ - (void)networkStateChanged {
 /**
  * This is an empty method and expect to be overridden in sub classes.
  */
-- (NSURLRequest *)createRequest:(NSObject *)data {
+- (NSURLRequest *)createRequest:(NSObject *)data appSecret:(NSString *)appSecret {
   (void)data;
+  (void)appSecret;
```
jaelim-ms: Can we do `_unused` in the signature instead?

AppCenter/AppCenter/Internals/Sender/MSHttpSenderPrivate.h
```diff
@@ -70,9 +70,10 @@
 /**
  * Create a request based on data. Must override this method in sub classes.
  * @param data A data instance that will be transformed to request body.
+ * @param appSecret The app secret.
```
jaelim-ms: Add an empty line before and after `@param`

AppCenter/AppCenter/Internals/Sender/MSSender.h
```diff
@@ -25,9 +25,10 @@ NS_ASSUME_NONNULL_BEGIN
  * Send data.
  *
  * @param data Instance that will be transformed to request body.
+ * @param appSecret The app secret.
  * @param handler Completion handler.
  */
-- (void)sendAsync:(nullable NSObject *)data completionHandler:(MSSendAsyncCompletionHandler)handler;
+- (void)sendAsync:(nullable NSObject *)data appSecret:(nullable NSString *)appSecret completionHandler:(MSSendAsyncCompletionHandler)handler;
```
jaelim-ms: `appSecret` is always nullable. Why is it only specified in Sender?

AppCenter/AppCenter/Internals/Sender/MSHttpSender.h
clpolet: Our code formatter does that :slightly_smiling_face:

AppCenter/AppCenter/Internals/Sender/MSSender.h
clpolet: It should just not be nullable on the setAppSecret maybe.

PR #1030, opened 2018-06-29
Feature/transmission target enabled

clpolet:
This PR adds the enabled state logic to transmission targets.

Calling `setEnabled:` on a nested transmission target will propagate the enabled state to all its child targets.

The enabled state of Analytics is also propagated to registered transmission targets.


AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -37,10 +56,49 @@ - (MSAnalyticsTransmissionTarget *)transmissionTargetForToken:(NSString *)token
...
+      MSLogWarning([MSAnalytics logTag],
+                   @"Can't enable; parent transmission target and/or Analytics service is disabled.");
+      return;
+    }
+
+    // Persist the enabled status.
+    [self.storage setObject:@(isEnabled) forKey:self.isEnabledKey];
+  }
+
+  // Propagate to nested transmission targets. TODO Find a more effective approach.
```
clpolet: Can probably found a better algorithm to travers the child tree but it will do the trick for now.

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -724,9 +724,10 @@ - (void)testContentsOfDirectory {
   BOOL forceOverwrite = NO;
   BOOL atomical = YES;
   NSUInteger fileCount;
-  for (fileCount = 0; fileCount<3 ; fileCount++){
+  for (fileCount = 0; fileCount < 3; fileCount++) {
```
jwallra: I know this is just formatting, but why 3? This should be in a constant

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
jwallra: You have do do a single operation for each child so it will still take O(N = members of child tree), I think.  We might be able to reduce recursion but this is probably fine.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -105,4 +122,140 @@ - (void)testTransmissionTargetForToken {
       [MSAnalytics trackEvent:event3 withProperties:properties forTransmissionTarget:childTransmissionTarget3]));
 }
 
+- (void)testTransmissionTargetEnabledState {
```
jwallra: It is probably worth doing a test with a long chain of children and a test with a large number of first-level children.  And we should include a test for the case @guperrot found where parts of a tree haven't been stored locally yet.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -105,4 +122,140 @@ - (void)testTransmissionTargetForToken {
...
+
+  // Then
+  XCTAssertFalse([target isEnabled]);
+  XCTAssertFalse([childTarget isEnabled]);
+  XCTAssertFalse([subChildTarget isEnabled]);
+  XCTAssertFalse([childTarget2 isEnabled]);
+
+  // When
+
+  // Enabling a parent enables its childs.
```
jwallra: children

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
clpolet: OK I'll do the test with the long chain of children and a large number of first level children. The other test is already covered by the later addition of "childTarget2-guid" into the tree.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
clpolet: Ow good catch.

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsTransmissionTargetPrivate.h
```diff
@@ -1,16 +1,45 @@
...
+
 /**
  * Child transmission targets nested to this transmission target.
  */
 @property(nonatomic) NSMutableDictionary<NSString *, MSAnalyticsTransmissionTarget *> *childTransmissionTargets;
 
+/**
+ * Storage used for persistence.
+ */
+@property(nonatomic) MSUserDefaults *storage;
```
jaelim-ms: Any reason of having this as a property of the class?

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -1,16 +1,17 @@
+#import "MSAnalytics+Validation.h"
```
jaelim-ms: I think the correct order is below ` #import "MSAnalytics.h"`. The rule of sorting import isn't ascii order if I remember correctly.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -208,18 +218,22 @@ - (void)trackEvent:(NSString *)eventName
   // Create an event log.
   MSEventLog *log = [MSEventLog new];
 
+  // Add transmission target token.
+  if (transmissionTarget) {
+    if (transmissionTarget.isEnabled) {
+      [log addTransmissionTargetToken:[transmissionTarget transmissionTargetToken]];
+    } else {
+      MSLogError([MSAnalytics logTag], @"This transmission target is disabled.");
```
jaelim-ms: This isn't an error log. Should be information.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -1,18 +1,37 @@
...
 
 @implementation MSAnalyticsTransmissionTarget
 
-- (instancetype)initWithTransmissionTargetToken:(NSString *)transmissionTargetToken {
-  self = [super init];
-  if (self) {
-    _transmissionTargetToken = transmissionTargetToken;
+- (instancetype)initWithTransmissionTargetToken:(NSString *)token
+                                   parentTarget:(nullable MSAnalyticsTransmissionTarget *)parentTarget
+                                        storage:(MSUserDefaults *)storage {
```
jaelim-ms: Can you explain why you need this as a parameter?

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -37,10 +56,49 @@ - (MSAnalyticsTransmissionTarget *)transmissionTargetForToken:(NSString *)token
...
+    [self.childTransmissionTargets[token] setEnabled:isEnabled];
+  }
+}
+
+/**
+ * Check ancestor enabled state, the ancestor is either the immediate target parent if there is one or Analytics.
+ *
+ * @return YES if the immediate ancestor is enabled.
+ */
+- (BOOL)isAncestorEnabled {
```
jaelim-ms: Is the method somewhat we've agreed with other platforms? It only checks it's immediate ancestor, it would be better to say parent I guess.
The explanation for the method isn't quite understandable. Can you revise this one?

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -1,24 +1,26 @@
+#import "MSAnalytics+Validation.h"
```
jaelim-ms: Same here, order.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -1,23 +1,37 @@
...
 #pragma mark - Tests
 
 - (void)testInitialization {
 
   // When
   MSAnalyticsTransmissionTarget *transmissionTarget =
-      [[MSAnalyticsTransmissionTarget alloc] initWithTransmissionTargetToken:kMSTestTransmissionToken];
+      [[MSAnalyticsTransmissionTarget alloc] initWithTransmissionTargetToken:kMSTestTransmissionToken
+                                                                parentTarget:nil
+                                                                     storage:self.storageMock];
```
jaelim-ms: Probably this is why you added an init with storage parameter in private header. We have settings mock which also mocks `NSUserDefaults` and you can use this one instead.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
jwallra: Yeah, we should change the name of this method.  I thought we needed to mirror the behavior in android but it's not necessary since we copy the state of the parent when creating the transmission target.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
jaelim-ms: We can use `MSMockUserDefault` in this case as discussed offline. Let's add a TODO for this one if you don't have time to address.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -37,10 +56,53 @@ - (MSAnalyticsTransmissionTarget *)transmissionTargetForToken:(NSString *)token
...
+        MSLogWarning([MSAnalytics logTag],
+                     @"Can't enable; parent transmission target and/or Analytics service is disabled.");
+        return;
+      }
+
+      // Persist the enabled status.
+      [self.storage setObject:@(isEnabled) forKey:self.isEnabledKey];
+    }
+
+    // Propagate to nested transmission targets. TODO Find a more effective approach.
```
jaelim-ms: You can remove TODO here. As @jwallra mentioned, it would be O(n) for the best case.

clpolet: Done, it was more on the memory usage then the complexity but it should be fine the way it.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
clpolet: Yes `MSMockUserDefault` will replace `NSUserDefaults` instance at run time, so no need to pass it around. will need to clean the AbstractService class/tests too. TODO added.

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -1635,7 +1629,6 @@
 				3568A7E6201A8A6300F3B860 /* MSChannelProtocol.h in Headers */,
 				04754F761EA98182002CBA46 /* MSAppCenterPrivate.h in Headers */,
 				04754F771EA98182002CBA46 /* MSServiceAbstractInternal.h in Headers */,
-				04754F791EA98182002CBA46 /* MSServiceAbstractPrivate.h in Headers */,
```
TroubleMakerBen: Less headers! I like it!

AppCenter/AppCenter/MSServiceAbstract.m
```diff
@@ -29,7 +23,7 @@ - (BOOL)isEnabled {
 
   // Get isEnabled value from persistence.
   // No need to cache the value in a property, user settings already have their cache mechanism.
-  NSNumber *isEnabledNumber = [self.storage objectForKey:self.isEnabledKey];
+  NSNumber *isEnabledNumber = [MS_USER_DEFAULTS objectForKey:self.isEnabledKey];
```
TroubleMakerBen: `self.storage` here was so bad! Thx for cleaning this up!

PR #1032, opened 2018-06-29
Implement start from library

jaelim-ms:

PR #1033, opened 2018-07-02
Add APIs to start/configure app center with no app secret

achocron:


jaelim-ms:
I mark this PR as don't merge because of #1032. I want to minimize a big PR being struggled with merge issues that could be merged silently.

AppCenter/AppCenter/MSAppCenter.h
```diff
@@ -31,6 +38,14 @@
  */
 + (void)start:(NSString *)appSecret withServices:(NSArray<Class> *)services;
 
+/**
+ * Configure the SDK with an array of services to start.
+ *
+ * @discussion This may be called only once per application process lifetime.
```
jaelim-ms: We prevent from starting AC from application multiple times already.

AppCenter/AppCenter/MSAppCenter.h
```diff
@@ -31,6 +38,14 @@
  */
 + (void)start:(NSString *)appSecret withServices:(NSArray<Class> *)services;
 
+/**
+ * Configure the SDK with an array of services to start.
```
jaelim-ms: Should we mention that it can start without app secret explicitly?

AppCenter/AppCenter/MSAppCenter.h
TroubleMakerBen: I'd still mention this in the header file.

AppCenter/AppCenter/MSAppCenter.h
TroubleMakerBen: I'd point out that a `configure`-call needs to be balanced with a `start` call.

AppCenter/AppCenter/MSAppCenter.h
clpolet: Yes, still good to have.

PR #1034, opened 2018-07-04
Add support for MSCommonSchemaLog in the log buffer

MatkovIvan:


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -519,9 +532,10 @@ - (void)channel:(id<MSChannelProtocol>)__unused channel
       NSString *bufferId = [NSString stringWithCString:it->internalId.c_str() encoding:NSUTF8StringEncoding];
       if (bufferId && bufferId.length > 0 && [bufferId isEqualToString:internalId]) {
         MSLogVerbose([MSCrashes logTag], @"Deleting a log from buffer with id %@", internalId);
-        it->buffer = [@"" cStringUsingEncoding:NSUTF8StringEncoding];
-        it->timestamp = [@"" cStringUsingEncoding:NSUTF8StringEncoding];
-        it->internalId = [@"" cStringUsingEncoding:NSUTF8StringEncoding];
+        it->buffer = "";
+        it->targetToken = "";
+        it->timestamp = "";
+        it->internalId = "";
```
TroubleMakerBen: Why did you replace the cString logic here? Using c strings here is the right thing to do here in my opinion.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -435,19 +435,14 @@ - (void)testSetupLogBufferWorks {
...
 - (void)testEmptyLogBufferFiles {
+
   // If
   NSString *testName = @"afilename";
   NSString *dataString = @"SomeBufferedData";
   NSData *someData = [dataString dataUsingEncoding:NSUTF8StringEncoding];
   NSString *filePath = [NSString stringWithFormat:@"%@/%@", self.sut.logBufferPathComponent,
                                                   [testName stringByAppendingString:@".mscrasheslogbuffer"]];
-
-#if TARGET_OS_OSX
```
TroubleMakerBen: Good catch! How the heck was this in there the first place?! ;)

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: Actually nothing changed since `=` here is overridden operator for `std::string`. It will copy the right part anyway, so I don't see any reason for additional conversion from obj-c string.

PR #1036, opened 2018-07-05
Merge changes to start with no app secret.

TroubleMakerBen:
Duplicate of Zander's PR, because of the transient error in CI that we can't repro locally at all.


TroubleMakerBen:
using #1033 instead =)

PR #1037, opened 2018-07-06
Add start from library features to sasquatch app

jwallra:
Also 2 bugfixes:

- Transmission target token not being set when start library was called first
- Using long secret string instead of app secret when updating config


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -214,6 +214,7 @@ - (BOOL)configureWithSecretString:(NSString *)secretString fromApplication:(BOOL
       }
       if (!self.defaultTransmissionTargetToken) {
         self.defaultTransmissionTargetToken = [MSUtility transmissionTargetTokenFrom:secretString];
+        NSLog(@"TTT: %@", self.defaultTransmissionTargetToken);
```
TroubleMakerBen: This shouldn't be `NSLog(..)` but I thinnk it's just ann oversight from testing?

jwallra: Yup, needs to be removed

Puppet/Puppet/AppDelegate.m
```diff
@@ -44,18 +50,28 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   [MSPush setDelegate:self];
   [MSAnalytics setDelegate:self];
   [MSAppCenter setLogLevel:MSLogLevelVerbose];
-
+
```
jwallra: Remove extra whitespace

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -251,7 +252,7 @@ - (void)start:(NSString *)appSecret withServices:(NSArray<Class> *)services from
                    (fromApplication ? @"an application" : @"a library"));
       NSMutableArray<NSString *> *servicesNames = [NSMutableArray arrayWithCapacity:sortedServices.count];
       for (Class service in sortedServices) {
-        if ([self startService:service withAppSecret:appSecret andSendLog:NO fromApplication:fromApplication]) {
+        if ([self startService:service withAppSecret:self.appSecret andSendLog:NO fromApplication:fromApplication]) {
```
TroubleMakerBen: great catch! I'm thinking maybe we should move `self.appSecret = appSecret` outside of `configureWithSecretString:fromApplication` or at least make it more obvious where we are assigning the value to the property 🤔

Puppet/Puppet/MSMainViewController.m
```diff
@@ -70,4 +72,30 @@ - (IBAction)enableOneCollectorSwitchUpdated:(UISwitch *)sender {
   [self presentViewController:alert animated:YES completion:nil];
 }
 
+- (IBAction)selectTarget:(UISegmentedControl *)sender {
```
jwallra: Question for reviewers: I copied this from another method but this formatting looks atrocious to me.  Is this how it's supposed to be in this case?  Line up all the `:`?

Puppet/Puppet/MSAnalyticsViewController.m
```diff
@@ -55,6 +55,8 @@ - (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)toInterfa
 }
 
 - (IBAction)trackEvent {
+  [MSAnalytics trackEvent:self.eventName.text withProperties:self.properties];
```
TroubleMakerBen: is this intentional?

jwallra:
Looks like we also need to fix some tests.

AppCenter/AppCenter/MSAppCenter.m
jwallra: Yeah I don't know if I would have found this bug without @clpolet's sharp eyes - I'll look at making it clearer.

jaelim-ms: This change was intentional. I needed to use the current appSecret to start service because of some reasons. I'll take a look at this code again.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -252,6 +254,7 @@ - (void)start:(NSString *)secretString withServices:(NSArray<Class> *)services f
     BOOL configured = [self configureWithAppSecret:appSecret
                            transmissionTargetToken:transmissionTargetToken
                                    fromApplication:fromApplication];
+
```
jaelim-ms: Remove empty line.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -315,6 +318,7 @@ - (BOOL)startService:(Class)clazz
       MSLogError([MSAppCenter logTag], @"App Center has not been configured so it couldn't start the service.");
       return NO;
     }
+
```
jaelim-ms: Remove empty line.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -145,6 +148,15 @@ - (BOOL)isAppSecretRequired {
   return NO;
 }
 
+- (void)updateConfigurationWithAppSecret:(NSString *)appSecret transmissionTargetToken:(NSString *)token {
+  [super updateConfigurationWithAppSecret:appSecret transmissionTargetToken:token];
+
+  // Create the default target if not already created in start.
+  if (token && !self.defaultTransmissionTarget) {
```
jaelim-ms: Why do you need this one? The callback will only be called when the SDK is started from library and then starting from application. `self.defaultTransmissionTarget` would always be `nil`.

Puppet/Puppet/Base.lproj/Main.storyboard
```diff
@@ -1343,11 +1365,12 @@
                         <outlet property="logUrl" destination="XnS-uI-vJv" id="2Vi-KP-bXN"/>
                         <outlet property="oneCollectorEnabled" destination="xOV-jw-JKQ" id="ETy-vH-bIw"/>
                         <outlet property="sdkVersion" destination="7FT-b5-EHq" id="f37-CJ-6xs"/>
+                        <outlet property="startTarget" destination="hRG-Uj-IS8" id="G6q-ou-YBn"/>
                     </connections>
                 </tableViewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="G4r-se-P85" userLabel="First Responder" sceneMemberID="firstResponder"/>
             </objects>
-            <point key="canvasLocation" x="745" y="850"/>
+            <point key="canvasLocation" x="744.79999999999995" y="849.62518740629696"/>
```
jaelim-ms: Can you manually revert this line?

Puppet/Puppet/MSAnalyticsViewController.m
```diff
@@ -55,6 +55,8 @@ - (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)toInterfa
...
+
```
jaelim-ms: No empty line.

Puppet/Puppet/MSAnalyticsViewController.m
jaelim-ms: Is this intentional?

TestApps/Sasquatch/SasquatchSwift/AppDelegate.swift
```diff
@@ -23,19 +27,34 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
 
     // Start App Center SDK.
     let useOneCollector = UserDefaults.standard.bool(forKey: "isOneCollectorEnabled");
-    if useOneCollector {
+    let startTarget = UserDefaults.standard.integer(forKey: "startTarget");
+    
+    let secretString = useOneCollector
+      ? "target=5a06bf4972a44a059d59c757e6d0b595-cb71af5d-2d79-4fb4-b969-01840f1543e9-6845;appsecret=0dbca56b-b9ae-4d53-856a-7c2856137d85"
```
jaelim-ms: Use `1dd3a9a64e144fcbbd4ce31c5def22e0-e57d4574-c5e7-4f89-a745-b2e850b54185-7090`

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
jaelim-ms: Oops. it is a target instance not target token. NM.

Puppet/Puppet/MSAnalyticsViewController.m
jwallra: Offline discussion - yes, it was intentional, but it prevents us from testing the scenario where both transmission targets are enabled but we want the runtime target to override the default target.  This was enabled in the previous version of the test app but is not in this one.  

We are going to create a card to improve the test app UI and functionality to 1) reenable this scenario and 2) be more easily understood.  Right now it's kind of tribal knowledge which buttons have to be pressed to get certain behavior, and we should try to make it more clear.

PR #1038, opened 2018-07-06
 Remove app secret from a method signature and add an error check

jaelim-ms:
For the comment in #1037 
https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1037#discussion_r200711883

### DIFF
--- a/AppCenter/AppCenter/MSAppCenter.m
+++ b/AppCenter/AppCenter/MSAppCenter.m
@@ -305,8 +305,10 @@ static NSString *const kMSGroupId = @"AppCenter";
     }
     if (service.isAppSecretRequired) {
 
-      // All services that require app secret cannot be started from library. This condition should be checked before
-      // app secret validation.
+      /*
+       * All services that require app secret cannot be started from library. This condition should be checked before
+       * app secret validation.
+       */
       if (!fromApplication) {
         MSLogError([MSAppCenter logTag], @"Cannot start service %@. The service cannot be started from a library.",
                    clazz);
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -309,13 +303,25 @@ - (BOOL)startService:(Class)clazz
...
+        return NO;
+      }
 
       // Service requires an app secret but none is provided.
-      MSLogError([MSAppCenter logTag],
-                 @"Cannot start service %@. App Center was started without app secret, but the service requires it.",
-                 clazz);
-      return NO;
+      if (![self.appSecret length]) {
+        MSLogError([MSAppCenter logTag],
```
TroubleMakerBen: Do we really want to log this as error? I'm wondering if we should log this as a warning.

jwallra: Also, this will change with #1033 and no longer be an error

jaelim-ms: https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java#L731

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -286,12 +283,12 @@ - (NSArray *)sortServices:(NSArray<Class> *)services {
...
 }
 
-- (BOOL)startService:(Class)clazz
-       withAppSecret:(NSString *)appSecret
-          andSendLog:(BOOL)sendLog
-     fromApplication:(BOOL)fromApplication {
+- (BOOL)startService:(Class)clazz andSendLog:(BOOL)sendLog fromApplication:(BOOL)fromApplication {
   @synchronized(self) {
 
+    // If it is started from library, it shouldn't use app secret.
```
clpolet: Nit. Just to be consistent we should do the same with the target token. Since library can't start with a token either.

jaelim-ms:
Submitting a new PR with different approach.

PR #1041, opened 2018-07-06
Use "ingestion" instead of "sender" throughout the SDK

TroubleMakerBen:
- Rename `*sender*` classes, protocols and enums
- Rename methods containing `*sender*`
- Fix comments containing `*sender*`

PR #1042, opened 2018-07-10
Use __unused instead of (void)ingestion for MSIngestionDelegate callbacks

TroubleMakerBen:
@jaelim-ms requested this in #1041 - It's probably okay but I don't like that we do this. Marking the param as `__unused` means that classes that implement the callback cannot use the `ingestion` param at all.
Imho the protocol should leave it to the implementation to determine if the param is used or not.


jaelim-ms:
I meant `__unused` only in concrete classes, not in protocols. Sorry for any confusions.
BTW, why did you submit a new PR? This one contains all of #1041

PR #1043, opened 2018-07-11
Adds common property feature to transmission targets

clpolet:


AppCenter/AppCenterTests/Util/MSMockUserDefaults.m
```diff
@@ -16,11 +15,6 @@ - (instancetype)init {
   self = [super init];
   if (self) {
     _dictionary = [NSMutableDictionary new];
-    _mockNSUserDefaults = OCMClassMock([NSUserDefaults class]);
-    OCMStub([_mockNSUserDefaults objectForKey:OCMOCK_ANY]).andCall(self, @selector(objectForKey:));
```
clpolet: We don't need all the stubbing, tests uses the mock methods directly since its instance is returned by [_mockMSUserDefaults shared] + the SDK uses the MSUserDefaults class instead of the NSUserDefaults. We were facing some weird errors on this mock so we cleaned it up.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -24,13 +25,33 @@ - (instancetype)initWithTransmissionTargetToken:(NSString *)token
   return self;
 }
 
+- (void)setEventPropertyString:(NSString *)propertyValue forKey:(NSString *)propertyKey {
+  @synchronized([MSAnalytics sharedInstance]) {
+    if (!propertyValue || !propertyKey) {
```
jaelim-ms: Value could be null.

AppCenter/AppCenterTests/Util/MSMockUserDefaults.m
jaelim-ms: Looks it is fine since all unit tests are succeeded but I guess the initial approach was to class mock `NSUserDefaults` that is used inside `MSUserDefaults`.

PR #1048, opened 2018-07-16
Feature/add scenario to sasquatch

achocron:
WIP. Changes so far:

1. Moved transmission target logic to its own page to simplify other pages
2. Removed Push page and made it a cell on the main page (it was weird to have an entire page dedicated to a single switch)
3. Flattened UI by using tab view controller instead of navigation. Motivation was to make it easier to quickly switch between Analytics and Transmission pages because they have properties that depend on each other.
4. Change UI elements that force restart to just warn that they won't take effect until next launch
5. Change segmented control for selecting a startup target to checkbox cells
6. With the "Analytics Events" switch, track event calls can be enabled/disabled per target (including default)
7. Probably other minor changes that I'm forgetting to mention

I've attached some screen shots but I recommend running the app locally to test it out.
screenshotscreenshotscreenshotscreenshot


jwallra:
Please add screenshots of the changes when this is ready for review :)

TroubleMakerBen:
This is a lot better than the old one!

TroubleMakerBen:
I can't repro the crash anymore. Looking good.

PR #1055, opened 2018-07-19
Override common schema Part A properties

heya-msft:
Implemented overriding Part A properties in the transmission target using PropertyConfigurator as a new channel delegate. Override will do nothing when the transmission target is disabled.


AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -487,4 +510,191 @@ - (void)testEventPropertiesCascading {
   XCTAssertEqual(eventLog.properties[@"g"], @"7777");
 }
 
+- (void)testOverridingCommonSchemaProperties {
```
jwallra: I don't particularly like this style of writing tests.  I would like for this to be split up into more discrete test cases.

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsTransmissionTargetPrivate.h
```diff
@@ -27,6 +28,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property(nonatomic, readonly) NSString *isEnabledKey;
 
+/**
+ * The channel group.
+ */
+@property(nonatomic) id<MSChannelGroupProtocol> channelGroup;
```
clpolet: We can probably make this one readonly too :)

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
```diff
@@ -0,0 +1,32 @@
...
+#import "MSChannelDelegate.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSPropertyConfigurator : NSObject <MSChannelDelegate>
+
+/**
+ * Override the application version.
+ *
+ * @param appVersion New application version for a tranmission target.
```
clpolet: *transmission

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+- (void)setAppVersion:(NSString *)appVersion;
+
+/**
+ * Override the application name.
+ *
+ * @param appName New application name for a tranmission target.
```
clpolet: *transmission

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+- (void)setAppName:(NSString *)appName;
+
+/**
+ * Override the application locale.
+ *
+ * @param appLocale New application locale for a tranmission target.
```
clpolet: *transmission

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.m
```diff
@@ -0,0 +1,64 @@
...
+- (void)setAppLocale:(NSString *)appLocale {
+  _appLocale = appLocale;
+}
+
+#pragma mark - MSChannelDelegate
+
+- (void)channel:(id<MSChannelProtocol>)__unused channel prepareLog:(id<MSLog>)log {
+  if ([log isKindOfClass:[MSCommonSchemaLog class]] && [self.transmissionTarget isEnabled]) {
+
+    // Override the application version.
```
clpolet: I was thinking that there is probably a way to share code using [KVC](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueCoding/index.html). Since it's all about updating `propertyConfigurator` properties. But I'll let you decide. Not a blocker.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -21,6 +26,9 @@ - (instancetype)initWithTransmissionTargetToken:(NSString *)token
     if (![self isImmediateParent]) {
       [MS_USER_DEFAULTS setObject:@(NO) forKey:self.isEnabledKey];
     }
+
+    // Add property configurator to the channel group as a delegate.
+    [_channelGroup addDelegate:_propertyConfigurator];
```
TroubleMakerBen: I approved but had a question about this that Jae answered in slack.
I don't like how the TransmissionTarget object holds onto the channel group reference and modifies it (it adds a new delegate). I understand that the PropertyConfigurator is specific to a TransmissionTarget but all other options that come to mind feel dirty, too.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -487,4 +510,198 @@ - (void)testEventPropertiesCascading {
...
```
clpolet: Should we add a test for when there is nothing to override, no properties set on the propertyConfigurator ? Also for now the SDK doesn't set the app name, should we add a test case for when the app name is not set but still overridden by customer?

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
jaelim-ms: Yes. This is not the best but better than all other possible options.

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
jaelim-ms: Oops.

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
jaelim-ms: Ooops.

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
jaelim-ms: Oooops.

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.m
jaelim-ms: Same discussion happened for Android implementation. For now we only have three so it is kinda okay but we should definitely consider better option when we add more properties here. Won't fix just for now.

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.h
TroubleMakerBen: oops

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsTransmissionTargetPrivate.h
```diff
@@ -27,6 +28,16 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ * Property configurator.
+ */
+@property(nonatomic, readonly) MSPropertyConfigurator *propertyConfigurator;
```
clpolet: Shouldn't this property be public?

jaelim-ms: Great catch!!

AppCenterAnalytics/AppCenterAnalytics/Internals/PropertyConfigurator/MSPropertyConfigurator.m
heya-msft: @clpolet good point.  We don't like that many while loops either.  We can add a TODO for now until we come up with something better.

PR #1062, opened 2018-07-23
Fix 1ds logs enqueue from log buffer

MatkovIvan:
OC logs from the buffer are currently queued to an AC channel which is wrong.

The main idea of the PR: in the 1ds delegate re-enqueue 1ds logs if they aren't in 1ds channel

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -423,7 +423,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
   _enableMachExceptionHandler = enableMachExceptionHandler;
 }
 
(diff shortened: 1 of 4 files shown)


jwallra:
Please add a description

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -439,15 +439,6 @@ - (void)channel:(id<MSChannelProtocol>)__unused channel
      didPrepareLog:(id<MSLog>)log
     withInternalId:(NSString *)internalId {
 
-  /*
-   * FIXME: Don't buffer OC logs yet. OC logs from the buffer are currently queued to an AC channel which is wrong. It
```
jwallra: This looks like the gist of what the description should be, no?

PR #1064, opened 2018-07-23
Clean up test apps

TroubleMakerBen:
* Removes Puppet
* Add SasquatchPuppet target to Sasquatch that is based on SasquatchObjC but uses project references
* Move UI Tests from Puppet to SasquatchPuppet (but does not fix them, they are broken)
* Does NOT add SasquatchPuppet to the TestApps project


TestApps/Sasquatch/Sasquatch/AppCenterDelegate.swift
```diff
@@ -1,4 +1,7 @@
+#if PUPPET
```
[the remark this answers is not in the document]
jwallra: Hmm, that's weird.  Still could use the shortened logic :)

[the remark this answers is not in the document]
TroubleMakerBen: I changed it to `PUPPET` so the `GCC_PREPROCESSOR_DEFINITIONS` variable and the active compilation condition variable both match.

Sasquatch/Sasquatch.xcodeproj/project.pbxproj
```diff
@@ -1399,6 +2201,135 @@
...
+				DEFINES_MODULE = NO;
+				DEVELOPMENT_TEAM = "$(inherited)";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(SRCROOT)/../Vendor/iOS/**",
+					"$(SRCROOT)/../CrashLib/CrashLib",
+				);
+				GCC_PREPROCESSOR_DEFINITIONS = (
+					"DEBUG=1",
+					"$(inherited)",
+					"GCC_PREPROCESSOR_MACRO_PUPPET=1",
```
jaelim-ms: Is this a naming convention, `GCC_PREPROCESSOR_MACRO`?

Sasquatch/Sasquatch.xcodeproj/project.pbxproj
```diff
@@ -1399,6 +2201,135 @@
...
+				);
+				INFOPLIST_FILE = "$(SRCROOT)/../TestApps/Sasquatch/SasquatchObjC/Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
+				PRODUCT_BUNDLE_IDENTIFIER = com.microsoft.appcenter.sasquatchPuppet;
```
jaelim-ms: All lower-cases.

Sasquatch/Sasquatch.xcodeproj/project.pbxproj
TroubleMakerBen: I can't use a preprocessor macro for swift code because swift doesn't have a preprocessor. And the new `SWIFT_ACTIVE_COMPILATION_CONDITION` is swift only. The option would be to have both flags named `PUPPET`, have `IS_PUPPET` (like in earlier versions of this PR) or be more explicit and I wanted the flags to be as explicit as possible without writing documentation somewhere.

TroubleMakerBen:
FYI CI is expected to fail as I need to update our CI to use SasquatchPuppet instead of Puppet.

Sasquatch/Sasquatch.xcodeproj/project.pbxproj
jaelim-ms: My questions wasn't on GCC preprocessor macro. Just the name of the macro. 😄

TroubleMakerBen:
@MatkovIvan cherry-picked your changes – I only saw your comment after I started cleaning up myself 🤦‍♂️ – thx for this!

PR #1079, opened 2018-07-30
Fixes post clone script

TroubleMakerBen:
The script did a `cd ..` which is no longer necessary as it now lives in the root directory.


AppCenter.xcworkspace/contents.xcworkspacedata
```diff
@@ -45,7 +45,7 @@
       location = "group:AppCenter.podspec">
    </FileRef>
    <FileRef
-      location = "group:app-center-post-clone.sh">
+      location = "group:/Users/benny/Repositories/MS/mobile-center-sdk-ios/appcenter-post-clone.sh">
```
jaelim-ms: ???

PR #1087, opened 2018-08-15
Fix/make auth api asynchronous

TroubleMakerBen:
Makes the auth API truly asynchronous.

- Adds a delegate that needs to be implemented
- delegate protocol requires one method `acquireTokenWithCompletionHandler:` to be implemented
- the completion block needs to be called during authentication flow and takes two params

Usage looks like this:

```swift

  func acquireToken(completionHandler : @escaping MSAnalyticsAuthenticationProviderCompletionBlock) {

// some auth flow
if success {
   completionHandler(authToken, expiryDate))
}
else {
   completionHandler(nil, nil))
}

}
```


AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsAuthenticationProviderDelegate.h
```diff
@@ -0,0 +1,24 @@
...
+@class MSAnalyticsAuthenticationProvider;
+
+/**
+ * Completion handler that returns the authentication token and the expiry date.
+ */
+typedef void (^MSAnalyticsAuthenticationProviderCompletionBlock)(
+    NSString *token, NSDate *expiryDate);
+
+@protocol MSAnalyticsAuthenticationProviderDelegate <NSObject>
+/**
```
guperrot: Add a blank line before the comment

PR #1109, opened 2018-09-08
Feature/update test app to collect device

achocron:


Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -1,94 +1,99 @@
...
-      
-      // Set cell to respond to being edited.
-      cell!.valueField.addTarget(self, action: #selector(propertyValueChanged), for: .editingChanged)
-      cell!.valueField.addTarget(self, action: #selector(dismissKeyboard), for: .editingDidEndOnExit)
-      return cell!
+    case kDeviceIdRow:
+        let cell = tableView.dequeueReusableCell(withIdentifier: switchCellIdentifier)!
+        let switcher: UISwitch? = cell.getSubview()
+        let selectedTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget()!
+        switcher!.isOn = collectDeviceIdStates[selectedTarget!]!
```
MatkovIvan: ` Fatal error: Unexpectedly found nil while unwrapping an Optional value`

ElektrojungeMS: I fixed the crash and cleaned up the underlying implementation to fix the runtime target tokens for Sasqutch ObjC and Puppet.

jwallra:
Please add screenshots

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -1,94 +1,102 @@
...
-    }
-    
-    switch tableIndex {
-    case CommonSchemaProperty.AppName.rawValue:
+    propertyValues[selectedTarget!]![propertyIndex] = sender.text!
+    switch propertyIndex {
+    case 0: // App Name.
       target.propertyConfigurator.setAppName(sender.text!)
       break
-    case CommonSchemaProperty.AppVersion.rawValue:
```
jwallra: Why are we now hard-coding these values?  We shouldn't do this if we can avoid it...

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -1,94 +1,102 @@
...
+
+  override func propertyAtRow(row: Int) -> (String, String) {
+    let selectedTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget()
+    let propertyIndex = row - propertyCellOffset()
+    let value = propertyValues[selectedTarget!]![propertyIndex]
+    return (propertyKeys[row - numberOfCustomHeaderCells()], value)
+  }
+
+  override func numberOfCustomHeaderCells() -> Int {
+    return 2
```
jwallra: Here's another magic number... can we codify this somehow?

guperrot:
The check box for collect deviceId appears with a greyed style though we can actually toggle it on/off.

screen shot 2018-09-10 at 2 01 53 pm

PR #1110, opened 2018-09-11
Fix issues in the sasquatch apps

ElektrojungeMS:
* Makes sure SasquatchSwift-Preview uses the SasquatchSwift runtime token
*  Make sure the UI doesn't look like it's possible to enable a child target with the runtime target (parent) disabled

### DIFF
--- a/Sasquatch/Sasquatch/ViewControllers/MSTransmissionTargetsViewController.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/MSTransmissionTargetsViewController.swift
@@ -177,7 +177,10 @@ class MSTransmissionTargetsViewController: UITableViewController {
   func targetEnabledSwitchValueChanged(sender: UISwitch!) {
     let sectionIndex = getCellSection(forView: sender)
     let section = transmissionTargetSections![sectionIndex]
-    if (sectionIndex == Section.Runtime.rawValue) {
+    if (sectionIndex == Section.Default.rawValue) {
+      section.setTransmissionTargetEnabled(sender!.isOn)
+    }
+    else if (sectionIndex == Section.Runtime.rawValue) {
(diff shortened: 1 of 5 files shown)


Sasquatch/Sasquatch/ViewControllers/MSTransmissionTargetsViewController.swift
```diff
@@ -183,6 +192,17 @@ class MSTransmissionTargetsViewController: UITableViewController {
...
     }
+    else if (sectionIndex == Section.Child1.rawValue || sectionIndex == Section.Child2.rawValue) {
+      let switchEnabled = sender!.isOn
+      section.setTransmissionTargetEnabled(switchEnabled)
+      if(switchEnabled && (section.isTransmissionTargetEnabled() == false)) {
+        
+        // Switch tried to enable the transmission target but it didn't work.
+        sender!.setOn(false, animated: true)
+        section.setTransmissionTargetEnabled(false)
+        sender?.isEnabled = false
```
MatkovIvan: I guess sender can't be `null` here (`!` is used above).

Sasquatch/Sasquatch/ViewControllers/MSTransmissionTargetsViewController.swift
```diff
@@ -134,9 +140,9 @@ class MSTransmissionTargetsViewController: UITableViewController {
       let switcher: UISwitch? = cell.getSubview()
       switcher?.isOn = section.isTransmissionTargetEnabled()
       switcher?.addTarget(self, action: #selector(targetEnabledSwitchValueChanged), for: .valueChanged)
-
+
```
jwallra: Remove extra whitespace / format ;)

PR #1112, opened 2018-09-11
Fix validation of nested event properties for CSLogs

ElektrojungeMS:
We are validating properties in the SDK because we want to make sure they are string values and keys only. Unfortunately the implementation of this didn't take nested properties into account. 
This PR also brings the validation logic closer to the ACLog validation log logic. Invalid properties are now stripped but the event will still be sent. In the past, if a common schema log contained an invalid property, the entire event was discarded.


ElektrojungeMS:
Codecov doesn't really make sense here -.-

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -209,16 +209,37 @@ - (BOOL)validateLogName:(NSString *)name {
...
-  NSDictionary<NSString *, NSString *> *properties = data.properties;
+  return [self validateProperties:data.properties];
+}
+
+- (BOOL)validateProperties:(NSDictionary<NSString *, id> *) properties {
+  BOOL valueIsAString;
   for (NSString *key in properties) {
-    if (![key isKindOfClass:[NSString class]] ||
-        ![properties[key] isKindOfClass:[NSString class]]) {
+    BOOL keyIsNSString = [key isKindOfClass:[NSString class]]; // is this actyally a case?!
```
jwallra: Do you actually want this comment here?  Maybe a more general comment on why this is or shouldn't be a case would be more helpful

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegatePrivate.h
```diff
@@ -67,4 +67,14 @@ extern NSString *const kMSLogNameRegex;
  * @return YES if data is valid, NO otherwise.
  */
 - (BOOL)validateLogData:(MSCSData *)data;
+
+/**
+ * Validate a dictionary that contains properties for a common schema log.
+ *
+ * @param properties The properties.
+ *
```
jwallra: In android we don't have blank lines between `@param` and `@return` ... Is this auto-formatted already?

ElektrojungeMS: We don't have a convention around this, we sometimes have and sometimes don't have one. My preference would be to have `@param` and `@return` separated be an extra line.

guperrot:
@ElektrojungeMS looks some tests fail.

Can we also increase code coverage a bit on this PR so that it does not keep going down?

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -212,16 +212,37 @@ - (BOOL)validateLogName:(NSString *)name {
...
+}
+
+- (BOOL)validateProperties:(NSDictionary<NSString *, id> *) properties {
+  BOOL valueIsAString;
   for (NSString *key in properties) {
-    if (![key isKindOfClass:[NSString class]] ||
-        ![properties[key] isKindOfClass:[NSString class]]) {
+    BOOL keyIsNSString = [key isKindOfClass:[NSString class]];
+    BOOL valueIsADictionary = [properties[key] isKindOfClass:[NSDictionary class]];
+    valueIsAString = [properties[key] isKindOfClass:[NSString class]];
```
jaelim-ms: This line can be placed below line 235.

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegatePrivate.h
jaelim-ms: We do have a convention for this one @ElektrojungeMS and it follows the exact convention now. 😄

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -212,16 +212,37 @@ - (BOOL)validateLogName:(NSString *)name {
...
-                 @"%@ Properties key and value must be of type NSString.",
+                 @"%@ Property key must be of type NSString.",
                  kMSBaseErrorMsg);
       return NO;
     }
+    else {
+      if(valueIsADictionary) {
+        
+        // Properties can be nested, verify the nested properties here.
+        return [self validateProperties:properties[key]];
```
jaelim-ms: This will stop validating remaining key/value pairs. You should only return `NO` if nested property validation failed. Otherwise, you should continue the loop.

ElektrojungeMS: Good catch! thx!

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -211,18 +202,32 @@ - (BOOL)validateLogName:(NSString *)name {
...
   for (NSString *key in properties) {
-    if (![key isKindOfClass:[NSString class]] ||
-        ![properties[key] isKindOfClass:[NSString class]]) {
-      MSLogError([MSAppCenter logTag],
-                 @"%@ Properties key and value must be of type NSString.",
-                 kMSBaseErrorMsg);
-      return NO;
+    BOOL keyIsValid = NO;
+    if ([key isKindOfClass:[NSString class]] && key.length > 0) {
+      keyIsValid = YES;
```
jaelim-ms: If a key isn't valid, you can drop the key and continue. I think you can simply `continue` here.

jaelim-ms: I think it might be a miscommunication.
What I meant was that you don't even need `keyIsValid` variable.
```
if ([key isKindOfClass:[NSString class]] && key.length > 0) {
   id value = properties[key];
   if (value) {
       // Not checking for empty string, as values can be empty strings.
      if ([value isKindOfClass:[NSString class]]) {
          [validProperties setValue:value forKey:key];
      } else if ([value isKindOfClass:[NSDictionary class]]) {
        NSDictionary *nestedValidProperties = [self validateProperties:value];
        [validProperties setValue:nestedValidProperties forKey:key];
      }
   }
}

File: AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m [hunk 6]
```diff
@@ -211,18 +202,35 @@ - (BOOL)validateLogName:(NSString *)name {
...
+    else {
+      continue;
+    }
+    id value = properties[key];
+    if (value) {
+
+      // Not checking for empty string, as values can be empty strings.
+      if ([value isKindOfClass:[NSString class]]) {
+        if (keyIsValid) {
+          [validProperties setValue:value forKey:key];
```
achocron: Can you use `validPoperties[key] = value;` here?

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -211,18 +202,35 @@ - (BOOL)validateLogName:(NSString *)name {
...
+        }
+      } else if ([value isKindOfClass:[NSDictionary class]]) {
+        NSDictionary *nestedValidProperties = [self validateProperties:value];
+        if (keyIsValid) {
+          [validProperties setValue:nestedValidProperties forKey:key];
```
achocron: Can you use `validPoperties[key] = nestedValidProperties;` here?

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -260,15 +238,38 @@ - (void)trackEvent:(NSString *)eventName
...
+    // We have a valid key, so let's validate the value.
+    id value = properties[key];
+    if (value) {
+
+      // Not checking for empty string, as values can be empty strings.
+      if ([value isKindOfClass:[NSString class]]) {
+        [validProperties setValue:value forKey:key];
+      }
+    } else {
+      MSLogError([MSAnalytics logTag], @"Event property contains an invalid value, dropping the property.");
```
jaelim-ms: Can we also log `key`?

jaelim-ms: Warning, not error.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### AppCenterAnalytics
 
 * **[Feature]** Preparation work for a future change in transmission protocol and endpoint for Analytics data. There is no impact on your current workflow when using App Center.
+* **[Fix]** Fix an bug where nested custom properties for an event would not pass validation.
```
MatkovIvan: Do we include 1ds fixes in the GitHub change log?

ElektrojungeMS: If you ask me, yes. We're not 100% consistent there but I included it to make writing release notes for the next release easier.

PR #1113, opened 2018-09-12
Update startup mode UI in the test app

MatkovIvan:
simulator screen shot - iphone 6 - 2018-09-12 at 16 50 53 simulator screen shot - iphone 6 - 2018-09-12 at 16 50 58

### DIFF
--- a/Sasquatch/SasquatchObjC/AppDelegate.m
+++ b/Sasquatch/SasquatchObjC/AppDelegate.m
@@ -72,25 +72,27 @@ enum StartupMode {
   NSArray<Class> *services = @[
     [MSAnalytics class], [MSCrashes class], [MSDistribute class], [MSPush class]
   ];
-  NSInteger startTarget = [[NSUserDefaults standardUserDefaults] integerForKey:kMSStartTargetKey];
+  NSInteger startTarget =
+      [[NSUserDefaults standardUserDefaults] integerForKey:kMSStartTargetKey];
   switch (startTarget) {
   case APPCENTER:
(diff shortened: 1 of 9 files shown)


guperrot:
@MatkovIvan can we add a few more tests even if not related to the PR to increase it globally as opposed to decrease it?

guperrot:
Can we remove check box for default transmission target and display token like others?

MatkovIvan:
@guperrot

If you mean code coverage we can revisit it in the separate PR because this one doesn't touch SDK code at all, but only the test app. JFYI coverage always shows unstable results here.

I especially add the disabled checkbox to keep ability to show the current state.

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -59,46 +68,32 @@ - (BOOL)application:(UIApplication *)application
   [MSAppCenter setLogLevel:MSLogLevelVerbose];
 
   // Start App Center SDK.
-  BOOL useOneCollector = [[NSUserDefaults standardUserDefaults]
-      boolForKey:kMSOneCollectorEnabledKey];
-  long startTarget =
+  [MSAppCenter startFromLibraryWithServices:@[ [MSAnalytics class] ]];
```
guperrot: Should be moved to transmission tab so that we can check Analytics mode appears disabled when we go to Analytics tab directly, like in the Android and React Native apps.

guperrot: Can you do another PR for that but leave vsts task open? I will merge this increment.

MatkovIvan: #1117

PR #1117, opened 2018-09-17
Move startFromLibrary to transmission tab

MatkovIvan:


Sasquatch/SasquatchObjC/AppCenterDelegateObjC.m
```diff
@@ -29,58 +29,76 @@ @implementation AppCenterDelegateObjC
...
 - (BOOL)isDebuggerAttached {
   return [MSAppCenter isDebuggerAttached];
 }
+
 - (void)setCustomProperties:(MSCustomProperties *)customProperties {
   [MSAppCenter setCustomProperties:customProperties];
 }
 
+- (void)startAnalyticsFromLibrary {
+  [MSAppCenter startFromLibraryWithServices:@[ [MSAnalytics class] ]];
```
jaelim-ms: Not a big deal but can you take a list of services as a parameter?
For now we only have Analytics that can be started from library but let's make is flexible.

MatkovIvan: The problem is we cannot directly refer SDK services from shared code when we are building it as puppet (without introducing special bridge header). Even more, I believe that we shouldn't do it because then there would be no point in splitting into multiple applications.

ElektrojungeMS: I agree with Ivan.

jaelim-ms: I see.

PR #1119, opened 2018-09-19
Update test App to set storage size

MatkovIvan:
simulator screen shot - iphone 6 - 2018-09-19 at 18 47 37


ElektrojungeMS:
There's an issue with the text field to change the storage size, it cannot be dismissed.

simulator screen shot - iphone5 ios 9 0 - 2018-09-24 at 09 55 55

Sasquatch/SasquatchSwift/AppDelegate.swift
```diff
@@ -29,6 +29,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
...
     MSAppCenter.setLogLevel(MSLogLevel.verbose)
 
+    // Set max storage size.
+    let storageMaxSize = UserDefaults.standard.integer(forKey: kMSStorageMaxSizeKey)
+    if storageMaxSize > 0 {
+      MSAppCenter.setMaxStorageSize(storageMaxSize, completionHandler: { success in
+        if !success {
+          DispatchQueue.main.async {
+            let alertController = UIAlertController(title: "Warning!",
+                                                    message: "The maximum size of the internal storage wasn't set.",
```
ElektrojungeMS: Can you change this to say `...could not be set?`

Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -40,22 +49,42 @@ class MSMainViewController: UITableViewController, AppCenterProtocol {
...
     _ = MSTransmissionTargets.shared
 
+    // Storage size section.
+    let storageMaxSize = UserDefaults.standard.integer(forKey: kMSStorageMaxSizeKey)
+    self.storageMaxSizeField.text = "\(storageMaxSize / 1024)"
+    self.storageMaxSizeField.addTarget(self, action: #selector(storageMaxSizeUpdated(_:)), for: .editingChanged)
+    if let supportDirectory = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
+      let dbFile = supportDirectory.appendingPathComponent("com.microsoft.appcenter").appendingPathComponent("Logs.sqlite")
+      func getFileSize(_ file: URL) -> Int {
+        return (try? file.resourceValues(forKeys:[.fileSizeKey]))?.fileSize ?? 0
```
clpolet: I think the default max size printed shouldn't be zero, the default max size from the SDK would be ideal but we won't access it from all the Sasquatch app  so I would say we should print a `-` or anything but a number. That way we know it hasn't been set.

ElektrojungeMS:
If a invalid value is set, e.g. 0, it means that the max size of the DB is the default (10MiB). Can you refresh the UI on `setMaxStorageSize` failing so it displays the default size, like Android does in 
https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java#L249?

MatkovIvan:
@ElektrojungeMS Done, but the Android test app sets the actual file size instead of default size in this case. Should we update the Android test app to expected behavior?

guperrot:
@ElektrojungeMS we never want to display 0 on Android as it's very confusing and not what actually happens on failure. If file size is 32KB, you want to set maximum size to 20KB, then Android framework actually set size to 32KB.

PR #1120, opened 2018-09-20
Add api to set maxStorageSize

ElektrojungeMS:
* adds `setMaxStorageSize:completionhandler:` which needs to be called before start
* the API can be only called once
* default size is 10MiB
* minimum upper size limit is 20KiB
* page size is same as Android SDK
* completion handler can be `nil`
* the `completionHandler` param is passed on to the storage implementation which uses `dispatch_once` to make sure it's only ever called once. I was debating about having `MSDBStorage` call the completionHandler on `MSAppCenter` but I think it really needs to be captured in a property and called where it belongs. The fact that `MSAppCenter` captures the block in a property, too, is simply because we need to hold onto it until `start:` has been called.


ElektrojungeMS:
Added the don't merge label because one unit test needs to be fixed/implemented.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.h
```diff
@@ -86,6 +92,20 @@ static NSString *const kMSSQLiteConstraintAutoincrement = @"AUTOINCREMENT";
...
 
+/**
+ * Set the maximum size of the internal storage. This method must be called before App Center is started.
+ *
+ * @discussion This only sets the maximum size of the database, but App Center modules might store additional data.
+ * The value passed to this method is not persisted on disk. The default maximum database size is 10485760 bytes (10
+ * MiB).
+ *
+ * @param sizeInBytes Maximum size of in bytes. This will be rounded up to the nearest multiple of 4096. Values below
+ * 2097152 (2 MiB) will be ignored.
```
jwallra: Comment needs to be updated to 20KiB.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
```diff
@@ -17,15 +17,13 @@ static short const kMSHTTPMinGZipLength = 1400;
...
+typedef NS_ENUM(NSInteger, MSHTTPCodesNo) { // Informational
+      MSHTTPCodesNo1XXInformationalUnknown = 1,
   MSHTTPCodesNo100Continue = 100,
-  MSHTTPCodesNo101SwitchingProtocols = 101,
-  MSHTTPCodesNo102Processing = 102,
+  MSHTTPCodesNo101SwitchingProtocols = 101, MSHTTPCodesNo102Processing = 102,
 
   // Success
-  MSHTTPCodesNo2XXSuccessUnknown = 2,
+      MSHTTPCodesNo2XXSuccessUnknown = 2,
```
jwallra: There's some weird formatting changes in this file...

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -407,6 +411,33 @@ - (void)setLogUrl:(NSString *)logUrl {
   }
 }
 
+- (void)setMaxStorageSize:(long)sizeInBytes completionHandler:(void (^)(BOOL))completionHandler {
```
jwallra: I think it makes sense to prevent small values < 20KiB at this point instead of later.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -231,8 +211,75 @@ - (void)deleteDatabase {
...
+    return NULL;
+  }
+  MSLogVerbose([MSAppCenter logTag], @"Opened database.");
+  NSString *statement = [NSString stringWithFormat:@"PRAGMA max_page_count = %i;", maxPageCount];
+  char *errorMessage;
+  *result = sqlite3_exec(db, [statement UTF8String], NULL, NULL, &errorMessage);
+  if (*result != SQLITE_OK) {
+    errorMessage = errorMessage ? errorMessage : "(nil)";
+    NSString *printableErrorMessage = [NSString stringWithCString:errorMessage encoding:NSUTF8StringEncoding];
+    MSLogWarning([MSAppCenter logTag], @"Failed to open database with specified"
```
dhei: This should be `MSLogError` instead of `MSLogWarning`, right?

AppCenter/AppCenter/Internals/Storage/MSDBStorage.h
ElektrojungeMS: Good catch, thx!

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -231,8 +211,75 @@ - (void)deleteDatabase {
...
+
+    // Call the callback for setting the maximum storage size only once.
+    dispatch_once(&setMaxStorageSizeOnceToken, ^{
+      if (self.maxStorageSizeCompletionHandler) {
+        self.maxStorageSizeCompletionHandler(NO);
+      }
+    });
+  } else {
+    MSLogVerbose([MSAppCenter logTag], @"Database has maximum page count of %i.", maxPageCount);
+
```
dhei: In Android we log a warning message like "Could change database size but is slightly larger than expected (next multiple of 4KB)..." when we set the max storage size to the next multiple of 4KB (see https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java#L530-L532).

For consistency purpose, we should do the same in iOS if we're not already doing that.

ElektrojungeMS: Good idea.

ElektrojungeMS: So I made a change to this https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1120/commits/8abe706a4351d465d0935246e567e7ea57ebbcf2

I'm not sure if we want to keep it as it will be called a lot. Note that iOS is not reusing one connection but opening/closing them all the time. This means the max size will be set time and again, so the warning will be logged a lot.
I am thinking of changing it to verbose, what do you think?

dhei: @ElektrojungeMS, okay, either `MSLogInfo` or `MSLogVerbose` should be fine.

AppCenter/AppCenter/MSAppCenter.m
achocron: Would you mind elaborating on that? The original rationale was that since the storage limitation is on the Storage side of things, which app center is aware of, the separation of responsibility is most clear if that limit is enforced by `MSLogDbStorage` class (there shouldn't be a noticeable performance difference either way).

Not opposed to the change if there's a good reason, just wondering though.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -414,19 +414,30 @@ - (void)setLogUrl:(NSString *)logUrl {
...
-        @"Unable to set storage size after the application has configured App"
-        "Center");
-    if (completionHandler) {
-      completionHandler(NO);
+  @synchronized (self) {
+    if (self.setMaxStorageSizeHasBeenCalled) {
+      MSLogWarning([MSAppCenter logTag], @"setMaxStorageSize:completionHandler: may only be called once per app "
+                                         "launch");
+      if (completionHandler) {
+        completionHandler(NO);
```
achocron: We can run into deadlocks by allowing the completion handler to get called inside of a lock (@clpolet fixed an issue a while back where that was the cause.) Can we move that code? Or call it on another thread?

achocron: Also, since we are interacting with the database, I think we should move most of the work on a background thread so that this call doesn't block. (We probably meant to do that but forgot to)

achocron: I guess since there's just one lock, it *shouldn't* deadlock (but there could be hidden interactions with another lock elsewhere in the code), but even so, if they do something time-consuming in the callback, then all other app center operations will be blocked for that time

AppCenter/AppCenter/MSAppCenter.m
dhei: The 20KiB size check is like parameter validation of the `setMaxStorageSize` API, so it make sense to have the validation logic here instead of further down in database class. I agree with @jwallra on this.
Android implementation does 20KiB check in the `setMaxStorageSize` implementation, though it's fine to have slightly different implementation on Android/iOS.

AppCenter/AppCenter/MSAppCenter.m
achocron: Or it can at least cause App Center APIs to take a long time

clpolet: Yes I agree, anything that's executing SQL statement should be done in a background thread.

clpolet: The `completionHandler` can surely be called at the end after the synchronized if we bring up some kind of success state.

AppCenter/AppCenter/Internals/Storage/MSDBStoragePrivate.h
```diff
@@ -4,6 +4,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 typedef int (^MSDBStorageQueryBlock)(void *);
 
+// 4 KiB.
+static const long kMSDefaultPageSizeInBytes = 4096;
+
+// 10 MiB.
+static const long kMSDefaultDatabaseSizeInBytes = 10 * 1024 * 1024;
```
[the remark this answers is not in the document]
MatkovIvan: It is computed at compile time and improves readability, so I guess not

AppCenter/AppCenter/MSAppCenter.m
jwallra: @achocron, @dhei basically has the crux of it, checking here is basically parameter validation, and we can improve performance because we already know here whether or not we can set the given size because it's part of restrictions of the API itself.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -349,11 +314,26 @@ - (NSUInteger)countLogs {
...
-                                   kMSLogTableName, kMSTargetTokenColumnName];
+    NSString *migrationQuery = [NSString stringWithFormat:@"ALTER TABLE \"%@\" ADD COLUMN \"%@\" "
+                                                          "TEXT", kMSLogTableName, kMSTargetTokenColumnName];
     [MSDBStorage executeNonSelectionQuery:migrationQuery inOpenedDatabase:db];
   }
 }
 
+#pragma mark - DB size
+
+- (void)setMaxStorageSize:(long)sizeInBytes completionHandler:(nullable void (^)(BOOL))completionHandler {
```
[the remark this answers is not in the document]
jwallra: That makes sense, but how do we use other DBs in iOS?  I'm just curious, if we use other non-log DBs, is there a possibility that we could want a similar API to `setMaxStorageSize` for those ones too?  Right now the API only works for LogDBs anyways.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupProtocol.h
```diff
@@ -46,6 +44,20 @@ addChannelUnitWithConfiguration:(MSChannelUnitConfiguration *)configuration
  */
 - (void)setLogUrl:(NSString *)logUrl;
 
+/**
+ * Set the maximum size of the internal storage. This method must be called
+ * before App Center is started.
+ *
+ * @discussion The default maximum database size is 10485760 bytes (10 MiB).
```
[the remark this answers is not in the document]
jaelim-ms: Won't fix since we have inconsistent order in all the files. Need to address the issue separately.

AppCenter/AppCenter/Internals/Ingestion/MSOneCollectorIngestion.m
```diff
@@ -1,48 +1,38 @@
...
 NSString *const kMSOneCollectorApiVersion = @"1.0";
 NSString *const kMSOneCollectorClientVersionKey = @"Client-Version";
-NSString *const kMSOneCollectorContentType =
-    @"application/x-json-stream; charset=utf-8";
+NSString *const kMSOneCollectorContentType = @"application/x-json-stream; charset=utf-8";
 NSString *const kMSOneCollectorLogSeparator = @"\n";
 NSString *const kMSOneCollectorTicketsKey = @"Tickets";
 NSString *const kMSOneCollectorUploadTimeKey = @"Upload-Time";
 
 @implementation MSOneCollectorIngestion
```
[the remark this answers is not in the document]
achocron: Just formatting.

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -346,44 +298,156 @@ - (void)testCount {
...
+}
+
+- (void)testCompletionHandlerCanBeNil {
+
+  // When
+  [self.sut setMaxStorageSize:kMSDefaultPageSizeInBytes completionHandler:nil];
+  [self addGuysToTheTableWithCount:100];
+
+  // Then
+  // Didn't crash.
```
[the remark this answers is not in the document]
achocron: It's synchronous so it should be captured.

PR #1122, opened 2018-09-20
Fix sending stored events when startFromLibrary was called

MatkovIvan:
## Repro steps:
- Run the app with AppCenter startup mode (WiFi is on)
- Turn off WiFi
- Click “Track event” on Analytics tab
- Switch to Skip startup mode
- Kill the app
- Turn on WiFi
- Run the app
- Start AppCenter from library (go to transmission tab)

## Result:
Event is not shown on the portal (even if AppCenter started with app secret later)

## Log error:
```
SasquatchSwift[2613:617899] [AppCenter] VERBOSE: -[MSHttpIngestion sendCallAsync:]_block_invoke/276 HTTP response received with status code: 400, payload:
Value cannot be null.

Parameter name: App-Secret
```


jwallra:
Please add a description

jaelim-ms:
I don't think this is correct because it will use an app-secret which might be changed after logs were persisted. Let's discuss this one offline and think about what would be the best option for the case.

AppCenter/AppCenter/Internals/Ingestion/MSHttpIngestion.m
```diff
@@ -96,13 +96,13 @@ - (id)initWithBaseUrl:(NSString *)baseUrl
 
 #pragma mark - MSIngestion
 
+- (BOOL)isReadyToSend {
```
jaelim-ms: You don't need to override this in `MSOneCollectorIngestion`?

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -229,6 +224,11 @@ - (void)flushQueue {
     return;
   }
 
+  // Ingestion is not ready.
+  if (!self.ingestion.isReadyToSend) {
+    return;
```
jaelim-ms: This will keep all pending logs in the persistence and let it keep trying batch if there are logs more than 50. We should think about how we handle logs in this case.

AppCenter/AppCenter/Internals/MSAppCenterPrivate.h
```diff
@@ -1,11 +1,13 @@
...
-#import "MSOneCollectorChannelDelegate.h"
 
-@interface MSAppCenter ()
+@class MSAppCenterIngestion;
+@class MSOneCollectorChannelDelegate;
 
-@property(nonatomic) id <MSChannelUnitProtocol> channelUnit;
+@interface MSAppCenter ()
 
+@property(nonatomic) MSAppCenterIngestion *ingestion;
```
jaelim-ms: I'm not sure this is a good idea to be placed in AppCenter class. @ElektrojungeMS @clpolet What do you guys think?

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -6,6 +6,7 @@
 #import "MSChannelUnitConfiguration.h"
 #import "MSDeviceTracker.h"
 #import "MSIngestionProtocol.h"
+#import "MSAppCenterIngestion.h"
```
jaelim-ms: Please sort

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -532,14 +513,16 @@ - (void)testNextBatchSentIfPendingQueueGotRoomAgain {
...
-  [sut setAppSecret:MS_UUID_STRING];
 
   // When
   [sut enqueueItem:[self getValidMockLog]];
 
   // Try to release one batch.
   dispatch_async(self.logsDispatchQueue, ^{
-    ingestionBlock([@(1) stringValue], 200, nil, nil);
+    XCTAssertNotNil(ingestionBlock);
+    if (ingestionBlock) {
```
jaelim-ms: Redundant. Any warnings for this one?

AppCenter/AppCenter/Internals/Ingestion/MSHttpIngestion.m
MatkovIvan: No, because target tokens are stored in the logs itself and we don't need to make additional actions to prepare ingestion for sending.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
MatkovIvan: No, batches will have the same size, because `loadLogsWithGroupId:limit:withCompletion:` has limit parameter. The PR doesn't change this logic.

AppCenter/AppCenter/Internals/MSAppCenterPrivate.h
MatkovIvan: It is already stored in the channel group, but we need to access to it from `MSAppCenter` class to update the app secret. Storing the pointer is better than get it for channel group and blindly cast protocol to implementation.

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
MatkovIvan: Disagree. This makes assert instead of test crash if something goes wrong. You can see the same usage in the similar places in the file.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
jaelim-ms: I meant kind of an edge because it doesn't clean up logs. But I agree that this doesn't change any behaviors, we are going to rethink about this one later.

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
jaelim-ms: Oh.. This is async block. Yes, keep it as it is, I missed just one live above.
=> If `ingestionBlock` is `nil`, it should stop the test at that time unless it is not running in the same thread.

AppCenter/AppCenter/Internals/MSAppCenterPrivate.h
jaelim-ms: Any reason to have a reference to Ingestion from AppCenter class in terms of design? I don't see any reasons, "better" is just for convenience.

clpolet: The other possibility is to forward the AppSecret to the channelGroup which will in turn set the AppSecret to its ingestion. Then no ingestion in AppCenter class.

ElektrojungeMS: @jaelim-ms and I are talking about this right now, too. We both think that it's wrong to have a reference to ingestion in `MSAppCenter`. We favor what @clpolet is suggesting, how about we add an API to `MSChannelGroupProtocol` to update the app secret. We can use `setAppSecret:(NSString *)appSecret` for now and if we need a more generic name in the future, we can rename the api as it's not public.

jaelim-ms: This will also align the design with Android. ChannelGroup is the one that keeps all information and instances, coordinate core service in it. ChannelGroup is the right place to keep app secret, it can receive an app secret as @ElektrojungeMS mentioned and then set it to Ingestion would be the best approach.

MatkovIvan: Thanks guys, I agree, it's better. 👍

PR #1131, opened 2018-09-25
Added UI for testing pause/resume API

bmourat:
| Resumed | Paused |
| --- |  --- |
| resumed| paused|


achocron:
Since we are going with the two button layout, can we also somehow highlight the last button that was pressed so you can know what state it's supposed to be in?

dhei:
> Since we are going with the two button layout, can we also somehow highlight the last button that was pressed so you can know what state it's supposed to be in?

I see only a "Set Paused" toggle, not any "resume" button or toggle from the [screenshoot](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1131#issue-217865344). How are you going to allow pressing pause/resume twice in a row in the UI? I think using a toggle for pause/resume is more confusing, as there's no such thing as toggle-off. Maybe use a button and changes it's color after pressed?

jwallra:
Yeah, this shouldn't be two toggles, since untoggling doesn't do anything.  It allows the scenario I was mentioning, but I agree with @dhei that it's confusing.

bmourat:
@dhei  what do you mean you see only "Set Paused" toggle? There are now two buttons: pause and resume and no toggle. Screenshots are from the previous version.

MatkovIvan:
@dhei @jwallra Here are two buttons

@bmourat Please update the screenshots

PR #1134, opened 2018-09-25
Fix message

achocron:


achocron:
@jaelim-ms @dhei Android sdk worded it differently by saying the new max size, but doing this in iOS is slightly more complicated. But the existing statement in Android seems useful so not sure how much we care about making them the same. Thoughts?

PR #1146, opened 2018-09-29
Test notification foreground and tap detection in objc app

guperrot:
Test app changes to verify https://github.com/MicrosoftDocs/appcenter-docs-pr/pull/875.
Will do swift after this one is approved to avoid having to address same feedbacks twice.

### DIFF
--- a/Sasquatch/SasquatchObjC/AppDelegate.m
+++ b/Sasquatch/SasquatchObjC/AppDelegate.m
@@ -27,23 +27,19 @@
 #define APP_SECRET_VALUE "3ccfe7f5-ec01-4de5-883c-f563bbbe147a"
 #endif
 
-enum StartupMode {
-  APPCENTER,
-  ONECOLLECTOR,
-  BOTH,
-  NONE,
-  SKIP
-};
+enum StartupMode { APPCENTER, ONECOLLECTOR, BOTH, NONE, SKIP };
 
 @interface AppDelegate () <
 #if GCC_PREPROCESSOR_MACRO_PUPPET
     MSAnalyticsDelegate,
 #endif
-    MSCrashesDelegate, MSDistributeDelegate, MSPushDelegate, UNUserNotificationCenterDelegate>
+    MSCrashesDelegate, MSDistributeDelegate, MSPushDelegate,
+    UNUserNotificationCenterDelegate>
 
 @property(nonatomic) MSAnalyticsResult *analyticsResult;
 
 @property(nonatomic) BOOL didReceiveNotificationInForeground;
+@property(nonatomic, copy) NSString *notificationResponse;
 @end
 
 @implementation AppDelegate
@@ -66,7 +62,8 @@ enum StartupMode {
 
   // Customize App Center SDK.
   if (@available(iOS 10.0, *)) {
-    UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+    UNUserNotificationCenter *center =
+        [UNUserNotificationCenter currentNotificationCenter];
     center.delegate = self;
   }
   [MSPush setDelegate:self];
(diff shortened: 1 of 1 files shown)


Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -383,7 +389,33 @@ - (BOOL)distribute:(MSDistribute *)distribute
...
+  completionHandler(UNNotificationPresentationOptionNone);
+}
+
+// iOS 10 and later, asks the delegate to process the user's response to a
+// delivered notification.
+- (void)userNotificationCenter:(UNUserNotificationCenter *)center
+    didReceiveNotificationResponse:(UNNotificationResponse *)response
+             withCompletionHandler:(void (^)(void))completionHandler
+    API_AVAILABLE(ios(10.0)) {
+  self.notificationResponse = [response actionIdentifier];
```
MatkovIvan: Since [UNNotificationDismissActionIdentifier](https://developer.apple.com/documentation/usernotifications/unnotificationdismissactionidentifier)
> is delivered only if the notification’s category object was configured with the customDismissAction option

and there are only 2 system actions + we are not supporting categories with custom actions, it looks redundant.
Lets just set flag to `NO` here:
```objc
self.didReceiveNotificationInForeground = NO;
```

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -383,7 +389,33 @@ - (BOOL)distribute:(MSDistribute *)distribute
...
+// When this callback is called, this disables the other callback that MSPush
+// handles.
+- (void)userNotificationCenter:(UNUserNotificationCenter *)center
+       willPresentNotification:(UNNotification *)notification
+         withCompletionHandler:
+             (void (^)(UNNotificationPresentationOptions options))
+                 completionHandler API_AVAILABLE(ios(10.0)) {
+  self.didReceiveNotificationInForeground = YES;
+  [MSPush didReceiveRemoteNotification:notification.request.content.userInfo];
+  completionHandler(UNNotificationPresentationOptionNone);
```
MatkovIvan: Let's expose it to UI to test system alert in the foreground case.

clpolet: What do you want to expose ?

Sasquatch/SasquatchObjC/AppDelegate.m
clpolet: Do you mean adding a new flag `self.userDidTapNotification` instead of the `notificationResponse` ?

Sasquatch/SasquatchObjC/AppDelegate.m
MatkovIvan: There is "Alert" option. In this case the callback is called twice - here and when user clicked

MatkovIvan: So, either picker or just switch "show system notification on foreground push"

Sasquatch/SasquatchObjC/AppDelegate.m
MatkovIvan: Just setting `didReceiveNotificationInForeground` to `NO` would be enough

Sasquatch/SasquatchObjC/AppDelegate.m
guperrot: I don't understand this feedback, can you rephrase?

Sasquatch/SasquatchObjC/AppDelegate.m
guperrot: If background the notification is always tapped.
If foreground the notification is never tapped.
We could indeed simplify this + the docs.
Let's keep the boolean for both app and docs.

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -26,22 +27,18 @@
...
     MSAnalyticsDelegate,
 #endif
-    MSCrashesDelegate, MSDistributeDelegate, MSPushDelegate>
+    MSCrashesDelegate, MSDistributeDelegate, MSPushDelegate,
+    UNUserNotificationCenterDelegate>
 
 @property(nonatomic) MSAnalyticsResult *analyticsResult;
 
+@property(nonatomic) BOOL didTapNotification;
 @end
```
jaelim-ms: Empty line before `@end`

PR #1136, opened 2018-09-26
Feature/pause resume analytics

achocron:
Added two APIs: `[MSAnalytics pause]` and `[MSAnalytics resume]`.

PR #1147, opened 2018-10-01
Feature/pause resume transmissiontarget api

jaelim-ms:
Still testing features on test app, the PR might contain some bugs or issues.
Will update the description after testing is done. In the meantime, giving reviewers earlier view for implementations.
The PR also contains all changes in #1136, please wait until the PR is merged.

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -1,6 +1,7 @@
 #import "MSChannelGroupDefault.h"
 #import "AppCenter+Internal.h"
 #import "MSAppCenterIngestion.h"
+#import "MSChannelUnitConfiguration.h"
 #import "MSChannelUnitDefault.h"
 #import "MSLogDBStorage.h"
 
@@ -55,6 +56,15 @@ static char *const kMSlogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
                               }];
   }
   return channel;
+}
+
+- (id<MSChannelUnitProtocol>)channelUnitForGroupId:(NSString *)groupId {
+  for (MSChannelUnitDefault *channel in self.channels) {
+    if ([channel.configuration.groupId isEqualToString:groupId]) {
+      return channel;
+    }
+  }
+  return nil;
 }
 
 #pragma mark - Delegate
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -28,7 +28,7 @@
     _discardLogs = NO;
     _delegates = [NSHashTable weakObjectsHashTable];
     _pausedIdentifyingObjects = [NSHashTable weakObjectsHashTable];
-    _pausedTargetKeys = [NSHashTable new];
+    _pausedTargetKeys = [NSMutableSet new];
   }
   return self;
 }
@@ -535,15 +535,15 @@
 
 - (void)pauseSendingLogsWithToken:(NSString *)token {
   NSString *targetKey = [MSUtility targetKeyFromTargetToken:token];
-  @synchronized (self.pausedTargetKeys) {
+  dispatch_async(self.logsDispatchQueue, ^{
     MSLogDebug([MSAppCenter logTag], @"Pause channel for target key %@.", targetKey);
     [self.pausedTargetKeys addObject:targetKey];
-  }
+  });
 }
 
 - (void)resumeSendingLogsWithToken:(NSString *)token {
   NSString *targetKey = [MSUtility targetKeyFromTargetToken:token];
-  @synchronized (self.pausedTargetKeys) {
+  dispatch_async(self.logsDispatchQueue, ^{
     MSLogDebug([MSAppCenter logTag], @"Resume channel for target key %@.", targetKey);
     [self.pausedTargetKeys removeObject:targetKey];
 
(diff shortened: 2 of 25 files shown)


achocron:
Looks good to me, but someone else should review because I contributed to this PR.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -54,6 +55,15 @@ - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion {
   return channel;
 }
 
+- (id<MSChannelUnitProtocol>)channelUnitForGroupId:(NSString *)groupId {
+  for (MSChannelUnitDefault *channel in self.channels) {
```
guperrot: nit: any reason its not a dictionary and just look up by key?

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -542,6 +533,26 @@ - (void)resumeWithIdentifyingObjectSync:(id <NSObject>)identifyingObject {
...
+}
+
+- (void)resumeSendingLogsWithToken:(NSString *)token {
+  NSString *targetKey = [MSUtility targetKeyFromTargetToken:token];
+  @synchronized (self.pausedTargetKeys) {
+    MSLogDebug([MSAppCenter logTag], @"Resume channel for target key %@.", targetKey);
+    [self.pausedTargetKeys removeObject:targetKey];
+
+    // Update item count and check logs if it meets the conditions to send logs.
+    self.itemsCount = [self.storage countLogs];
```
guperrot: It should be itemCount not itemsCount

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -9,7 +10,6 @@
 #import "MSOneCollectorIngestion.h"
 #import "MSSDKExtension.h"
 
-static NSString *const kMSOneCollectorGroupIdSuffix = @"/one";
 static NSString *const kMSOneCollectorBaseUrl =
```
guperrot: Can probably remove this TODO if we're not planning to address it.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -351,6 +359,26 @@ - (MSAnalyticsTransmissionTarget *)transmissionTargetForToken:(NSString *)transm
   return transmissionTarget;
 }
 
+- (void)pauseTransmissionTargetForToken:(NSString *)token {
+  if (self.oneCollectorChannelUnit) {
```
guperrot: if null means disabled? In Android we emit an error log when calling APIs while service is disabled

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
jaelim-ms: Yes, totally agree with you but not in the scope so we will refactor separately.

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
jaelim-ms: Not in scope for this PR but valid comment. Will do it separately.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
clpolet: Doing the count now works but ideally we should refresh the count every-time  we call `checkPendingLogs`. Basically querying for count within `checkPendingLogs` instead of the letting the channel doing the counting itself line 199. That way we can use `checkPendingLogs` everywhere instead of flush.

clpolet: Not a blocker but the count here will also count logs from still paused tokens.

jaelim-ms: Added a comment in the code.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
jaelim-ms: This is an implementation difference. Unfortunately Apple SDK doesn't release any channel resources on disabled state. This will only be called if targets are enabled.

jaelim-ms: Error logs were added.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
jaelim-ms: Low cost since most of cases are iterating all channels but definitely we can improve it.
For now, we have only one case to get channelUnit, called only once, and it is being kept in the memory so I'll leave it.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -27,6 +28,7 @@ - (instancetype)init {
     _discardLogs = NO;
     _delegates = [NSHashTable weakObjectsHashTable];
     _pausedIdentifyingObjects = [NSHashTable weakObjectsHashTable];
+    _pausedTargetKeys = [NSHashTable new];
```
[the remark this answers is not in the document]
jaelim-ms: @achocron `NSHashTable` is thread-safe but functions are not. We need to revisit this one.

PR #1150, opened 2018-10-02
Catch additional exceptions on macOS

MatkovIvan:
This approach is used [in HockeyApp](https://support.hockeyapp.net/kb/client-integration-ios-mac-os-x-tvos/hockeyapp-for-macos#3-3-3-catch-additional-exceptions), but swizzling is the better way to do it.


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -277,6 +279,22 @@ - (instancetype)init {
...
+    /*
+     * Exceptions on the main thread of a Cocoa application do not typically rise to the level of
+     * the uncaught exception handler because the global application object catches all such exceptions.
+     * See: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Concepts/UncaughtExceptions.html
+     */
+    if (![MSAppCenter isDebuggerAttached]) {
+      SEL selector = @selector(reportException:);
+      Method method = class_getInstanceMethod([NSApplication class], selector);
+      IMP implementation = class_getMethodImplementation([MSCrashes class], selector);
+      if (method && implementation) {
```
ElektrojungeMS: I think this needs broader discussion. Personally, I think this is a great feature - this was planned in HockeyApp, see https://github.com/bitstadium/HockeySDK-Mac/blob/6a6e31c3cafd4991672bcbaf1a48e135a878ae5b/Classes/CrashReporting/BITCrashExceptionApplication.h - but we want to be cautious about the swizzling. I'd like to see this as an opt-in API or at least have a discussion around it. Also: have we looked into side-effects of this thoroughly?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -277,6 +279,22 @@ - (instancetype)init {
...
      */
     [[NSUserDefaults standardUserDefaults] registerDefaults:@{ @"NSApplicationCrashOnExceptions" : @YES }];
+
+    /*
+     * Exceptions on the main thread of a Cocoa application do not typically rise to the level of
+     * the uncaught exception handler because the global application object catches all such exceptions.
+     * See: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Concepts/UncaughtExceptions.html
+     */
+    if (![MSAppCenter isDebuggerAttached]) {
+      SEL selector = @selector(reportException:);
```
clpolet: Some people just don't like swizzling so we usually provide a way to opt-out.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -277,6 +279,22 @@ - (instancetype)init {
...
+      Method method = class_getInstanceMethod([NSApplication class], selector);
+      IMP implementation = class_getMethodImplementation([MSCrashes class], selector);
```
clpolet: What do we do with the original implementation? We should also execute the original implementation, it feels like people are expecting this method [to write the exception to the console](https://developer.apple.com/documentation/appkit/nsapplication/1428396-reportexception?language=objc).

ElektrojungeMS: Good point!

ElektrojungeMS:
Overall, this is just awesome sauce!

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: Calling the original one cause the crash with SIGILL (due to `NSApplicationCrashOnExceptions` setting I guess), and calling PLCR handler also aborts the execution, so calling the original one is problematic. I don't want to change this setting to break anything

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: HockeyApp has the way without swizzling, but "one line to integrate the SDK" is better approach I guess.

guperrot: We already have a Info.plist to disable swizzling, the default is still swizzling.
We should make sure we read that existing flag and document.

MatkovIvan: that it isn't swizzling App**Delegate** but `NSApplication`, so it doesn't look like `AppCenterAppDelegateForwarderEnabled` is good for the case

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
ElektrojungeMS: Thx for the clarification Ivan!

One suggestion: Can we rename the entry to the `info.plist` something that requires less technical knowledge. I suggest `kMSIsAppCenterCrashForwardingMacOsExceptionsAsCrashesKey`.
I'm not blocking on this one as it's the naming of a variable.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
ElektrojungeMS: As discussed, we want an opt-out for the swizzling.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * **[Fix]** Add missing network request error logging.
 
+### AppCenterCrashes
+
+* **[Improvement]** Catch additional exceptions on macOS.
```
ElektrojungeMS: Can we be more vocal here and describe previous and new behavior?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
clpolet: If I understood correctly if we call the original method before PLCR handler it will crash depending on the setting but if we call it afterwards then it won't do anything since PLCR will stop the execution. OK we are stuck that means we'll have to document the limitation of this approach. The crash will still be logged in the console by the SDK correct log level activated anyways so it's not really an issue.

clpolet: Basically if the customer shouldn't call `reportException` themselves if this is used.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -262,10 +273,31 @@ - (instancetype)init {
...
      */
-    [[NSUserDefaults standardUserDefaults] registerDefaults:@{ @"NSApplicationCrashOnExceptions" : @YES }];
+    NSNumber *crashForwarderValue = [[NSBundle mainBundle] objectForInfoDictionaryKey:kMSIsAppCenterCrashForwarderEnabledKey];
+    BOOL crashForwarderEnabled = crashForwarderValue ? [crashForwarderValue boolValue] : YES;
+    if (crashForwarderEnabled) {
+      SEL selector = @selector(reportException:);
+      Method method = class_getInstanceMethod([NSApplication class], selector);
+      IMP implementation = class_getMethodImplementation([MSCrashes class], selector);
+      if (method && implementation) {
+        method_setImplementation(method, implementation);
```
clpolet: Is it possible that the system would be calling `reportException` at the time the swizzling happens? That would cause an undefined behavior I guess.

CHANGELOG.md
clpolet: We should probably mention swizzling so that existing customers can be aware of it in case they don't want it.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: @clpolet It isn't changed by this PR. Calling `reportException` crashes the application because the SDK sets that flag. I've already suggested get rid of this behavior and report handled exceptions instead.

MatkovIvan: @ElektrojungeMS This name makes confusion because even if it disabled the SDK changes behavior to crash the application on exception (and it stored for next runs).

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: 1. It used only for reporting exception from the main thread.
2. **BOTH** implementations will report the crash to the AppCenter. But swizzled one allows get exception information instead of `SIGILL`.

clpolet: And the Crashes service init happens on the main thread so both the wiring the swizzling and the reportException can't happen at the same time.

PR #1156, opened 2018-10-04
Adding strong typed option to trackEvent

jwallra:
~Note: Formatting is not correct, that will be fixed soon~


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -33,8 +33,8 @@ - (instancetype)init {
   return self;
 }
 
-- (instancetype)initWithIngestion:(nullable id<MSIngestionProtocol>)ingestion
-                          storage:(id<MSStorage>)storage
+- (instancetype)initWithIngestion:(nullable id <MSIngestionProtocol>)ingestion
```
jaelim-ms: I only see format changes in this file. Can you revert?

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.h
```diff
@@ -1,4 +1,5 @@
 #import "MSLogWithNameAndProperties.h"
+#import "MSEventProperties.h"
```
jaelim-ms: Reorder import

AppCenterAnalytics/AppCenterAnalytics/Internals/MSAnalyticsPrivate.h
```diff
@@ -44,6 +44,17 @@ NS_ASSUME_NONNULL_BEGIN
            withProperties:(nullable NSDictionary<NSString *, NSString *> *)properties
     forTransmissionTarget:(nullable MSAnalyticsTransmissionTarget *)transmissionTarget;
 
+/**
+ * Track an event with typed properties.
+ *
+ * @param eventName  event name.
+ * @param properties typed properties.
+ * @param transmissionTarget Transmission target to associate with the event.
```
jaelim-ms: Looks it is copied from above method, **T**ransmission -> **t**ransmission

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+#import "MSTypedProperty.h"
```
jaelim-ms: Line break between `#import <...>` and `#import "..."`

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
```diff
@@ -0,0 +1,49 @@
+#import "MSBooleanTypedProperty.h"
+#import "MSConstants+Internal.h"
+
+@implementation MSBooleanTypedProperty
+
+- (instancetype)init {
+  if ((self = [super init])) {
+    self.type = @"boolean";
```
jaelim-ms: Constant. This is a pattern that we are using in data models.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
```diff
@@ -0,0 +1,49 @@
+#import "MSBooleanTypedProperty.h"
+#import "MSConstants+Internal.h"
+
+@implementation MSBooleanTypedProperty
```
jaelim-ms: Need to reorder methods in the class.
Look at one of log models.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Serialize this object to a dictionary.
+ *
+ * @return A dictionary representing this object.
+ */
+- (NSMutableDictionary *)serializeToDictionary {
+  NSMutableDictionary *dict = [super serializeToDictionary];
+  dict[kMSTypedPropertyValue] = @(self.value);
```
jaelim-ms: Does it work? We had an issue when we serialize BOOL to dictionary, we have used below pattern.
`dict[kMSTypedPropertyValue] = self.value ? @YES : @NO`
Maybe **@(** ... **)** does the thing?

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
```diff
@@ -0,0 +1,49 @@
...
+  return dict;
+}
+
+- (instancetype)createValidCopyForAppCenter {
```
jaelim-ms: Can you decouple business logic from data model?

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
jaelim-ms: For all typed property classes.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
jaelim-ms: For all typed property classes.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
jaelim-ms: For all typed property classes.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.h
jaelim-ms: For all typed property classes.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSStringTypedProperty.m
```diff
@@ -0,0 +1,55 @@
+#import "MSStringTypedProperty.h"
```
jaelim-ms: Please reorder. I know you can get confused but we haven't had a rule to specifying class header file at the top, there was a confusion. Just sort it alphabetically.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.h
```diff
@@ -16,14 +18,33 @@ NS_ASSUME_NONNULL_BEGIN
...
     withProperties:(nullable NSDictionary<NSString *, NSString *> *)properties;
 
+/**
+ * Track an event with typed properties.
+ *
+ * @param eventName  event name.
+ * @param properties an MSEventProperties object.
+ *
+ * @discussion For events going to App Center, the following validation rules are applied:
+ *
```
jaelim-ms: Extra spaces are not needed.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.h
```diff
@@ -16,14 +18,33 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (void)trackEvent:(NSString *)eventName
     withProperties:(nullable NSDictionary<NSString *, NSString *> *)properties;
 
+/**
+ * Track an event with typed properties.
+ *
+ * @param eventName  event name.
+ * @param properties an MSEventProperties object.
+ *
+ * @discussion For events going to App Center, the following validation rules are applied:
```
jaelim-ms: Do you want to mention validation rules for OC?

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
```diff
@@ -0,0 +1,153 @@
...
+  if ((self = [self init])) {
+    [coder decodeObject];
+  }
+  return self;
+}
+
+#pragma mark - Public methods
+
+- (instancetype)setString:(NSString *)value forKey:(NSString *)key {
+  if (!key) {
```
jaelim-ms: You can extract this to a method. Then you don't need to define a constant for warning message.

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
```diff
@@ -0,0 +1,153 @@
...
+    MSLogWarning([MSAnalytics logTag], kMSNullPropertyKeyMessage);
```
jaelim-ms: Need to align log level with Android implementation. Android uses `error` but looks it is a `warn`. Same for all validation logs in the file.

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
```diff
@@ -0,0 +1,153 @@
...
+  return self;
+}
+
+- (instancetype)setDouble:(double)value forKey:(NSString *)key {
+  if (!key) {
+    MSLogWarning([MSAnalytics logTag], kMSNullPropertyKeyMessage);
+    return self;
+  }
+  if (value == (double)INFINITY || value == (double)NAN) {
+    MSLogError([MSAnalytics logTag], @"Double value for property '%@' must be finite (cannot be INFINITY or NAN).", key);
```
jaelim-ms: Not consistent log level.

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
```diff
@@ -0,0 +1,153 @@
...
+}
+
+#pragma mark - Public methods
+
+- (instancetype)setString:(NSString *)value forKey:(NSString *)key {
+  if (!key) {
+    MSLogWarning([MSAnalytics logTag], kMSNullPropertyKeyMessage);
+    return self;
+  }
+  if (!value) {
```
jaelim-ms: You can also extract this `null` check for values. This is a common validation rule for all types (except primitive data types).

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
```diff
@@ -0,0 +1,153 @@
...
+
+- (NSMutableArray *)serializeToArray {
+  NSMutableArray *propertiesArray = [NSMutableArray new];
+  for (MSTypedProperty *typedProperty in self.properties) {
+    [propertiesArray addObject:[typedProperty serializeToDictionary]];
+  }
+  return propertiesArray;
+}
+
+- (instancetype)createValidCopyForAppCenter {
```
jaelim-ms: Same comment here. Should be moved from data model.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -5,12 +5,18 @@
...
+#import "MSDateTimeTypedProperty.h"
+#import "MSDoubleTypedProperty.h"
 #import "MSEventLog.h"
+#import "MSLongTypedProperty.h"
 #import "MSMockUserDefaults.h"
 #import "MSPageLog.h"
 #import "MSSessionTrackerPrivate.h"
+#import "MSStringTypedProperty.h"
 #import "MSTestFrameworks.h"
+#import "MSEventPropertiesInternal.h"
```
jaelim-ms: Reorder.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
jaelim-ms: `MSAnalytics+Validation` would be the right place.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSDateTimeTypedProperty.m
```diff
@@ -0,0 +1,52 @@
+#import "MSDateTimeTypedProperty.h"
+#import "MSConstants+Internal.h"
+#import "MSUtility+Date.h"
+
+@implementation MSDateTimeTypedProperty
+
+- (instancetype)init {
+  if ((self = [super init])) {
+    self.type = @"dateTime";
```
ElektrojungeMS: Please use a constant for this.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSDoubleTypedProperty.m
```diff
@@ -0,0 +1,49 @@
...
+- (void)encodeWithCoder:(NSCoder *)coder {
+  [super encodeWithCoder:coder];
+  [coder encodeDouble:self.value forKey:kMSTypedPropertyValue];
+}
+
+/**
+ * Serialize this object to a dictionary.
+ *
+ * @return A dictionary representing this object.
+ */
```
ElektrojungeMS: We're usually not adding API docs to .m files. I don't think this is necessary.

ElektrojungeMS: This applies to other model objects, too.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.h
ElektrojungeMS: I think validation rules in the API doc would be awesome!

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -11,6 +11,10 @@
 #import "MSPageLog.h"
 #import "MSServiceAbstractProtected.h"
 #import "MSUtility+StringFormatting.h"
+#import "MSEventProperties.h"
+#import "MSEventPropertiesInternal.h"
+#import "MSTypedProperty.h"
+#import "MSStringTypedProperty.h"
```
ElektrojungeMS: Please reorder imports here.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
clpolet: I agree, it's better if the validation happens in only one place.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSTypedProperty.m
```diff
@@ -0,0 +1,50 @@
...
+ */
+- (NSMutableDictionary *)serializeToDictionary {
+    NSMutableDictionary * dict = [NSMutableDictionary new];
+    dict[kMSTypedPropertyType] = self.type;
+    dict[kMSTypedPropertyName] = self.name;
+    return dict;
+}
+
+- (instancetype)createValidCopyForAppCenter {
+    if ([self.name length] > 125) {
```
clpolet: The property key validation can be shared between string only `properties` and `typedProperties` once moved to `Analytics+Validation`

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
achocron: Yep that makes it an `NSNumber` (so same as `@YES`)

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
achocron: We used Android's log levels for reference and it looked like they were warnings for these scenarios. Double checking

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
achocron: Again, here actually android used error, we thought. Can double check.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSBooleanTypedProperty.m
achocron: Sure, but since we only will have an `MSTypedProperty` we will need to switch on type and then cast to whatever it is to validate, which is quite unfortunate. (That's why we decided to move it into the data models themselves.)

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.h
achocron: They are there because when you look at the generated description, you only get line breaks this way. It's very confusing when it's all on one line (you can test by removing the extra lines and looking at the quick help generated).

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.h
achocron: We thought that we would leave them out for this PR, since it is unrelated to one collector. (Whoever does the OC work will probably benefit by adding the rules themselves, as it's a useful sanity check when you write them in)

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/Properties/MSStringTypedProperty.m
achocron: I think we need to change that then, because with `#import`s, that can actually break certain files. @ElektrojungeMS

achocron: (the order matters for imports in objective c, it's not like java or C#)

AppCenterAnalytics/AppCenterAnalytics/MSEventProperties.m
achocron: Double checked - they were indeed errors. Fixed.

AppCenter/AppCenter/Internals/Model/MSCustomPropertiesLog.m
```diff
@@ -1,16 +1,10 @@
+#import "MSConstants+Internal.h"
 #import "MSCustomPropertiesLog.h"
 #import "MSUtility+Date.h"
 
 static NSString *const kMSCustomProperties = @"customProperties";
+
 static NSString *const kMSProperties = @"properties";
-static NSString *const kMSPropertyType = @"type";
```
achocron: Moved these to `MSConstants+Internal.h` so they could be reused.

achocron:
Refactored the whole thing. Now it's simpler. There are no longer any derived types from `MSTypedProperties`. Also moved validation logic to `MSAnalytics+Validation.h`. Other small fixes too.

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+#import "MSConstants+Internal.h"
+#import "MSTestFrameworks.h"
+
+@interface MSBooleanTypedPropertyTests : XCTestCase
+
+@end
+
+@implementation MSBooleanTypedPropertyTests
+
+- (void)testNSCodingSerializationAndDeserializationWorks {
```
jwallra: It's redundant to put "works" in test names, we can remove these

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+
+  // If
+  MSBooleanTypedProperty *sut = [MSBooleanTypedProperty new];
```
jwallra: Why use "sut" here instead of just "property"?  It doesn't make any sense in this context to me

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+  XCTAssertEqualObjects(dictionary[@"type"], sut.type);
+  XCTAssertEqualObjects(dictionary[@"name"], sut.name);
+  XCTAssertEqual([dictionary[@"value"] boolValue], sut.value);
+}
+
+- (void)testCreateValidCopyForAppCenterWorksWhenNameIsTooLong {
+
+  // If
+  MSBooleanTypedProperty *sut = [MSBooleanTypedProperty new];
+  sut.name = [@"" stringByPaddingToLength:kMSMaxPropertyKeyLength + 2 withString:@"hi" startingAtIndex:0];
```
jwallra: Do we not already have a string generation utility?  That would be useful and easier to read.

AppCenterAnalytics/AppCenterAnalyticsTests/MSDoubleTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+
+@implementation MSDoubleTypedPropertyTests
+
+- (void)testNSCodingSerializationAndDeserializationWorks {
+
+  // If
+  MSDoubleTypedProperty *sut = [MSDoubleTypedProperty new];
+  sut.type = @"type";
+  sut.name = @"name";
+  sut.value = 12.23432;
```
jwallra: We should also test with other values using different declaration methods, like `1.23e1`

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+  sut.value = YES;
+
+  // When
+  MSBooleanTypedProperty *validCopy = [sut createValidCopyForAppCenter];
+
+  // Then
+  XCTAssertEqualObjects(validCopy.type, sut.type);
+  XCTAssertEqualObjects(validCopy.name, [sut.name substringToIndex:kMSMaxPropertyKeyLength]);
+  XCTAssertEqual(validCopy.value, sut.value);
```
jwallra: The `sut` is being used interchangeably as the actual and expected value in this file (see this and line 69).  I think the pattern is `XCTAssert(actual, expected)` so we should follow that pattern even for the `XCTAssertEqual` and `XCTAssertObjectEqual` methods

AppCenterAnalytics/AppCenterAnalyticsTests/MSDoubleTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+  XCTAssertEqualObjects(actual.type, sut.type);
+  XCTAssertEqual(actual.value, sut.value);
+}
+
+- (void)testSerializeToDictionaryWorks {
+
+  // If
+  MSDoubleTypedProperty *sut = [MSDoubleTypedProperty new];
+  sut.name =  @"propertyName";
+  sut.value = 0.123;
```
jwallra: Is there any reason we're using different values for different test cases?  It's a bad pattern to use.  If we want to use multiple values we should put them in a list and iterate over them for each test.

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
jwallra: I see that this is actually a pattern we use in iOS but I greatly dislike it...

AppCenterAnalytics/AppCenterAnalyticsTests/MSEventPropertiesTests.m
```diff
@@ -109,4 +109,22 @@ - (void)testAppCenterCopyHas20PropertiesWhenSelfHasMoreThan20 {
   XCTAssertEqual([appCenterCopy.properties count], 20);
 }
 
+- (void)testEventPropertiesHaveValidPropertiesForAppCenterWhenSutIsCopiedForAppCenter {
```
jwallra: I dislike using `sut` in the title.  We can use it for the variable name since it's a convention but we should just use something like `testEventPropertiesHaveValidPropertiesForAppCenterWhenCopiedFromAppCenter`

AppCenterAnalytics/AppCenterAnalyticsTests/MSLongTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
+#import "MSLongTypedProperty.h"
+#import "MSConstants+Internal.h"
+#import "MSTestFrameworks.h"
+
+@interface MSBooleanTypedPropertyTests : XCTestCase
```
jwallra: Wrong interface

AppCenterAnalytics/AppCenterAnalyticsTests/MSLongTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+
+@end
+
+@implementation MSBooleanTypedPropertyTests
```
jwallra: Wrong implementation

AppCenterAnalytics/AppCenterAnalyticsTests/MSLongTypedPropertyTests.m
```diff
@@ -0,0 +1,72 @@
...
+
+- (void)testNSCodingSerializationAndDeserializationWorks {
+
+  // If
+  MSLongTypedProperty *sut = [MSLongTypedProperty new];
+  sut.type = @"type";
+  sut.name = @"name";
+  sut.value = 12;
```
jwallra: Should use a value that wouldn't fit in a regular integer.  See if there's a `#define` for MAX_INT64` or similar, or use a value 2^32<x<2^64

jwallra: Also should move this into a constant since it's the same for each test.

AppCenterAnalytics/AppCenterAnalyticsTests/MSBooleanTypedPropertyTests.m
ElektrojungeMS: True, this is a candidate for `MSUtility+StringFormatting`.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSTypedProperty.h
```diff
@@ -0,0 +1,53 @@
...
+*/
+@property(nonatomic, copy) NSString *name;
+
+/**
+* Property value.
+*/
+@property(nonatomic, copy) NSObject *value;
+
+/**
+ * Creates an instance with the `type` property set appropriately for string types.
```
ElektrojungeMS: missing empty line before `@return`.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSTypedProperty.h
```diff
@@ -0,0 +1,53 @@
+#import <Foundation/Foundation.h>
+#import "MSSerializableObject.h"
```
jaelim-ms: Empty line between imports.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSTypedProperty.h
```diff
@@ -0,0 +1,53 @@
...
+
+
```
jaelim-ms: Remove unnecessary empty line.

AppCenterAnalytics/AppCenterAnalytics/MSEventPropertiesInternal.h
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * String and date properties.
+ */
+@property (nonatomic) NSMutableDictionary<NSString *, MSTypedProperty *> *properties;
+
+/**
+ * Creates an instance of EventProperties with a string-string properties dictionary.
+ *
+ * @param properties A dictionary of properties.
+ * @return An instance of EventProperties.
```
ElektrojungeMS: missing `*` line before `@return`

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSTypedProperty.h
```diff
@@ -0,0 +1,53 @@
...
+/**
+* Property name.
+*/
+@property(nonatomic, copy) NSString *name;
+
+/**
+* Property value.
+*/
+@property(nonatomic, copy) NSObject *value;
```
jaelim-ms: Do you have any reasons for the recent change?
This would be dangerous since type and object type could mismatch.
If you don't have any strong reason to do this one, I would like to suggest you to make the class back to abstract and having separate child classes for different types.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -363,23 +413,53 @@ - (void)testTrackEventWithInvalidName {
...
+  [MSAppCenter configureWithAppSecret:kMSTestAppSecret];
+  id channelUnitMock = OCMProtocolMock(@protocol(MSChannelUnitProtocol));
+  id channelGroupMock = OCMProtocolMock(@protocol(MSChannelGroupProtocol));
+  OCMStub([channelGroupMock addChannelUnitWithConfiguration:OCMOCK_ANY]).andReturn(channelUnitMock);
+  [[MSAnalytics sharedInstance] startWithChannelGroup:channelGroupMock
+                                            appSecret:kMSTestAppSecret
+                              transmissionTargetToken:nil
+                                      fromApplication:YES];
+
+  // FIXME: logManager holds session tracker somehow and it causes other test failures. Stop it for hack.
```
ElektrojungeMS: Just a side note:
a) we're not using `logManager` as a term anymore
b) we need to look into this again, it shouldn't be like that.

achocron: I can remove that (it was just copied from another test)

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSTypedProperty.h
achocron: How can a mismatch occur? I just found that the code was far simpler this way, didn't see the need to introduce so many types when one will suffice. You can also look at how we implemented `MSCustomProperties` - there isn't a separate type for each type

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -311,7 +314,32 @@ - (void)testTrackEventWhenAnalyticsDisabled {
   OCMVerifyAll(channelUnitMock);
 }
 
-- (void)testTrackEventWhenTransmissionTargetDisabled {
+- (void)testTrackEventWithTypedPropertiesNilWhenAnalyticsDisabled {
```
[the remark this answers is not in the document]
ElektrojungeMS: Let's track this as a separate bug?

PR #1162, opened 2018-10-11
Fix list of binary images for arm64e devices

ElektrojungeMS:
* Normalizes memory addresses for arm64(e) CPUS to make sure we can detect relevant image binaries for a crash report. This does not actually submit normalized binary images, this still needs to be done server-side. The problem was that arm64e CPUs would not contain all relevant images because the change to arm64e breaks existing memory address comparison that is used to determine relevant binary images.
* Refactors one function into a proper static method to allow it to be used in unit tests.


ElektrojungeMS:
FYI I tested on armv7s, arm64 and arm64e devices.

PR #1172, opened 2018-10-16
Add more unit tests for logger

jaelim-ms:


AppCenter/AppCenterTests/MSLoggerTests.m
```diff
@@ -48,4 +48,17 @@ - (void)testSetCurrentLoglevelWorks {
...
 
+- (void)testLoglevelNoneDoesNotLogMessages {
+
+  // If
+  MSLogMessageProvider messageProvider = ^() {
+    XCTFail(@"Log shouldn't be printed.");
+    return @"";
+  };
+
+  // When
```
ElektrojungeMS: No `// Then`?

jaelim-ms: The line 55 is kind of `// Then` but basically it is testing that the operation doesn't do anything.

PR #1174, opened 2018-10-17
Update max storage size field in case it was changed by sdk automatically

bmourat:

When setting max storage size, SDK could increase value that was provided by user due to page size used by storage. In this case test apps were displaying incorrect value for storage size, this PR fixes it.

### DIFF
--- a/Sasquatch/SasquatchSwift/AppDelegate.swift
+++ b/Sasquatch/SasquatchSwift/AppDelegate.swift
@@ -38,18 +38,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
     let storageMaxSize = UserDefaults.standard.object(forKey: kMSStorageMaxSizeKey) as? Int
     if storageMaxSize != nil {
       MSAppCenter.setMaxStorageSize(storageMaxSize!, completionHandler: { success in
-        if !success {
-          DispatchQueue.main.async {
-
+        DispatchQueue.main.async {
+          if !success {
             // Remove invalid value.
             UserDefaults.standard.removeObject(forKey: kMSStorageMaxSizeKey)
 
             // Show alert.
             let alertController = UIAlertController(title: "Warning!",
-                    message: "The maximum size of the internal storage could not be set.",
-                    preferredStyle: .alert)
+                                                    message: "The maximum size of the internal storage could not be set.",
+                                                    preferredStyle: .alert)
             alertController.addAction(UIAlertAction(title: "OK", style: .default))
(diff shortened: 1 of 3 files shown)


Sasquatch/SasquatchSwift/AppDelegate.swift
```diff
@@ -38,18 +38,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
     let storageMaxSize = UserDefaults.standard.object(forKey: kMSStorageMaxSizeKey) as? Int
     if storageMaxSize != nil {
       MSAppCenter.setMaxStorageSize(storageMaxSize!, completionHandler: { success in
-        if !success {
-          DispatchQueue.main.async {
-
+        DispatchQueue.main.async {
+          if !success {
```
jaelim-ms: Remove `!` and swap statements for true and false conditions.
```
if success {
            let realSize = Int64((ceil(Double(storageMaxSize!) / Double(kMSStoragePageSize)))) * Int64(kMSStoragePageSize)
            UserDefaults.standard.set(realSize, forKey: kMSStorageMaxSizeKey)
} else {
            // Remove invalid value.
            UserDefaults.standard.removeObject(forKey: kMSStorageMaxSizeKey)
             // Show alert.
            let alertController = UIAlertController(title: "Warning!",
                                                    message: "The maximum size of the internal storage could not be set.",
                                                    preferredStyle: .alert)
            alertController.addAction(UIAlertAction(title: "OK", style: .default))
            self.window?.rootViewController?.present(alertController, animated: true)
}

File: Sasquatch/SasquatchSwift/AppDelegate.swift [hunk 2]
```diff
@@ -38,18 +38,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MSCrashesDelegate, MSDist
...
             // Show alert.
             let alertController = UIAlertController(title: "Warning!",
-                    message: "The maximum size of the internal storage could not be set.",
-                    preferredStyle: .alert)
+                                                    message: "The maximum size of the internal storage could not be set.",
+                                                    preferredStyle: .alert)
             alertController.addAction(UIAlertAction(title: "OK", style: .default))
             self.window?.rootViewController?.present(alertController, animated: true)
+          } else {
+            let realSize = Int64((ceil(Double(storageMaxSize!) / Double(kMSStoragePageSize)))) * Int64(kMSStoragePageSize)
```
jwallra: Is there an extra set of parentheses here, around `ceil(...)`?

PR #1180, opened 2018-10-18
Add script to format files and update contribution guidelines

achocron:

## Description

Added a script that uses clang-format and our custom format file to format all files that have changes against origin/develop. Updated readme. Make sure to run `brew install clang-format` before running the script.

### DIFF
--- a/clang-format-changed-files.sh
+++ b/clang-format-changed-files.sh
@@ -2,9 +2,10 @@
 
 # This script runs clang_format on all .h, .m, and .mm files that have changes from develop branch.
 # Before running this, ensure that your local develop branch is up to date.
-git fetch
+upstream=$(git remote -v | grep "Microsoft" | awk '{print $1}' | uniq)
+git fetch $upstream
 i=0
-for modified_file in $(git diff origin/develop --name-only *.h *.m *.mm)
+for modified_file in $(git diff $upstream/develop --name-only *.h *.m *.mm)
 do 
   clang-format -i -style=file $modified_file
   exit_code=$?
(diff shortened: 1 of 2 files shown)


clang-format-changed-files.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
guperrot: Can't tell on github UI but please make that the `chmod u+x` for this file is part of the commit.

dhei: Yes, we did that.

clang-format-changed-files.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# This script runs clang_format on all .h, .m, and .mm files that have changes from develop branch.
+# Before running this, ensure that your local develop branch is up to date.
+git fetch
+i=0
+for modified_file in $(git diff origin/develop --name-only *.h *.m *.mm)
```
guperrot: On forks origin/develop will likely be stale, while upstream/develop might be the right one to serve as a reference (and could be another name).

This makes this script highly unreliable. We might need to adapt logic to find the right remote to fetch and compare against.

guperrot: Would do: UPSTREAM=`git remote -v | grep ":Microsoft" | awk '{print $1}' | uniq`
git fetch $UPSTREAM
git diff "$UPSTREAM/develop"

clang-format-changed-files.sh
```diff
@@ -2,9 +2,10 @@
 
 # This script runs clang_format on all .h, .m, and .mm files that have changes from develop branch.
 # Before running this, ensure that your local develop branch is up to date.
-git fetch
+upstream=$(git remote -v | grep "Microsoft" | awk '{print $1}' | uniq)
```
guperrot: nit: ":Microsoft/" to avoid matching users that would use Microsoft somewhere in their account or org.

PR #1183, opened 2018-10-19
Logs not associated to the right target keys

achocron:

*~[ ] Has `CHANGELOG.md` been updated?~
* [x] Did you add unit tests?

## Description

Fixes an issue where an event created before setting an authentication provider would use the subsequently added provider instead of what was set prior to the log being sent.

Found that is not harmful to remove the `dispatch_async`s from surrounding the `enqueueItem` in the `MSOneCollectorChannelDelegate` because `enqueueItem` is not a blocking method (in `trackEvent` the SDK is already calling `enqueueItem` on the calling thread).

## Related PRs or issues

https://github.com/Microsoft/AppCenter-SDK-Android/pull/837

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
@@ -85,9 +85,7 @@ NSString *const kMSLogNameRegex = @"^[a-zA-Z0-9]((\\.(?!(\\.|$)))|[_a-zA-Z0-9]){
   if ([(NSObject *)log isKindOfClass:[MSCommonSchemaLog class]] && ![self isOneCollectorGroup:groupId]) {
     oneCollectorChannelUnit = self.oneCollectorChannels[groupId];
     if (oneCollectorChannelUnit) {
-      dispatch_async(oneCollectorChannelUnit.logsDispatchQueue, ^{
         [oneCollectorChannelUnit enqueueItem:log];
-      });
     }
     return;
   }
@@ -101,9 +99,7 @@ NSString *const kMSLogNameRegex = @"^[a-zA-Z0-9]((\\.(?!(\\.|$)))|[_a-zA-Z0-9]){
   id<MSLogConversion> logConversion = (id<MSLogConversion>)log;
   NSArray<MSCommonSchemaLog *> *commonSchemaLogs = [logConversion toCommonSchemaLogs];
   for (MSCommonSchemaLog *commonSchemaLog in commonSchemaLogs) {
-    dispatch_async(oneCollectorChannelUnit.logsDispatchQueue, ^{
       [oneCollectorChannelUnit enqueueItem:commonSchemaLog];
-    });
   }
 }
 
--- a/AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
@@ -1,11 +1,11 @@
 #import "MSAbstractLogInternal.h"
 #import "MSAppCenterInternal.h"
-#import "MSChannelUnitConfiguration.h"
-#import "MSChannelUnitProtocol.h"
-#import "MSConstants+Internal.h"
 #import "MSCSData.h"
 #import "MSCSEpochAndSeq.h"
 #import "MSCSExtensions.h"
+#import "MSChannelUnitConfiguration.h"
+#import "MSChannelUnitProtocol.h"
+#import "MSConstants+Internal.h"
 #import "MSOneCollectorChannelDelegatePrivate.h"
 #import "MSOneCollectorIngestion.h"
 #import "MSSDKExtension.h"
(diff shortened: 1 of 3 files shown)


AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
```diff
@@ -101,9 +99,7 @@ - (void)channel:(id<MSChannelProtocol>)channel didPrepareLog:(id<MSLog>)log with
   id<MSLogConversion> logConversion = (id<MSLogConversion>)log;
   NSArray<MSCommonSchemaLog *> *commonSchemaLogs = [logConversion toCommonSchemaLogs];
   for (MSCommonSchemaLog *commonSchemaLog in commonSchemaLogs) {
-    dispatch_async(oneCollectorChannelUnit.logsDispatchQueue, ^{
-      [oneCollectorChannelUnit enqueueItem:commonSchemaLog];
-    });
+    [oneCollectorChannelUnit enqueueItem:commonSchemaLog];
```
guperrot: does this do disk access? If yes we shouldn't execute in UI thread.

guperrot:
We should have a common branch in 1ds docs for changelog related to 1ds with don't merge.

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
achocron: It shouldn't. We execute enqueueItem in the UI thread every time TrackEvent is called.

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -251,6 +251,38 @@ - (void)testDidEnqueueLogToOneCollectorChannelWhenLogHasTargetTokensAndLogIsNotC
...
+  id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
+  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@[ commonSchemaLog ]);
+  OCMStub(mockLog.transmissionTargetTokens).andReturn(transmissionTargetTokens);
+
+  /*
+   * Make sure that the common schema log is enqueued synchronously by putting a task on the log queue that won't return
+   * by the time verify is called.
+   */
+  dispatch_async(oneCollectorChannelUnitMock.logsDispatchQueue, ^{
+    sleep(1);
```
guperrot: Why sleep though?

AppCenter/AppCenter/Internals/Channel/MSOneCollectorChannelDelegate.m
achocron: So if we are, a broader change is needed

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
jwallra: We chose sleep to simulate the behavior when doing something like configuring an authentication provider.  We could do that in this block to mirror the case for which the unintended behavior was observed, but it isn't necessarily limited to that scenario.

It is kind of a weird choice; if you have a suggestion to make it clearer it would be appreciated 😄

guperrot: We could use a semaphore, or equivalent, to guarantee order of operations during the test instead of sleeping.

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -251,6 +251,39 @@ - (void)testDidEnqueueLogToOneCollectorChannelWhenLogHasTargetTokensAndLogIsNotC
...
+  MSCommonSchemaLog *commonSchemaLog = [MSCommonSchemaLog new];
+  id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
+  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@[ commonSchemaLog ]);
+  OCMStub(mockLog.transmissionTargetTokens).andReturn(transmissionTargetTokens);
+  dispatch_semaphore_t sem = dispatch_semaphore_create(0);
+  /*
+   * Make sure that the common schema log is enqueued synchronously by putting a task on the log queue that won't return
+   * by the time verify is called.
+   */
+  dispatch_async(oneCollectorChannelUnitMock.logsDispatchQueue, ^{
```
jaelim-ms: Why do we need this one? The unit test still passes without this one. Either unit test is wrong or semaphore isn't necessary.

achocron: Though the test passes without this, it doesn't test the right thing. This guarantees that the code is not running on the dispatch queue. If it did run on the dispatch queue, it could fail.

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -251,6 +251,39 @@ - (void)testDidEnqueueLogToOneCollectorChannelWhenLogHasTargetTokensAndLogIsNotC
...
+  OCMStub(oneCollectorChannelUnitMock.logsDispatchQueue).andReturn(self.logsDispatchQueue);
+  OCMStub([channelGroupMock addChannelUnitWithConfiguration:OCMOCK_ANY withIngestion:OCMOCK_ANY]).andReturn(oneCollectorChannelUnitMock);
+  NSMutableSet *transmissionTargetTokens = [NSMutableSet new];
+  [transmissionTargetTokens addObject:@"fake-transmission-target-token"];
+  MSCommonSchemaLog *commonSchemaLog = [MSCommonSchemaLog new];
+  id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
+  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@[ commonSchemaLog ]);
+  OCMStub(mockLog.transmissionTargetTokens).andReturn(transmissionTargetTokens);
+  dispatch_semaphore_t sem = dispatch_semaphore_create(0);
+  /*
```
clpolet: Need a new line here.

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
clpolet: BTW in case of failure will the test actually fail or will it hang there forever waiting on the semaphore?

achocron: It actually failed when we tested

PR #1184, opened 2018-10-22
Use standard SQL syntax

MatkovIvan:
* [x] ~Did you add unit tests?~

## Description

`DELETE FROM \"%@\" ORDER BY \"%@\" ASC LIMIT %ld` is not a standard SQL

See [SQLite Query Language: DELETE](https://www.sqlite.org/lang_delete.html):
> If SQLite is compiled with the [SQLITE_ENABLE_UPDATE_DELETE_LIMIT](https://www.sqlite.org/compile.html#enable_update_delete_limit) compile-time option, then the syntax of the DELETE statement is extended by the addition of optional ORDER BY and LIMIT clauses


dhei:
@MatkovIvan, 
Per [SQLITE_ENABLE_UPDATE_DELETE_LIMIT](https://www.sqlite.org/compile.html#enable_update_delete_limit) documentation:

> If this option is defined, then it must also be defined when using the Lemon parser generator tool to generate a parse.c file. Because of this, this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website. 

Is that the SQLite has to be built from source the reason why we don't want to enable this compile-time option, hence should not use this non-standard DELETE LIMIT syntax?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * **[Fix]** Fix an issue where concurrent modification of custom properties was not thread safe.
 * **[Fix]** Fix validating and discarding Not a Number (NaN) and infinite double values for custom properties.
+* **[Fix]** Do not use non-standard SQL syntax to avoid affect users with custom SQLite library.
```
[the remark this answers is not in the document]
jwallra: Also `affect` -> `affecting` and `library` - > `libraries`

MatkovIvan:
@dhei Apparently this option is available in the out-of-the-box iOS library.

@jaelim-ms There is a test for this logic. What kind of test do you suggest here?

@achocron @ElektrojungeMS @jwallra Fixed

MatkovIvan:
Well, to make it fail I've used official sqlite without any custom flags (in unit tests only).
Now the test fails with (on original implementation):
`Query "DELETE FROM "logs" ORDER BY "id" ASC LIMIT 1" failed with error: 1 - near "ORDER": syntax error`

Also infinite loop on error issue has been fixed.

MatkovIvan:
@ElektrojungeMS yes

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -737,6 +737,9 @@
 		F803BBF91E8E3989004B1E7A /* MSCustomPropertiesLog.m in Sources */ = {isa = PBXBuildFile; fileRef = F803BBF81E8E3989004B1E7A /* MSCustomPropertiesLog.m */; };
 		F803BC381E8E6927004B1E7A /* MSCustomPropertiesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F803BC371E8E6927004B1E7A /* MSCustomPropertiesTests.m */; };
 		F803BC3A1E8E6963004B1E7A /* MSCustomPropertiesLogTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F803BC391E8E6963004B1E7A /* MSCustomPropertiesLogTests.m */; };
+		F82E4C6D217F159A00EDAB34 /* sqlite3.c in Sources */ = {isa = PBXBuildFile; fileRef = F82E4C6C217F159A00EDAB34 /* sqlite3.c */; settings = {COMPILER_FLAGS = "-w"; }; };
```
jaelim-ms: Is it really required? I could run all testing without this one. Can you double check?

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -780,4 +786,13 @@ - (BOOL)logs:(NSArray<id<MSLog>> *)logs containLogWithSessionId:(NSString *)sess
   return NO;
 }
 
+- (void)validateQuery:(NSString *)query {
```
jaelim-ms: I initially thought this method validates expected operations but the implementation only see the query is not malformed. The method name might lead wrong expectation, please add a doc for the method.

AppCenter/AppCenter.xcodeproj/project.pbxproj
MatkovIvan: `sqlite3` is required since we really use it.
It can be used by link system one (was used before) or any another implementation.
The latest official implementation has been added instead of system one (see [xcconfig](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1184/files#diff-8a9c66c4d1bf5fc63394d7af037ec1bcR4)).

And yes, the tests passed with system implementation, but key thing here is official implementation can handle non-standard sql syntax (topic of the PR). So, this change allows detect the issue with previous implementation and makes the test meaningful (fail before, pass after).

jaelim-ms:
@MatkovIvan @clpolet Sorry for any confusions. What I meant was just normal unit tests that we have and Ivan's answer was enough for my comment. The change seems good to have since it could detect non standard SQL syntax but I wonder if that's that we really want to do in unit tests.
Unit tests are verifying the code that we wrote, not preventing from some issues. This kind of protection should be done during PR reviews + code doc (comments).

jaelim-ms:
Sync'ed with @clpolet offline. The only point that I don't like is to have sqlite3.c in the project but looks this way is better (not the best) than any other solutions. Merging the PR.

jaelim-ms:
Build failed. Close and open to trigger CI build.

PR #1186, opened 2018-10-22
Fix/dont return default transmission target

achocron:

* [x] Has `CHANGELOG.md` been updated? (https://msasg.visualstudio.com/Shared%20Data/_git/1dsdocs/pullrequest/809581?_a=files&path=%2Fios-changelog.md)
* [x] Did you add unit tests?

## Description

* Don't return the default transmission target from the `transmissionTargetForToken` method
* Introduce "tag" to logs so that the property configurator doesn't apply properties to the wrong logs

## Related PRs or issues

https://github.com/Microsoft/AppCenter-SDK-Android/pull/850

### DIFF
--- a/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
+++ b/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
@@ -61,7 +61,8 @@
     return NO;
   }
   MSAbstractLog *log = (MSAbstractLog *)object;
-  return ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
+  return ((!self.tag && !log.tag) || [self.tag isEqual:log.tag]) &&
+         ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
          ((!self.timestamp && !log.timestamp) || [self.timestamp isEqualToDate:log.timestamp]) &&
          ((!self.sid && !log.sid) || [self.sid isEqualToString:log.sid]) &&
          ((!self.distributionGroupId && !log.distributionGroupId) || [self.distributionGroupId isEqualToString:log.distributionGroupId]) &&
--- a/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
+++ b/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
@@ -1,6 +1,6 @@
+#import "MSACModelConstants.h"
 #import "MSAbstractLogInternal.h"
 #import "MSAbstractLogPrivate.h"
-#import "MSACModelConstants.h"
 #import "MSAppExtension.h"
 #import "MSCSExtensions.h"
 #import "MSCSModelConstants.h"
@@ -62,8 +62,7 @@
     return NO;
   }
   MSAbstractLog *log = (MSAbstractLog *)object;
-  return ((!self.tag && !log.tag) || [self.tag isEqual:log.tag]) &&
-         ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
+  return ((!self.tag && !log.tag) || [self.tag isEqual:log.tag]) && ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
          ((!self.timestamp && !log.timestamp) || [self.timestamp isEqualToDate:log.timestamp]) &&
          ((!self.sid && !log.sid) || [self.sid isEqualToString:log.sid]) &&
          ((!self.distributionGroupId && !log.distributionGroupId) || [self.distributionGroupId isEqualToString:log.distributionGroupId]) &&
--- a/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
+++ b/AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
@@ -1,6 +1,6 @@
-#import "MSACModelConstants.h"
 #import "MSAbstractLogInternal.h"
 #import "MSAbstractLogPrivate.h"
+#import "MSACModelConstants.h"
 #import "MSAppExtension.h"
 #import "MSCSExtensions.h"
 #import "MSCSModelConstants.h"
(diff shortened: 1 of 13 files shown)


AppCenter/AppCenter/Internals/Model/MSLog.h
```diff
@@ -30,6 +30,11 @@
  */
 @property(nonatomic) MSDevice *device;
 
+/**
+ * Transient object tag.
```
clpolet: I am a little confused about the tag thing, what is it used for? Is it really necessary?

clpolet: OK we are using it to remember which target is this log tied to. Should probably be more explicit about it.

clpolet: Also there is another issue, in the future an event could be sent to multiple targets, for instance the session log could be sent to all existing targets. This is why targets are stored into a set.

achocron: Tag is a general data field that can be set to anything. For instance, it can be set to a list of transmission targets. Then you could check if it exists. Alternatively, we could make the tag into a list of objects. That probably depends on future discussions.

As far as being more explicit, I think I opted to align with the Android implementation. In this case, I don't think we should be inconsistent. Also, we could use it for other reasons in the future (maybe some kind of PII scenario?) so I like that it is general.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -840,6 +853,7 @@ - (void)testOverridingCommonSchemaPropertiesWithTwoChildrenUnderTheSameParent {
 
   // Set log1.
```
achocron: Can the author of this test provide some context? @jaelim-ms I don't know if you wrote it initially, but I think you may have? It was weird to adapt because I don't know what it was testing exactly. Who initially would have called "trackEvent"? Since events can only originate from one target, it wasn't clear to me who would "own" the properties. (Multiple transmission targets for one log is a future scenario that the PR doesn't support for now.)

AppCenter/AppCenter/Internals/Model/MSLog.h
clpolet: It can be general purpose but it would require an example to me in the comment.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -153,7 +158,12 @@ - (void)updateConfigurationWithAppSecret:(NSString *)appSecret transmissionTarge
 
   // Create the default target if not already created in start.
   if (token && !self.defaultTransmissionTarget) {
-    self.defaultTransmissionTarget = [self transmissionTargetForToken:token];
+
+    /*
+     * Don't use [self transmissionTargetForToken] because that will add the default transmission target to the cache, but it should be
+     * separate.
+     */
+    self.defaultTransmissionTarget = [self createTransmissionTargetForToken:token];
```
jaelim-ms: This isn't covered by unit test. Please add one.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -1,6 +1,6 @@
+#import "MSACModelConstants.h"
```
ElektrojungeMS: This import should not be at the top imho.

AppCenter/AppCenter/Internals/Model/MSLog.h
ElektrojungeMS: I think this is a good idea.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -61,7 +62,7 @@ - (BOOL)isEqual:(id)object {
     return NO;
   }
   MSAbstractLog *log = (MSAbstractLog *)object;
-  return ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
+  return ((!self.tag && !log.tag) || [self.tag isEqual:log.tag]) && ((!self.type && !log.type) || [self.type isEqualToString:log.type]) &&
```
jaelim-ms: I saw Android also updated equals/hashCode methods with tag but I wonder if this is the expected behavior for comparing logs. If logs are same but have different tags, do we really consider the logs as different?

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -376,6 +383,15 @@ - (MSAnalyticsTransmissionTarget *)transmissionTargetForToken:(NSString *)transm
   return transmissionTarget;
 }
 
+- (MSAnalyticsTransmissionTarget *)createTransmissionTargetForToken:(NSString *)transmissionTargetToken {
+  MSAnalyticsTransmissionTarget *target = [[MSAnalyticsTransmissionTarget alloc] initWithTransmissionTargetToken:transmissionTargetToken
+                                                                                                    parentTarget:nil
+                                                                                                    channelGroup:self.channelGroup];
+  MSLogDebug([MSAnalytics logTag], @"Created transmission target with id %@.",
```
jaelim-ms: `id` -> `target key`

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
jwallra: This wasn't covered before the change and I'm not actually sure that it can be hit.  Will require further investigation but is out of scope for now.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
jwallra: I would think yes, but I'd be open to hearing why that shouldn't be the case.  Reasoning?

AppCenter/AppCenter/Internals/Model/MSLog.h
jwallra: I'll add the example we currently use to the comment

PR #1193, opened 2018-10-25
Feature/set flags on commonschemalog

achocron:

* [x] Did you add unit tests?

## Description

Set the "flags" property of common schema logs when passed to the channel.


AppCenter/AppCenter/MSConstants.h
```diff
@@ -63,3 +63,9 @@ typedef NS_ENUM(NSInteger, MSInitializationPriority) {
   MSInitializationPriorityHigh = 750,
   MSInitializationPriorityMax = 999
 };
+
+typedef NS_OPTIONS(NSUInteger, MSFlags) {
```
clpolet: So we decided to go with MSFlags for all app center flags? My only concern is people won't know which flag works with which API without looking at the documentation. But it's your call.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -772,7 +772,7 @@ - (void)processLogBufferAfterCrash {
 
           // Buffered logs are used sending their own channel. It will never contain more than 50 logs.
           MSLogDebug([MSCrashes logTag], @"Re-enqueueing buffered log, type: %@.", item.type);
-          [self.bufferChannelUnit enqueueItem:item];
+          [self.bufferChannelUnit enqueueItem:item flags:NO];
```
clpolet: Shouldn't it be an explicit flag here?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -847,7 +847,7 @@ - (void)sendErrorAttachments:(NSArray<MSErrorAttachmentLog *> *)errorAttachments
       MSLogError([MSCrashes logTag], @"Not all required fields are present in MSErrorAttachmentLog.");
       continue;
     }
-    [self.channelUnit enqueueItem:attachment];
+    [self.channelUnit enqueueItem:attachment flags:NO];
```
clpolet: Shouldn't it be an explicit flag here?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -1126,7 +1128,7 @@ - (void)notifyWithUserConfirmation:(MSUserConfirmation)userConfirmation {
     log.sid = [[MSSessionContext sharedInstance] sessionIdAt:log.timestamp];
 
     // Then, enqueue crash log.
-    [self.channelUnit enqueueItem:log];
+    [self.channelUnit enqueueItem:log flags:YES];
```
clpolet: Shouldn't it be an explicit flag here?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -1161,12 +1163,13 @@ - (void)trackModelException:(MSException *)exception withProperties:(NSDictionar
...
-    log.properties =
-        [MSUtility validateProperties:properties forLogName:[NSString stringWithFormat:@"ErrorLog: %@", log.errorId] type:log.type];
+    log.properties = [MSUtility validateProperties:properties
+                                        forLogName:[NSString stringWithFormat:@"ErrorLog: %@", log.errorId]
+                                              type:log.type];
   }
 
   // Enqueue log.
-  [self.channelUnit enqueueItem:log];
+  [self.channelUnit enqueueItem:log flags:NO];
```
clpolet: Shouldn't it be an explicit flag here?

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -308,7 +308,7 @@ - (void)testProcessCrashesWithErrorAttachments {
                             }]])
       .andReturn(channelUnitMock);
   for (NSUInteger i = 0; i < invalidLogs.count; i++) {
-    OCMReject([channelUnitMock enqueueItem:invalidLogs[i]]);
+    OCMReject([channelUnitMock enqueueItem:invalidLogs[i] flags:NO]);
```
clpolet: Booleans instead of flags in many tests :blush:

achocron: and a few in the actual implementations...fixed - good catch!

AppCenter/AppCenter/Internals/Storage/MSLogDBStoragePrivate.h
```diff
@@ -11,6 +11,7 @@ static NSString *const kMSGroupIdColumnName = @"groupId";
 static NSString *const kMSLogColumnName = @"log";
 static NSString *const kMSTargetTokenColumnName = @"targetToken";
 static NSString *const kMSTargetKeyColumnName = @"targetKey";
+static NSString *const kMSLogPersistencePriorityColumnName = @"logPersistencePriority";
```
guperrot: Why not just priority?

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -313,15 +315,18 @@ - (NSUInteger)countLogs {
...
   if (version < kMSTargetKeyVersion) {
-    NSString *migrationQuery = [NSString stringWithFormat:@"ALTER TABLE \"%@\" ADD COLUMN \"%@\" "
-                                                           "TEXT",
-                                                          kMSLogTableName, kMSTargetKeyColumnName];
+    NSString *migrationQuery =
+        [NSString stringWithFormat:@"ALTER TABLE \"%@\" ADD COLUMN \"%@\" %@", kMSLogTableName, kMSTargetKeyColumnName, kMSSQLiteTypeText];
+    [MSDBStorage executeNonSelectionQuery:migrationQuery inOpenedDatabase:db];
+  }
+  if (version < kMSLogPersistencePriorityVersion) {
+    NSString *migrationQuery = [NSString stringWithFormat:@"ALTER TABLE \"%@\" ADD COLUMN \"%@\" %@", kMSLogTableName,
```
guperrot: Should add `DEFAULT 1` at the end of alter table since default value is 0x01 for normal events in CS3 (critical is 0x02).

AppCenter/AppCenter/MSConstants.h
```diff
@@ -63,3 +63,9 @@ typedef NS_ENUM(NSInteger, MSInitializationPriority) {
...
+  MSFlagsNone = (0 << 0),                // => 00000000
```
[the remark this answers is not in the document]
jaelim-ms: This is the one that we talked in review meeting. I'm the person who does't like to have default or none value here (for both platforms) but we made the final decision that we will add None (for Apple platform) and DEFAULT (for Android platform).

AppCenterPush/AppCenterPush/MSPush.m
```diff
@@ -247,7 +248,7 @@ - (NSString *)convertTokenToString:(NSData *)token {
 - (void)sendPushToken:(NSString *)token {
   MSPushLog *log = [MSPushLog new];
   log.pushToken = token;
-  [self.channelUnit enqueueItem:log];
+  [self.channelUnit enqueueItem:log flags:MSFlagsNone];
```
jaelim-ms: Should it be "Normal"?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -225,7 +225,7 @@ - (void)sendFirstSessionUpdateLog {
   MSDistributionStartSessionLog *log = [[MSDistributionStartSessionLog alloc] init];
 
   // Send log to log manager.
-  [self.channelUnit enqueueItem:log];
+  [self.channelUnit enqueueItem:log flags:MSFlagsNone];
```
jaelim-ms: Should it be `Normal`?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -1161,12 +1166,13 @@ - (void)trackModelException:(MSException *)exception withProperties:(NSDictionar
...
-    log.properties =
-        [MSUtility validateProperties:properties forLogName:[NSString stringWithFormat:@"ErrorLog: %@", log.errorId] type:log.type];
+    log.properties = [MSUtility validateProperties:properties
+                                        forLogName:[NSString stringWithFormat:@"ErrorLog: %@", log.errorId]
+                                              type:log.type];
   }
 
   // Enqueue log.
-  [self.channelUnit enqueueItem:log];
+  [self.channelUnit enqueueItem:log flags:MSFlagsNone];
```
jaelim-ms: Should it be `Normal`?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -847,7 +850,7 @@ - (void)sendErrorAttachments:(NSArray<MSErrorAttachmentLog *> *)errorAttachments
       MSLogError([MSCrashes logTag], @"Not all required fields are present in MSErrorAttachmentLog.");
       continue;
     }
-    [self.channelUnit enqueueItem:attachment];
+    [self.channelUnit enqueueItem:attachment flags:MSFlagsNone];
```
jaelim-ms: Should it be `Normal`?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -772,7 +775,7 @@ - (void)processLogBufferAfterCrash {
 
           // Buffered logs are used sending their own channel. It will never contain more than 50 logs.
           MSLogDebug([MSCrashes logTag], @"Re-enqueueing buffered log, type: %@.", item.type);
-          [self.bufferChannelUnit enqueueItem:item];
+          [self.bufferChannelUnit enqueueItem:item flags:MSFlagsNone];
```
jaelim-ms: Should it be `Normal`?

AppCenterAnalytics/AppCenterAnalyticsTests/MSEventLogTests.m
```diff
@@ -579,7 +579,7 @@ - (void)testToCommonSchemaLogForTargetToken {
   self.sut.tag = [NSObject new];
 
   // When
-  MSCommonSchemaLog *csLog = [self.sut toCommonSchemaLogForTargetToken:targetToken];
+  MSCommonSchemaLog *csLog = [self.sut toCommonSchemaLogForTargetToken:targetToken flags:0];
```
jaelim-ms: Can we use a real flag instead of a number?

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -146,38 +143,37 @@ - (void)testLoadCommonSchemaLogsWhilePendingBatchesWithSpecificTargetKeys {
   MSCommonSchemaLog *log1 = [MSCommonSchemaLog new];
   [log1 addTransmissionTargetToken:@"1-t"];
   log1.iKey = @"o:1";
-  [self.sut saveLog:log1 withGroupId:kMSTestGroupId];
+  [self.sut saveLog:log1 withGroupId:kMSTestGroupId critical:nil];
```
jwallra: I can't tell what the intended behavior is here, is using `critical:nil` supposed to have the same behavior as using `critical:NO`?  Did you choose to use the former because the test is criticality-agnostic?  I would prefer `criticality:NO` if it doesn't matter since the behavior isn't obvious.

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -452,7 +452,7 @@ - (void)testFiltersNonOneCollectorLogWhenLogHasTargetTokens {
   [transmissionTargetTokens addObject:@"fake-transmission-target-token"];
   MSCommonSchemaLog *commonSchemaLog = [MSCommonSchemaLog new];
   id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
-  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@[ commonSchemaLog ]);
+  OCMStub([mockLog toCommonSchemaLogsWithFlags:0]).andReturn(@[ commonSchemaLog ]);
```
jaelim-ms: Can we use a real flag instead of a number?

AppCenter/AppCenterTests/MSStoragePerfomanceTests.m
```diff
@@ -35,7 +35,7 @@ - (void)testDatabaseWriteShortLogsPerformance {
   NSArray<MSStartServiceLog *> *arrayOfLogs = [self generateLogsWithShortServicesNames:kMSNumLogs withNumService:kMSNumServices];
   [self measureBlock:^{
     for (MSStartServiceLog *log in arrayOfLogs) {
-      [self.dbStorage saveLog:log withGroupId:kMSTestGroupId];
+      [self.dbStorage saveLog:log withGroupId:kMSTestGroupId critical:nil];
```
jaelim-ms: Why do you pass `nil` here?

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -369,14 +369,14 @@ - (void)testDidNotEnqueueLogWhenLogHasNilTargetTokens {
   id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
   OCMStub([mockLog isKindOfClass:[MSCommonSchemaLog class]]).andReturn(NO);
   OCMStub(mockLog.transmissionTargetTokens).andReturn(nil);
-  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@ [[MSCommonSchemaLog new]]);
+  OCMStub([mockLog toCommonSchemaLogsWithFlags:0]).andReturn(@ [[MSCommonSchemaLog new]]);
```
jaelim-ms: Can we use a real flag instead of a number?

AppCenter/AppCenterTests/MSOneCollectorChannelDelegateTests.m
```diff
@@ -264,10 +264,10 @@ - (void)testDidEnqueueLogToOneCollectorChannelSynchronously {
   [transmissionTargetTokens addObject:@"fake-transmission-target-token"];
   MSCommonSchemaLog *commonSchemaLog = [MSCommonSchemaLog new];
   id<MSMockLogWithConversion> mockLog = OCMProtocolMock(@protocol(MSMockLogWithConversion));
-  OCMStub([mockLog toCommonSchemaLogs]).andReturn(@[ commonSchemaLog ]);
+  OCMStub([mockLog toCommonSchemaLogsWithFlags:0]).andReturn(@[ commonSchemaLog ]);
```
jaelim-ms: Can we use a real flag instead of a number?

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -40,27 +40,28 @@ - (instancetype)init {
 
 #pragma mark - Save logs
 
-- (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId {
+- (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId critical:(BOOL)critical {
```
jwallra: Can we make a wrapper for this without the `critical` parameter that always passes `critical:nil`?  It would save passing a bunch of `nil` parameters

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -90,7 +90,7 @@ - (void)ingestionDidReceiveFatalError:(__unused id<MSIngestionProtocol>)ingestio
 
 #pragma mark - Managing queue
 
-- (void)enqueueItem:(id<MSLog>)item {
+- (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
```
jwallra: I think it would also be useful to have a wrapper without the `flags` parameter here that always passes `flags:MSFlagsNone`

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
```diff
@@ -126,10 +125,10 @@ - (void)addTransmissionTargetToken:(NSString *)token {
 
 #pragma mark - MSLogConversion
 
-- (NSArray<MSCommonSchemaLog *> *)toCommonSchemaLogs {
+- (NSArray<MSCommonSchemaLog *> *)toCommonSchemaLogsWithFlags:(int64_t)flags {
```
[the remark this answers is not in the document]
jwallra: Will this change in the future?  How many flags will we have?  I can see the logic for choosing a larger int size but I think if we have 64 flags we're doing something wrong.  I like being explicit, but having a smaller size (`int16_t` or `int32_t` would be my choice).  If we ever get to the point that we have more than 16/32 flags then we probably should be using a different system.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
guperrot: This is not a public API though.

AppCenter/AppCenter/Internals/Model/MSAbstractLog.m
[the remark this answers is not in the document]
guperrot: I think we should use explicit `int64` only when it makes specific sense. For a set of flags, int is fine and would raise less questions.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
guperrot: As discussed would rather switch to the flag directly, so an integer.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
guperrot: Let's avoid using nil indeed, I find it very confusing too.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: None needs to map to default values for each flags as discussed.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: None needs to map to default values for each flags as discussed.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: None needs to map to default values for each flags as discussed.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
guperrot: None needs to map to default values for each flags as discussed.

AppCenterPush/AppCenterPush/MSPush.m
guperrot: None needs to map to default values for each flags as discussed.

guperrot:
In case its buried, this one is still valid: https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1193#pullrequestreview-169581836

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: Oh wait crashes should be critical actually.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: Oh wait crashes should be critical actually.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
guperrot: Not sure about log lines but crashes should be critical while attachments and handled errors are normal.

AppCenter/AppCenter/Internals/Model/MSLogConversion.h
```diff
@@ -12,6 +12,6 @@
  *
  * @return An array of MCSCommonSchemaLog objects.
  */
-- (NSArray<MSCommonSchemaLog *> *)toCommonSchemaLogsWithFlags:(int64_t)flags;
+- (NSArray<MSCommonSchemaLog *> *)toCommonSchemaLogsWithFlags:(NSUInteger)flags;
```
guperrot: Can't we use MSFlags everywhere? The next file in the review list uses it in saveLog.

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -46,6 +46,24 @@ @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
 
+  [MSAppCenter start:@"target=8dfa00f19c154b2ba1308ea49929bc4e-a0715669-ea1e-47d8-928a-dd33841333c5-6753"
```
MatkovIvan: @jwallra This test code broke `setMaxStorageSize` (Unable to set storage size after the application has configured App Center). I guess this was merged by mistake and should be reverted.

PR #1201, opened 2018-11-01
Feature/delete on priority and age

jwallra:

* [x] Did you add unit tests?

## Description

When saving logs to storage, logs with lower priority will now be deleted first when storage is full.  Logs of equal priority will be deleted if there are no other logs with lower priority left.  Logs of higher priority than the inserted log will never be deleted.

### DIFF
--- a/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
+++ b/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
@@ -67,15 +67,28 @@ static const NSUInteger kMSSchemaVersion = 3;
                                    targetKey ? [NSString stringWithFormat:@"'%@'", targetKey] : @"NULL", (unsigned int)persistenceFlags];
   }
   return [self executeQueryUsingBlock:^int(void *db) {
-           // Selecting logs with equal or lower priority and ordering by priority then age.
            int result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
+           NSMutableArray<NSNumber *> *logsCanBeDeleted = nil;
+           if (result == SQLITE_FULL) {
+
+             // Selecting logs with equal or lower priority and ordering by priority then age.
+             NSString *query = [NSString stringWithFormat:@"SELECT \"%@\" FROM \"%@\" WHERE \"%@\" <= %u ORDER BY \"%@\" ASC, \"%@\" ASC",
+                                                          kMSIdColumnName, kMSLogTableName, kMSPriorityColumnName, (unsigned int)flags,
+                                                          kMSPriorityColumnName, kMSIdColumnName];
+             NSArray<NSArray *> *entries = [MSDBStorage executeSelectionQuery:query inOpenedDatabase:db];
+             logsCanBeDeleted = [NSMutableArray new];
+             for (NSMutableArray *row in entries) {
+               [logsCanBeDeleted addObject:row[0]];
+             }
+           }
 
            // If the database is full, delete logs until there is room to add the log.
            long countOfLogsDeleted = 0;
-           NSArray<NSNumber *> *ids = [self logsWithPriorityLowerThan:persistenceFlags inOpenedDatabase:db];
            NSUInteger index = 0;
-           while (result == SQLITE_FULL && index < [ids count]) {
-             result = [MSLogDBStorage deleteLogsFromDBWithColumnValues:@[ ids[index++] ] columnName:kMSIdColumnName inOpenedDatabase:db];
+           while (result == SQLITE_FULL && index < [logsCanBeDeleted count]) {
+             result = [MSLogDBStorage deleteLogsFromDBWithColumnValues:@[ logsCanBeDeleted[index++] ]
+                                                            columnName:kMSIdColumnName
+                                                      inOpenedDatabase:db];
              if (result != SQLITE_OK) {
                break;
              }
@@ -88,8 +101,8 @@ static const NSUInteger kMSSchemaVersion = 3;
            }
            if (result == SQLITE_OK) {
              MSLogVerbose([MSAppCenter logTag], @"Log is stored with id: '%ld'", (long)sqlite3_last_insert_rowid(db));
-           } else if (result == SQLITE_FULL && index == [ids count]) {
-             MSLogDebug([MSAppCenter logTag], @"No logs with equal or lower priority found and storage already full; discarding log.");
+           } else if (result == SQLITE_FULL && index == [logsCanBeDeleted count]) {
+             MSLogDebug([MSAppCenter logTag], @"No logs with equal or lower priority found and storage already full; discarding the log.");
            }
            return result;
          }] == SQLITE_OK;
@@ -302,18 +315,6 @@ static const NSUInteger kMSSchemaVersion = 3;
 
 - (NSUInteger)countLogs {
   return [self countEntriesForTable:kMSLogTableName condition:nil];
-}
-
-- (NSArray<NSNumber *> *)logsWithPriorityLowerThan:(MSFlags)flags inOpenedDatabase:(void *)db {
-  NSString *query =
-      [NSString stringWithFormat:@"SELECT \"%@\" FROM \"%@\" WHERE \"%@\" <= %u ORDER BY \"%@\" ASC, \"%@\" ASC", kMSIdColumnName,
-                                 kMSLogTableName, kMSPriorityColumnName, (unsigned int)flags, kMSPriorityColumnName, kMSIdColumnName];
-  NSArray<NSArray *> *entries = [MSDBStorage executeSelectionQuery:query inOpenedDatabase:db];
-  NSMutableArray *ids = [NSMutableArray new];
-  for (NSMutableArray *row in entries) {
-    [ids addObject:row[0]];
-  }
-  return ids;
 }
 
 #pragma mark - DB migration
--- a/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
+++ b/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
@@ -45,7 +45,7 @@ static const NSUInteger kMSSchemaVersion = 3;
   if (!log) {
     return NO;
   }
-  NSUInteger persistenceFlags = flags & kMSPersistenceFlagsMask;
+  MSFlags persistenceFlags = flags & kMSPersistenceFlagsMask;
 
   // Insert this log to the DB.
   NSData *logData = [NSKeyedArchiver archivedDataWithRootObject:log];
@@ -67,12 +67,15 @@ static const NSUInteger kMSSchemaVersion = 3;
                                    targetKey ? [NSString stringWithFormat:@"'%@'", targetKey] : @"NULL", (unsigned int)persistenceFlags];
   }
   return [self executeQueryUsingBlock:^int(void *db) {
+           // Selecting logs with equal or lower priority and ordering by priority then age.
            int result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
 
            // If the database is full, delete logs until there is room to add the log.
            long countOfLogsDeleted = 0;
-           while (result == SQLITE_FULL) {
-             result = [MSLogDBStorage deleteOldestLogsWithCount:1 inOpenedDatabase:db];
+           NSArray<NSNumber *> *ids = [self logsWithPriorityLowerThan:persistenceFlags inOpenedDatabase:db];
+           NSUInteger index = 0;
+           while (result == SQLITE_FULL && index < [ids count]) {
+             result = [MSLogDBStorage deleteLogsFromDBWithColumnValues:@[ ids[index++] ] columnName:kMSIdColumnName inOpenedDatabase:db];
              if (result != SQLITE_OK) {
                break;
              }
@@ -80,10 +83,13 @@ static const NSUInteger kMSSchemaVersion = 3;
              result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
            }
            if (countOfLogsDeleted > 0) {
-             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) deleted.", (long)countOfLogsDeleted);
+             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) with lower priority deleted.",
+                        (long)countOfLogsDeleted);
            }
            if (result == SQLITE_OK) {
              MSLogVerbose([MSAppCenter logTag], @"Log is stored with id: '%ld'", (long)sqlite3_last_insert_rowid(db));
+           } else if (result == SQLITE_FULL && index == [ids count]) {
+             MSLogDebug([MSAppCenter logTag], @"No logs with lower priority found and storage already full; discarding log.");
            }
            return result;
          }] == SQLITE_OK;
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
                                    targetKey ? [NSString stringWithFormat:@"'%@'", targetKey] : @"NULL", (unsigned int)persistenceFlags];
   }
   return [self executeQueryUsingBlock:^int(void *db) {
+           // Selecting logs with equal or lower priority and ordering by priority then age.
```
jwallra: Note - this has been formatted by the script

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
+             result = [MSLogDBStorage deleteLogsFromDBWithColumnValues:@[ ids[index++] ] columnName:kMSIdColumnName inOpenedDatabase:db];
              if (result != SQLITE_OK) {
                break;
              }
              ++countOfLogsDeleted;
              result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
            }
            if (countOfLogsDeleted > 0) {
-             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) deleted.", (long)countOfLogsDeleted);
+             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) with lower priority deleted.",
```
jwallra: Message should be "of equal or lower priority"

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
+                        (long)countOfLogsDeleted);
            }
            if (result == SQLITE_OK) {
              MSLogVerbose([MSAppCenter logTag], @"Log is stored with id: '%ld'", (long)sqlite3_last_insert_rowid(db));
+           } else if (result == SQLITE_FULL && index == [ids count]) {
+             MSLogDebug([MSAppCenter logTag], @"No logs with lower priority found and storage already full; discarding log.");
```
jwallra: Message should be "equal or lower priority"

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
            int result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
 
            // If the database is full, delete logs until there is room to add the log.
            long countOfLogsDeleted = 0;
-           while (result == SQLITE_FULL) {
-             result = [MSLogDBStorage deleteOldestLogsWithCount:1 inOpenedDatabase:db];
+           NSArray<NSNumber *> *ids = [self logsWithPriorityLowerThan:persistenceFlags inOpenedDatabase:db];
+           NSUInteger index = 0;
+           while (result == SQLITE_FULL && index < [ids count]) {
+             result = [MSLogDBStorage deleteLogsFromDBWithColumnValues:@[ ids[index++] ] columnName:kMSIdColumnName inOpenedDatabase:db];
```
ElektrojungeMS: Maybe it's just me, but having `@[ ids[index++] ]` is a bit too much nesting? What do y'all think?

achocron: I don't mind that nesting but I don't have a strong optionion

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
   }
   return [self executeQueryUsingBlock:^int(void *db) {
+           // Selecting logs with equal or lower priority and ordering by priority then age.
            int result = [MSDBStorage executeNonSelectionQuery:addLogQuery inOpenedDatabase:db];
 
            // If the database is full, delete logs until there is room to add the log.
            long countOfLogsDeleted = 0;
-           while (result == SQLITE_FULL) {
-             result = [MSLogDBStorage deleteOldestLogsWithCount:1 inOpenedDatabase:db];
+           NSArray<NSNumber *> *ids = [self logsWithPriorityLowerThan:persistenceFlags inOpenedDatabase:db];
```
achocron: Does this return logs ordered by age? If so, I think the method name should reflect that (for readability)

achocron: And if not, I believe it should

achocron: Remember that if we are getting SQLITE_FULL, we have already filled the database, which has a default max size of 10mb. So that means that in the default case, we are loading around 10mb of logs every time a single log is added (and this code path is executed). Further, we don't actually need all 10mb, we probably only need 1 or 2 logs.

This method seems to load all logs instead of the old way which loads them in chunks of 1. I propose we load them in chunks of, say, 5 logs at a time? That way we aren't loading all the logs in storage for every new log that comes through. (I think even just 1 log at a time would be better).

Does anyone else have any thoughts on that?

achocron: In the default case when this method is being invoked, we will be loading around 10 mb of data (since that is the default max storage size). In the typical case, we will be loading even more (since we expect devs to increase it for their app). Further, we will only really need to delete 1 or 2 logs on average to make room. 

Thus, I think we are loading several megabytes of data in memory when we probably need only a couple of kilobytes at the most. This seems very inefficient.

I propose that we load the logs in chunks of, say, 5 (but that's arbitrary). The previous implementation loaded 1 at a time, which would still be much more efficient. Also, the Android implementation seems to load logs individually.

Does anyone have thoughts on this?

jwallra: Yeah I think that this would be a good optimization.  If we're not concerned with matching Android exactly, I think that loading a few 5-10 logs at a time would be good.  

Is it also worth testing the worst-case scenarios to see if/how much the performance decreases if we have to load all the logs anyways?  I would be a bit concerned about the case where we have a DB full of critical logs and we try to add a normal one.  I think there are probably some customers who will run into this, and if it's much worse we should use larger chunks.

achocron: I would say we get chunks that are already filtered by priority, so in the latter case the query would return 0 instead of 5

achocron: Bsically just adding a count parameter to the method would do the trick

jaelim-ms: This method doesn't load whole logs. It only returns database ID which is a primary keys for logs. It should be okay.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
clpolet: I don't mind either.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
guperrot: I do share the concern though, can we do this to match Android cursor? https://stackoverflow.com/a/12387605/598520 the step thing.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
guperrot: If it's within formatter max width before splitting, then it should be fine.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -67,23 +67,29 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
            }
            if (countOfLogsDeleted > 0) {
-             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) deleted.", (long)countOfLogsDeleted);
+             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) with equal or lower priority deleted.",
+                        (long)countOfLogsDeleted);
            }
            if (result == SQLITE_OK) {
              MSLogVerbose([MSAppCenter logTag], @"Log is stored with id: '%ld'", (long)sqlite3_last_insert_rowid(db));
+           } else if (result == SQLITE_FULL && index == [ids count]) {
+             MSLogDebug([MSAppCenter logTag], @"No logs with equal or lower priority found and storage already full; discarding log.");
```
clpolet: Just reading through the code but `result == SQLITE_FULL && index == [ids count]` can happen for two reasons. First as described "No logs with equal or lower priority found and storage already full", that's when `index` and `[ids count]` are both 0. But the second situation is when `index` and `[ids count]` are equals but > 0, it happens when all the lower priority logs got deleted but that wasn't enough and the DB is still full. So maybe we want to revisit the logging here to log those two situations separately.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -622,36 +625,168 @@ - (void)testAddLogsDoesNotExceedCapacity {
   }
 }
 
-- (void)testOldestLogsAreDeletedFirstWhenCapacityIsReached {
+- (void)testOldestNormalPriorityLogsAreDeletedFirstWhenCapacityIsReachedAndNewNormalPriorityLogComes {
```
guperrot: I would suggest using shorter such as `pugNormalLogPurgesOldestNormalPriorityLogs`.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
clpolet: It's also not just lower it can be equal priority :)

AppCenter/AppCenterTests/MSLogDBStorageTests.m
guperrot: I would suggest using shorter such as `putNormalPriorityLogPurgesOldestNormalPriorityLogs`.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -622,36 +625,168 @@ - (void)testAddLogsDoesNotExceedCapacity {
...
+  XCTAssertFalse([self containsLogWithDbId:firstLogDbId]);
+
+  NSString *whereCondition = [NSString stringWithFormat:@"\"%@\" = '%@'", kMSGroupIdColumnName, kMSAnotherTestGroupId];
+  NSArray<id<MSLog>> *loadedLogs = [self loadLogsWhere:whereCondition];
+  XCTAssertEqual(loadedLogs.count, 1);
+  XCTAssertEqualObjects(loadedLogs[0].sid, additionalLog.sid);
+  XCTAssertEqual(1, [self findUnknownDBIdsFromKnownIdList:addedDbIds].count);
+}
+
+- (void)testOldestNormalPriorityLogsAreDeletedFirstWhenCapacityIsReachedAndNewCriticalPriorityLogComes {
```
guperrot: This is hard to read, I would suggest something shorter such as `putCriticalPriorityLogPurgesOldestNormalPriorityLogs`.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -622,36 +625,168 @@ - (void)testAddLogsDoesNotExceedCapacity {
...
+  XCTAssertFalse([self containsLogWithDbId:firstLogDbId]);
+
+  NSString *whereCondition = [NSString stringWithFormat:@"\"%@\" = '%@'", kMSGroupIdColumnName, kMSAnotherTestGroupId];
+  NSArray<id<MSLog>> *loadedLogs = [self loadLogsWhere:whereCondition];
+  XCTAssertEqual(loadedLogs.count, 1);
+  XCTAssertEqualObjects(loadedLogs[0].sid, additionalLog.sid);
+  XCTAssertEqual(1, [self findUnknownDBIdsFromKnownIdList:addedDbIds].count);
+}
+
+- (void)testNewNormalPriorityLogIsDiscardedWhenCapacityIsReachedWithAllCriticalPriorityLogs {
```
guperrot: Same comment for all titles, as suggested before there are ways to make it shorter without losing meaningful information.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
achocron: @jaelim-ms that's true, my mistake. but it could still be a lot so I think it might be worth changing.

clpolet: Even 1 at a time should do the trick but I would check the >= priority only if the database is full. Having the DB full is a bit of an edge for when the connection in down or advanced scenario where the transmission is paused for a while and even then the DB needs to reach its full capacity.

jaelim-ms: @clpolet That was considered while implementing it but I forgot to address. 😄 
@achocron It is even worse without this call. Deleting one log at a time will do the full scan anyway. I'll change this to call only when the db is full.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
jaelim-ms: Both conditions are actually same. Eventually the log will be discarded and the only difference is that there are some logs deleted before discarding the log.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
jaelim-ms: Just few lines of code. I removed the method.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
jaelim-ms: Everyone has different preferences. :cry:

ElektrojungeMS: Let's not block on preferences.

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -68,22 +68,41 @@ - (BOOL)saveLog:(id<MSLog>)log withGroupId:(NSString *)groupId flags:(MSFlags)fl
...
            }
            if (countOfLogsDeleted > 0) {
-             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) deleted.", (long)countOfLogsDeleted);
+             MSLogDebug([MSAppCenter logTag], @"Log storage was over capacity, %ld oldest log(s) with equal or lower priority deleted.",
+                        (long)countOfLogsDeleted);
            }
            if (result == SQLITE_OK) {
              MSLogVerbose([MSAppCenter logTag], @"Log is stored with id: '%ld'", (long)sqlite3_last_insert_rowid(db));
+           } else if (result == SQLITE_FULL && index == [logsCanBeDeleted count]) {
+             MSLogDebug([MSAppCenter logTag], @"No logs with equal or lower priority found and storage already full; discarding the log.");
```
dhei: wording - "No logs with equal or lower priority **were** found and storage **is** already full"

PR #1202, opened 2018-11-02
Load higher priority logs first

jaelim-ms:
* [x] Did you add unit tests?

## Description

Change the storage to return higher priority logs first for batches.

### DIFF
--- a/AppCenter/AppCenterTests/MSLogDBStorageTests.m
+++ b/AppCenter/AppCenterTests/MSLogDBStorageTests.m
@@ -50,8 +50,10 @@ static const long kMSTestStorageSizeMinimumUpperLimitInBytes = 10 * kMSDefaultPa
 
   // If
   NSUInteger expectedLogsCount = 5;
-  NSMutableArray *expectedLogs = [[self generateAndSaveLogsWithCount:expectedLogsCount + 1 groupId:kMSTestGroupId
-                                                               flags:MSFlagsDefault] mutableCopy];
+  NSMutableArray *expectedLogs = [[self generateAndSaveLogsWithCount:expectedLogsCount + 1
+                                                             groupId:kMSTestGroupId
+                                                               flags:MSFlagsDefault
+                                              andVerifyLogGeneration:YES] mutableCopy];
   [expectedLogs removeLastObject];
 
   // When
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -923,11 +983,27 @@ - (void)testMigrationFromSchema2to3 {
...
+                              andVerifyLogGeneration:(BOOL)verify {
+  return [self generateAndSaveLogsWithCount:count size:size groupId:groupId flags:flags storage:self.sut andVerifyLogGeneration:verify];
+}
+
+- (NSArray<id<MSLog>> *)generateAndSaveLogsWithCount:(NSUInteger)count
+                                                size:(NSNumber *)size
+                                             groupId:(NSString *)groupId
+                                               flags:(MSFlags)flags
+                                             storage:(MSDBStorage *)storage
+                              andVerifyLogGeneration:(BOOL)verify {
```
ElektrojungeMS: I understand why we do a verification as part of the log creation but the violation of the single responsibility principle makes me uneasy. Still fine as long as it's a conscious decision for a method that's used in tests.

Another question I have is: why do we need the flag? Why don't we verify every log we create?

jaelim-ms: The reason is on the "SELECT" statement below. We need to create logs with random prioirity but SELECT tries to verify all logs stored in the database for specific priority.
For example, if I'd like to generate 2 normal logs, 2 critical logs, and then 2 normal logs, the third generation for 2 normal logs will fail on verification since it expected to create 2 logs but there are 4 logs.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -66,11 +68,60 @@ - (void)testLoadTooManyLogs {
   XCTAssertTrue(moreLogsAvailable);
 }
 
-- (void)testLoadJustEnoughLogs {
+- (void)testLoadJustEnoughNormalLogs {
```
guperrot: Here we skip priority in titles, if that is fine then we can remove priority from titles in the other PR that has very long titles.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -941,11 +1017,14 @@ - (void)testMigrationFromSchema2to3 {
...
 
-  // Check the insertion worked.
-  trueLogCount = [storage countEntriesForTable:kMSLogTableName
-                                     condition:[NSString stringWithFormat:@"\"%@\" = '%@' AND \"%@\" = %u", kMSGroupIdColumnName, groupId,
-                                                                          kMSPriorityColumnName, (unsigned int)flags]];
-  assertThatUnsignedInteger(trueLogCount, equalToUnsignedInteger(count));
+  if (verify) {
+
+    // Check the insertion worked.
+    trueLogCount = [storage countEntriesForTable:kMSLogTableName
```
guperrot: What is a logCount that is not true? A non working log count? Not sure what is the intent here.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -66,11 +68,60 @@ - (void)testLoadTooManyLogs {
...
+                                                                      andVerifyLogGeneration:NO]] mutableCopy];
+
+  // Create 2 critical logs.
+  expectedLogs = [[expectedLogs arrayByAddingObjectsFromArray:[self generateAndSaveLogsWithCount:segmentLogCount
+                                                                                         groupId:kMSTestGroupId
+                                                                                           flags:MSFlagsPersistenceCritical
+                                                                          andVerifyLogGeneration:NO]] mutableCopy];
+
+  // Build expected logs
+  expectedLogs = [[expectedLogs arrayByAddingObjectsFromArray:normalLogs] mutableCopy];
```
guperrot: How does this handle the order check?

AppCenter/AppCenterTests/MSLogDBStorageTests.m
ElektrojungeMS: Good point and +1 for consistency!

AppCenter/AppCenterTests/MSLogDBStorageTests.m
ElektrojungeMS: I see. That's not obvious, can we add a comment that mentions this?

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -314,7 +375,7 @@ - (void)testDeleteLogsWithGroupId {
   // If
   self.sut = [MSLogDBStorage new];
   //  [self.sut.batches removeAllObjects];
```
ElektrojungeMS: Remove commented out code.

PR #1209, opened 2018-11-09
Database page size shouldn't be hard-coded

MatkovIvan:
* [ ] Did you add unit tests?

## Description

The page size can be configured and default value has been changed in 2016:
https://www.sqlite.org/pgszchng2016.html


AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -208,74 +215,87 @@ - (void)migrateDatabase:(void *)__unused db fromVersion:(NSUInteger)__unused ver
...
+  NSArray<NSArray *> *rows = [MSDBStorage executeSelectionQuery:@"PRAGMA page_count;" inOpenedDatabase:db];
+  return [(NSNumber *)rows[0][0] longValue];
+}
+
++ (long)getMaxPageCountInOpenedDatabase:(void *)db {
+  NSArray<NSArray *> *rows = [MSDBStorage executeSelectionQuery:@"PRAGMA max_page_count;" inOpenedDatabase:db];
+  return [(NSNumber *)rows[0][0] longValue];
+}
+
++ (int)setMaxPageCount:(long)maxPageCount inOpenedDatabase:(void *)db {
```
clpolet: Can't we use the `executeNonSelectionQuery:inOpenedDatabase:` here?

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -28,18 +31,22 @@ - (instancetype)initWithSchema:(MSDBSchema *)schema version:(NSUInteger)version
...
 - (int)executeQueryUsingBlock:(MSDBStorageQueryBlock)callback {
   int result;
-  sqlite3 *db = [self openDatabaseAtFileURL:self.dbFileURL withMaxPageCount:self.maxPageCount withResult:&result];
+  sqlite3 *db = [MSDBStorage openDatabaseAtFileURL:self.dbFileURL withResult:&result];
   if (!db) {
     return result;
   }
+
+  // The value is stored as part of the database connection and must be reset every time the database is opened.
+  long maxPageCount = self.maxSizeInBytes / self.pageSize;
```
clpolet: Do we really want to do the calculation every time we open the DB?

MatkovIvan: At first I got page size every time and at that moment it made sense =) but then I started to cache the page size because it cannot be changed on existing database.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: Probably. That code was as it is, but I will consider this refactoring

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: Actually, one integer division operation doesn't make any impact on performance. So, it's not big deal.
But we are setting and operating with size in bytes, so storing `maxSizeInBytes` property is more useful than some **internal** value.

clpolet: I guess it's fine since we don't use the result anywhere else.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -9,13 +9,17 @@ @implementation MSDBStorage
...
-    _maxPageCount = [MSDBStorage numberOfPagesInBytes:kMSDefaultDatabaseSizeInBytes];
+    _maxSizeInBytes = kMSDefaultDatabaseSizeInBytes;
 
     // If it is custom SQLite library we need to turn on URI filename capability.
     sqlite3_config(SQLITE_CONFIG_URI, 1);
 
     // Execute all initialize operation with one database instance.
-    [self executeQueryUsingBlock:^int(void *db) {
+    int result;
+    sqlite3 *db = [MSDBStorage openDatabaseAtFileURL:self.dbFileURL withResult:&result];
```
jaelim-ms: Seems `result` is ignored. Can you pass `nil` instead of a reference?

MatkovIvan: It's logged inside (just like before). No without refactoring the method to accept `nil`. Is it really necessary?

jaelim-ms: No. 😄

PR #1220, opened 2018-11-15
Min database size changed with index

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Fix minimum storage size verification to match actual value.
The minimum storage size is 24 KiB to be able to send the default logs (start service, start session, push installation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # App Center SDK for iOS and macOS Change Log
 
+## Version 1.12.0 (Under active development)
+
+### AppCenter
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # App Center SDK for iOS and macOS Change Log
 
+## Version 1.12.0 (Under active development)
```
dhei: Let's use `1.11.1` for now and bump to `1.12.0` when we need to.

jaelim-ms: We actually know that the next would be 1.12.0 but I agree this one since there is no features already added.

PR #1228, opened 2018-11-20
Add ext.app.userId to common schema

guperrot:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Standalone PR to share code between 2 tasks related to userId support in Common Schema.
* Add userId to model for Common Schema.
* Add the validation regex.

PR #1229, opened 2018-11-20
Add setAppUserId in PropertyConfigurator

guperrot:

* [x] Did you add unit tests?

## Description

Add setAppUserId in PropertyConfigurator.

## Related PRs or issues

#1228

### DIFF
--- a/AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.h
+++ b/AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.h
@@ -26,6 +26,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setAppLocale:(nullable NSString *)appLocale;
 
 /**
+ * Set the user identifier.
+ * The user identifier needs to start with c: or i: or d: or w: prefixes.
+ *
+ * @param appUserId user identifier.
+ */
+- (void)setAppUserId:(nullable NSString *)appUserId;
+
+/**
  * Set a string event property to be attached to events tracked by this transmission target and its child transmission targets.
  *
  * @param propertyValue Property value.
--- a/AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m
+++ b/AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m
@@ -12,6 +12,7 @@
 #import "MSAppExtension.h"
 #import "MSCSExtensions.h"
 #import "MSCommonSchemaLog.h"
+#import "MSConstants+Internal.h"
 #import "MSDeviceExtension.h"
 #import "MSEventPropertiesInternal.h"
 #import "MSLogger.h"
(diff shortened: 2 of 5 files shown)


AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m
```diff
@@ -46,6 +47,26 @@ - (void)setAppLocale:(NSString *)appLocale {
   _appLocale = appLocale;
 }
 
+- (void)setAppUserId:(NSString *)appUserId {
+  @synchronized([MSAnalytics sharedInstance]) {
```
jaelim-ms: Any reason for synchronization? If we need this then we need to synchronize other methods.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
```diff
@@ -1140,15 +1160,81 @@ - (void)testOverridingCommonSchemaPropertiesWithTwoChildrenUnderTheSameParent {
...
+  log.ext.appExt = [MSAppExtension new];
+
+  // When
+  [target.propertyConfigurator channel:nil prepareLog:log];
+
+  // Then
+  XCTAssertNil(log.ext.appExt.userId);
+}
+
+- (void)testOverridingValidAppUserIdThenUnset {
```
dhei: In addition to this test case, can we add a test case where overriding with valid AppUserId then overriding with invalid AppUserId to make sure that the second AppUserId doesn't override the first AppUserId?

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.h
```diff
@@ -25,6 +25,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setAppLocale:(nullable NSString *)appLocale;
 
+/**
+ * Set the user identifier.
+ * The user identifier needs to start with c: or i: or d: or w: prefixes.
```
clpolet: Try to fit everything in one line as much as possible otherwise the doc may get truncated by Xcode. To try it out keep the "option/alt" key pressed while clicking the method, Xcode will compute  the documentation.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m
guperrot: The check is not atomic with the update (if). We also do that in other existing methods that have more than just a variable assignment.

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.h
guperrot: screen shot 2018-11-20 at 1 30 33 pm

To me it looks good with the current, not sure what you are asking?

clpolet: Let me try again, not sure which version of Xcode didn't work.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTransmissionTargetTests.m
guperrot: Done in https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1229/commits/30851986d2b5b4679249622a2ad729fe948c0a4c

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m
guperrot: Also it creates the static cached regex only once

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.h
clpolet: Nevermind, that was Xcode 8 picking up the doc from the property instead of the setter :man_facepalming:. Xcode 10.1 takes the right doc. 
screen shot 2018-11-20 at 13 36 34

PR #1236, opened 2018-11-22
Use typed properties for count label

MatkovIvan:

### Repro steps:

- Build and run SasquatchPuppet app.
- Go to analytics tab
- Add property to event
- Track event
- Press "Results" button

### Actual:
Last event info is incorrect. Properties count field is always 0.
### Expected:
Show correct property count value.

### DIFF
--- a/Sasquatch/Sasquatch/Sasquatch-Bridging-Header.h
+++ b/Sasquatch/Sasquatch/Sasquatch-Bridging-Header.h
@@ -6,3 +6,7 @@
 #import "Constants.h"
 #import "CrashLib.h"
 #import "MSEventFilter.h"
+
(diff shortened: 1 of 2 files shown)


Sasquatch/Sasquatch/Sasquatch-Bridging-Header.h
```diff
@@ -6,3 +6,7 @@
 #import "Constants.h"
 #import "CrashLib.h"
 #import "MSEventFilter.h"
+
+#if GCC_PREPROCESSOR_MACRO_PUPPET
```
jaelim-ms: Why only for puppet app?

MatkovIvan: This is private API

PR #1238, opened 2018-11-27
Updating setUserId API to match new spec

dhei:

* [x] Did you add unit tests?

## Description

* `app.userId` becomes `user.localId` again
* If 'c:' prefix is missing for 1st party user id, it is now added automatically (only when sent to One Collector). If using both secrets, crashes will not get automatic prefix in the payload.
* If using another prefix than `c:` for `AppCenter.setUserId` when using target token or both, or when using `PropertyConfigurator`, emit an error and don’t accept `userId` update.


dhei:
I tested the following scenarios in Sasquatch Puppet:
- App Center user with valid userId and invalid userId with `startSession` and `event`.
- 1DS user with valid userid and invalid userId with `startSession` and `event`.
- Both app secret and One Collector with valid userid and invalid userId.

AppCenter/AppCenter/Internals/History/UserId/MSUserIdContext.h
```diff
@@ -67,6 +67,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (BOOL)checkUserIdValidForOneCollector:(nullable NSString *)userId;
 
+/**
+ * Add 'c:' prefix to userId if the userId has no prefix.
+ *
+ * @param userId userId.
+ *
+ * @return prefixed userId or null if the userId was null.
+ */
++ (nullable NSString *)toPrefixUserIdFromUserId:(nullable NSString *)userId;
```
achocron: this API should probably be called `prefixedUserIdFromUserId:` (the current name does not quite make sense to me)

jaelim-ms: We have a few internal methods that are `to` prefixed and we followed that convention.
Looks `prefixedUserIdFromUserId` make more sense and you also found a good example of the name, we can rename this one like that.
@ElektrojungeMS @clpolet do you have any suggestions before making change?

jaelim-ms: We also need to rename above two methods in the same class, let's think about it.

AppCenter/AppCenter/Internals/History/UserId/MSUserIdContext.h
```diff
@@ -65,7 +65,16 @@ NS_ASSUME_NONNULL_BEGIN
...
++ (BOOL)isUserIdValidForOneCollector:(nullable NSString *)userId;
+
+/**
+ * Add 'c:' prefix to userId if the userId has no prefix.
+ *
+ * @param userId userId.
+ *
+ * @return prefixed userId or null if the userId was null.
+ */
++ (nullable NSString *)prefixUserIdFromUserId:(nullable NSString *)userId;
```
jaelim-ms: prefix**ed**UserIdFromUserId

PR #1239, opened 2018-11-28
Improve error message when 1DS userId has only an invalid prefix and nothing after the prefix

dhei:
* [x] Did you add unit tests?

## Description
When 1DS userId is like `x:` that has an invalid prefix `x` with nothing after the `:`, the error log message is "userId must not be empty."

Now it is "userId prefix must be 'c', 'x' is not supported.".


dhei:
This PR is targeting "feature/implement-spec-change-setuserid" branch now for ease of review. Will target to "feature/develop-userid" branch after merging https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1238.

PR #1241, opened 2018-11-29
Fix/userid for target

jaelim-ms:
* [X] Did you add unit tests?

## Description

Fix bugs
1. UserId override issue when target has no userId but default target has.
2. Add a prefix in setUserId in MSPropertyConfigurator


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -104,9 +104,6 @@ - (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
   if (item && !item.device) {
     item.device = [[MSDeviceTracker sharedInstance] device];
   }
-  if (item && !item.userId) {
```
guperrot: Where is the code to set userId on both crashes and handled exceptions then?

PR #1242, opened 2018-11-30
Fix cleanup of device/session history

guperrot:

* [x] Did you add unit tests?

## Description

Fix cleanup of device/session history

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -8,6 +8,7 @@
 #import "MSCrashesInternal.h"
 #import "MSCrashesPrivate.h"
 #import "MSCrashesUtil.h"
+#import "MSDeviceTracker.h"
 #import "MSEncrypter.h"
 #import "MSErrorAttachmentLog.h"
 #import "MSErrorAttachmentLogInternal.h"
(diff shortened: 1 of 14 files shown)


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -342,7 +343,7 @@ - (void)applyEnabledState:(BOOL)isEnabled {
     [self emptyLogBufferFiles];
     [self removeAnalyzerFile];
     [self.plCrashReporter purgePendingCrashReport];
-    [[MSSessionContext sharedInstance] clearSessionHistory];
+    [self clearContextHistoryAndKeepCurrentSession:NO];
```
jaelim-ms: We should keep the most recent one or store the current session when Crashes is enabled.
I feel this way is more complicated because the Crashes should know it was disabled in the same life cycle.
Instead of differentiate it between disabled and then enabled vs Crashes started and then enabled, we can simply keep the most recent one here.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -51,11 +57,23 @@ @implementation MSCrashesTests
 - (void)setUp {
   [super setUp];
   self.sut = [MSCrashes new];
+  [MSDeviceTracker resetSharedInstance];
+  self.deviceTrackerMock = OCMClassMock([MSDeviceTracker class]);
+  OCMStub(ClassMethod([self.deviceTrackerMock sharedInstance])).andReturn(self.deviceTrackerMock);
```
jaelim-ms: `sharedInstance` only available for class method so `ClassMethod` could be omitted.

PR #1244, opened 2018-11-30
Fix Keychain permission issue in macOS

jaelim-ms:

## Description

MacOS SDK doesn't require any permissions to access keychain but it asks the permission since we share core code base with other Apple platforms and storage/crashes classes initialize MSEncrypter for transmission target token encryption/decryption.
This change will delay Keychain access until the SDK actually encrypts/decrypts transmission target tokens so that it couldn't ask the permission to users whoever use applications that don't send any logs to One Collector.
This will also improve user experience for iOS, any applications that are not sending logs to One Collector will not ask this permission unless the applications use "Distribute" service.

### DIFF
--- a/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
+++ b/AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
@@ -34,7 +34,7 @@ static const NSUInteger kMSSchemaVersion = 3;
     _logColumnIndex = ((NSNumber *)columnIndexes[kMSLogTableName][kMSLogColumnName]).unsignedIntegerValue;
     _targetTokenColumnIndex = ((NSNumber *)columnIndexes[kMSLogTableName][kMSTargetTokenColumnName]).unsignedIntegerValue;
     _batches = [NSMutableDictionary<NSString *, NSArray<NSNumber *> *> new];
-    _targetTokenEncrypter = [[MSEncrypter alloc] initWithDefaultKey];
+    _targetTokenEncrypter = [[MSEncrypter alloc] init];
   }
   return self;
 }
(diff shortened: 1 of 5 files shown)


AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -34,7 +34,7 @@ - (instancetype)init {
     _logColumnIndex = ((NSNumber *)columnIndexes[kMSLogTableName][kMSLogColumnName]).unsignedIntegerValue;
     _targetTokenColumnIndex = ((NSNumber *)columnIndexes[kMSLogTableName][kMSTargetTokenColumnName]).unsignedIntegerValue;
     _batches = [NSMutableDictionary<NSString *, NSArray<NSNumber *> *> new];
-    _targetTokenEncrypter = [[MSEncrypter alloc] initWithDefaultKey];
+    _targetTokenEncrypter = [[MSEncrypter alloc] init];
```
guperrot: Don't we use `new` in new code for that case?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -17,36 +17,54 @@ @interface MSEncrypter ()
...
+  if (self.cryptorCreated) {
+    return;
   }
-  return self;
+
+  // Load a key with keyTag.
+  NSData *key = [MSEncrypter loadKeyFromKeychainWithTag:keyTag];
+  if (!key) {
+
+    // Generate a new one if it doesn't exist.
```
jwallra: I think this comment is pretty obvious, maybe not needed

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -258,7 +258,7 @@ - (instancetype)init {
                                                                       flushInterval:1.0
                                                                      batchSizeLimit:1
                                                                 pendingBatchesLimit:3];
-    _targetTokenEncrypter = [[MSEncrypter alloc] initWithDefaultKey];
+    _targetTokenEncrypter = [[MSEncrypter alloc] init];
```
guperrot: here too then

PR #1250, opened 2018-12-04
Use xcconfig for CrashLib

MatkovIvan:


CrashLib/CrashLib.xcodeproj/project.pbxproj
```diff
@@ -650,333 +729,69 @@
...
-				CODE_SIGN_STYLE = Automatic;
-				COMBINE_HIDPI_IMAGES = YES;
-				DEBUG_INFORMATION_FORMAT = dwarf;
-				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
-				DYLIB_COMPATIBILITY_VERSION = 1;
-				DYLIB_CURRENT_VERSION = 1;
-				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				FRAMEWORK_VERSION = A;
 				INFOPLIST_FILE = CrashLib/Info.plist;
```
clpolet: I was just wondering why `INFOPLIST_FILE` and `PRODUCT_BUNDLE_IDENTIFIER` are kept in the project configuration?

CrashLib/Config/CrashLib.xcconfig
```diff
@@ -0,0 +1,16 @@
+#include "../../Config/App.xcconfig"
+
+PRODUCT_NAME = $(TARGET_NAME)
+MS_SIM_ARCHS = x86_64
```
ElektrojungeMS: I think we need to add `i386` for now.

CrashLib/CrashLib.xcodeproj/project.pbxproj
MatkovIvan: The settings like team, provision profile, etc are ignored because it's easier to maintain it from the editor

PR #1252, opened 2018-12-05
Remove http calls suspending

MatkovIvan:

## Description

Align the behavior with Android SDK. It should not have any visible impact on customers.

### DIFF
--- a/AppCenter/AppCenter/Internals/Ingestion/MSHttpIngestion.m
+++ b/AppCenter/AppCenter/Internals/Ingestion/MSHttpIngestion.m
@@ -58,7 +58,7 @@ static NSString *const kMSPartialURLComponentsName[] = {@"scheme", @"user", @"pa
 
     // Set query parameter.
     [queryStrings
-        enumerateKeysAndObjectsUsingBlock:^(id _Nonnull key, id _Nonnull queryString, __attribute__((unused)) BOOL *_Nonnull stop) {
+        enumerateKeysAndObjectsUsingBlock:^(id _Nonnull key, id _Nonnull queryString, __unused BOOL *_Nonnull stop) {
           [queryStringForEncoding
               appendString:[NSString stringWithFormat:@"%@%@=%@", [queryStringForEncoding length] > 0 ? @"&" : @"", key, queryString]];
         }];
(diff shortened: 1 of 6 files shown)


ElektrojungeMS:
The code change itself looks good. This new behavior is what we intended to do and we have confirmed it works as intended, correct?

clpolet:
@MatkovIvan Are we sure that a suspended http task ignores incoming response from servers while suspended? Can we test that with `OHHttpStubs`, not sure?

MatkovIvan:
@jaelim-ms It doesn't checks network connection. It checks the **error** if it's caused by network connection. What more suspiciously here is the check of http status code (only 200 is successful instead of 2xx), but it can be fixed in the separate PR

@clpolet The target of the PR is aligning the behavior with Android SDK. I didn't face with this issue during testing. It's a decision from you and Jae, right?

jaelim-ms:
@clpolet That's caused by [this line](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/3d5be7fe8c17a563fcc044794d86ea7902fb8f7b/AppCenter/AppCenter/Internals/Ingestion/MSIngestionCall.m#L84). We suspended ongoing calls before, it won't pause Ingestion from the method when network is already down.
Now we are not suspending ongoing calls, it could get results or errors after network is down and pause Ingestion if an error is caused by network down while Ingestion is already paused.
@MatkovIvan I think you can simply remove the line.

jaelim-ms:
I spent more time to see if the suggestion fixes the issue but it still have the issue.
After removing the line, I still see OneCollector ingestion that is paused twice but I could only see AppCenter ingestion that is paused only once. We probably need more investigation on it since `networkStateChanged` is called multiple times (2 - 4 times).

jaelim-ms:
Sorry for wrong commits.

MatkovIvan:
This is an reachability issue, it sends 3 events when the network is lost

```
2018-12-07 11:28:29.041752+0300 SasquatchPuppet[33797:4529950] Reachability Flag Status: -- ------- networkStatusForFlags
2018-12-07 11:28:29.041854+0300 SasquatchPuppet[33797:4529950] Reachability Flag Status: -R ------- networkStatusForFlags
2018-12-07 11:28:29.043302+0300 SasquatchPuppet[33797:4529950] Reachability Flag Status: -- ------- networkStatusForFlags
```

jaelim-ms:
We need to file a bug for reachability issue and fix it separately. However, the line that I mentioned above should be removed.

clpolet:
Also about "only 200 is successful instead of 2xx", it's definitely another topic. Of course they mean success but they shouldn't be handled the same way all together.

clpolet:
I just found another error code we can add to the "there is no network detection": -1005:
```
2018-12-07 15:26:36.304559-0800 SasquatchPuppet[637:122483] Task <6A9D2735-FA98-41A8-A1A3-74AC7CBCE197>.<7> finished with error - code: -1005
2018-12-07 15:26:36.320069-0800 SasquatchPuppet[637:122559] [AppCenter] DEBUG: -[MSHttpIngestion sendCallAsync:]_block_invoke/209 HTTP request error with code: -1005, domain: NSURLErrorDomain, description: The network connection was lost.
```
It's small enough so we can add it to this PR I think.

MatkovIvan:
@clpolet I've left `TODO` for 2xx. JFYI Android SDK handles all 2xx in the same way:

https://github.com/Microsoft/AppCenter-SDK-Android/blob/0659771e486ecfe1886076714c04e81cb0fc73af/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java#L276

PR #1254, opened 2018-12-06
Cleanup build scripts

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Remove tons of duplicated code.


jaelim-ms:
The purpose of this change is to extract part of duplicated script in the files but this change keeps all of those and only splits the build script files.
iOS, macOS, tvOS as a platform and a list of SDK could be passed to a single shell script as parameters.
Only part that you need to extract is [this](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/develop/Scripts/AppCenter/build-appcenter-ios-framework.sh#L3-L45).

Scripts/AppCenter/build-appcenter-ios-framework.sh
```diff
@@ -1,88 +1,10 @@
...
-# Lipo the binaries that were built with various Xcode versions.
-env DEVELOPER_DIR="${MS_ARM64E_XCODE_PATH}" lipo -create "${LIB_IPHONEOS_TEMP_DIR}/lib${FMK_NAME}.a" "${LIB_IPHONEOS_FINAL}" -output "${LIB_IPHONEOS_FINAL}"
-fi
-
-echo "Use arm64e Xcode and lipo -create the fat binary."
-env DEVELOPER_DIR="$MS_ARM64E_XCODE_PATH" lipo -create "${LIB_IPHONEOS_FINAL}" "${SIMULATOR_DIR}/lib${FMK_NAME}.a" -output "${INSTALL_DIR}/${FMK_NAME}"
-
-#End of arm64e code block.
-fi
+source $(dirname "$0")/../build-ios-framework.sh AppCenter
```
clpolet: Any reasons not to just use `./` instead of `$(dirname "$0")` in these script?

guperrot: The latter would allow you to run the script from another directory I guess.

MatkovIvan: The current directory isn't `Scripts/AppCenter`

MatkovIvan:
@jaelim-ms As discussed in Slack, this part is different for each platform too.

Scripts/build-ios-framework.sh
```diff
@@ -49,7 +49,14 @@ cp -f "${SRCROOT}/${FMK_NAME}/Support/iOS.modulemap" "${INSTALL_DIR}/Modules/mod
 
 # Copies the headers and resources files to the final product folder.
 cp -R "${SRCROOT}/${WRK_DIR}/Release-iphoneos/include/${FMK_NAME}/" "${INSTALL_DIR}/Headers/"
+
+# Copy the resource bundle for App Center Distribute.
+if [[ "$FMK_NAME" == "AppCenterDistribute" ]]; then
+echo "Copying resource bundle for $FMK_NAME."
 cp -R "${SRCROOT}/${WRK_DIR}/Release-iphoneos/${FMK_RESOURCE_BUNDLE}.bundle" "${PRODUCTS_DIR}" || true
```
MatkovIvan: @ElektrojungeMS @clpolet The error ignored by `|| true`, so it was not an error, just an message.
Personally I dislike the check of module name. I suggest the next options to replace it:
- Check if bundle exists
- Silent output

clpolet: I agree, my vote goes for "Check if bundle exists" since having a bundle named `${FMK_RESOURCE_BUNDLE}.bundle` seems to be conventional.

PR #1255, opened 2018-12-06
Enable Part B data

achocron:

* [x] Did you add unit tests?

## Description
Update Part B validation to allow setting baseType/baseData.

### DIFF
--- a/AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
+++ b/AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
@@ -98,6 +98,7 @@ static NSString *const kMSTypedProperties = @"typedProperties";
   if (self.typedProperties) {
     csProperties = [NSMutableDictionary new];
     metadata = [NSMutableDictionary new];
+    NSString *baseTypePrefix = [NSString stringWithFormat:@"%@.", kMSDataBaseType];
     NSString *baseDataPrefix = [NSString stringWithFormat:@"%@.", kMSDataBaseData];
 
     // If baseType is set and valid, make sure it's paired with at least 1 "baseData.*" property.
@@ -115,7 +116,7 @@ static NSString *const kMSTypedProperties = @"typedProperties";
       }
     }
 
-    // If there is no valid "baseType" property (could be set but invalid type), there must not be any "baseData.*" properties.
+    // If there is no valid "baseType" property, there must not be any "baseData.*" property.
     else {
       BOOL removedBaseData = NO;
       for (NSString *key in [self.typedProperties.properties allKeys]) {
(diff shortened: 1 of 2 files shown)


AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
```diff
@@ -98,11 +98,49 @@ - (void)setPropertiesAndMetadataForCSLog:(MSCommonSchemaLog *)csLog {
   if (self.typedProperties) {
     csProperties = [NSMutableDictionary new];
     metadata = [NSMutableDictionary new];
+    NSString *baseDataPrefix = [NSString stringWithFormat:@"%@.", kMSDataBaseData];
```
jaelim-ms: Synced offline. We should validate baseType and baseData and then remove properties if one of them is missing.

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
```diff
@@ -98,11 +98,46 @@ - (void)setPropertiesAndMetadataForCSLog:(MSCommonSchemaLog *)csLog {
...
+
+    // If there is no valid "baseType" property, there must not be "baseData.*" properties.
+    if (![self.typedProperties.properties[kMSDataBaseType] isKindOfClass:[MSStringTypedProperty class]]) {
```
jaelim-ms: nit: can you swap if block and else block by removing `!` from the condition?
```
if (!condition) {
  // Do A
} else {
  // Do B
}
```
can be
```
if (condition) {
  // Do B
} else {
  // Do A
}
```

AppCenterAnalytics/AppCenterAnalytics/Internals/Model/MSEventLog.m
```diff
@@ -98,11 +98,46 @@ - (void)setPropertiesAndMetadataForCSLog:(MSCommonSchemaLog *)csLog {
...
+      }
+      if (removedBaseData) {
+        MSLogWarning([MSAnalytics logTag], @"baseData was set but baseType is missing or invalid.");
+      }
+
+      // Base type might be set but invalid, so remove it.
+      [self.typedProperties.properties removeObjectForKey:kMSDataBaseType];
+    }
+
+    // If there are no "baseData.*" properties, there must not be a "baseType" property.
```
jwallra: nit: I prefer not to split if/else clauses this way, I would prefer for the comment to be just inside the else block

guperrot: That's not what we do usually and the comment to apply to the block. It should be same level as the other comment.

PR #1256, opened 2018-12-07
Clear unprocessed reports after processing

bmourat:
* [x] Did you add unit tests?

## Description
Fixes a bug with duplicate logs being sent from React-Native wrapper when automatic processing is on.


clpolet:
That's a good point from @guperrot , I think we need clean up there too.
.

achocron:
@guperrot @clpolet I have the same question for around line 348. Might be fine, just want to make sure.

PR #1264, opened 2018-12-12
Use default build settings for CrashLib and Sasquatch

ElektrojungeMS:

* ~[ ] Has `CHANGELOG.md` been updated?~ Not required as this is a change for the test apps.
* ~[ ] Did you add unit tests?~ This is pure configuration. No unit tests needed.

## Description

We discovered that the quality for C++ exceptions were better than expected, so we investigated. Turns out we distributed release builds with more debug information than usual which resulted in better crash reports. 

* Use default setting for “Strip Debug Symbols During Copy”
* Use default setting for “Debug Information Format”


Config/App.xcconfig
```diff
@@ -12,9 +12,6 @@ ONLY_ACTIVE_ARCH_Release = NO
 ONLY_ACTIVE_ARCH = $(ONLY_ACTIVE_ARCH_$(CONFIGURATION))
 
 // :Mark: Build Options
```
clpolet: nit: should we also get rid of this comment?

Issue #1269, opened 2019-01-07
Verify no buildSettings shell script isn't escaping paths


ksuther:
### **Description**

The `Verify no buildSettings` shell scripts fail when the project is in a path that requires escaping (such as spaces in directory names).

The variables in the shell scripts need to be quoted:

`${SRCROOT}/../Scripts/VerifyNoBuildSettings.swift ${PROJECT_NAME}.xcodeproj/project.pbxproj`

Should be something like:

`"${SRCROOT}"/../Scripts/VerifyNoBuildSettings.swift "${PROJECT_NAME}.xcodeproj/project.pbxproj"`

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Clone repository to a path with a space in one of the directories
2. Try to build

### **Details**

Cloned at 449bc09406974a7a7e45bd8007ad38dbe6144a42

guperrot:
This was fixed in version [1.13.0](https://github.com/Microsoft/AppCenter-SDK-Apple/releases/tag/1.13.0).

PR #1272, opened 2019-01-07
Escape all paths in pbxproj shell scripts that use SRCROOT

achocron:

* [ ] ~Are tests passing locally?~

Was able to reproduce, then verify fix locally.

## Description

Enable the SDK frameworks to be built when the absolute path of the source contains escaped characters.

## Related PRs or issues

Issue #1269.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # App Center SDK for iOS and macOS Change Log
 
-## Version 1.12.1
+## Version 1.12.1 (Under active development)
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # App Center SDK for iOS and macOS Change Log
 
+## Version 1.12.1
```
jaelim-ms: `(Under active development)` at the end.

PR #1274, opened 2019-01-11
Fix target memberships in the repo

jaelim-ms:

## Description

Some files has wrong targets or missing targets and private headers are also set to project level. This change fixes all targets in the repo.


achocron:
@jaelim-ms Just out of curiosity, what exactly is the difference between private and project level headers?

PR #1276, opened 2019-01-13
Add supporting of User Id to macOS application

AndreiShilkin:
## Description
- update delegates to set up User Id;
- update view controller and storyboard to provide User Id text field to set up this.

### Updated UI:
image

### User Id field state & logs for each state:
1. Empty User Id field:
UI:
image
Logs:
image

2. Set up User Id:
UI:
image
Logs:
image

### DIFF
--- a/SasquatchMac/SasquatchMacObjC/AppDelegate.m
+++ b/SasquatchMac/SasquatchMacObjC/AppDelegate.m
@@ -19,6 +19,12 @@ static NSString *const kSMLogTag = @"[SasquatchMac]";
 - (void)applicationDidFinishLaunching:(NSNotification *)notification {
   [MSAppCenter setLogLevel:MSLogLevelVerbose];
 
+  // Set user id.
+  NSString *userId = [[NSUserDefaults standardUserDefaults] objectForKey:@"userId"];
+  if (userId) {
+    [MSAppCenter setUserId:userId];
+  }
+
   // Customize services.
   [self setupCrashes];
   [self setupPush];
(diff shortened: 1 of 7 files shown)


SasquatchMac/SasquatchMacSwift/AppDelegate.swift
```diff
@@ -43,8 +43,14 @@ class AppDelegate: NSObject, NSApplicationDelegate, MSCrashesDelegate, MSPushDel
     // Push Delegate.
     MSPush.setDelegate(self);
 
-    // Start AppCenter.
     MSAppCenter.setLogLevel(MSLogLevel.verbose)
+    // Set user id.
```
jaelim-ms: We have a coding convention that there should be an empty line before any of comment in the code. Please add an empty line above.

SasquatchMac/SasquatchMacObjC/AppDelegate.m
```diff
@@ -19,6 +19,12 @@ @implementation AppDelegate
 - (void)applicationDidFinishLaunching:(NSNotification *)notification {
   [MSAppCenter setLogLevel:MSLogLevelVerbose];
 
+  // Set user id.
+  NSString *userId = [[NSUserDefaults standardUserDefaults] objectForKey:@"userId"];
```
jaelim-ms: Define a constant for "userId". You can see an example in iOS test app [here](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/develop/Sasquatch/Sasquatch/Constants.h#L32) and [here](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/develop/Sasquatch/SasquatchObjC/AppDelegate.m#L117)

jaelim-ms: I can see hard-coded "userId" in several places. Please address those, too.

SasquatchMac/SasquatchMacSwift/AppDelegate.swift
AndreiShilkin: Fixed. Thank you.

PR #1279, opened 2019-01-15
Move common variables to Constants file

AndreiShilkin:
## Description
- add new `Constants.h` file in SasquatchMac project;
- move common variables from SasquatchMac project to `Constants.h` file;
- reuse constants in dependent files.


jwallra:
Both this PR and #1280 have a lower coverage, but they shouldn't really affect coverage.  Not really sure why

PR #1281, opened 2019-01-17
Release/1.14.0

v-momoin:
## Description

Preparing for Feb Apple SDK release.


v-momoin:
>           We usually only bump the micro version after a release. So that would be 1.13.1 :)
Updated the changes as requested.

PR #1287, opened 2019-01-31
Fix signal handler setup when starting Crashes in background in Xamarin

guperrot:

* [ ] Did you add unit tests?

## Description

Fixes the bug described at https://github.com/Microsoft/AppCenter-SDK-DotNet/issues/759#issuecomment-458920638. And add some logging that was used to find the fix.

The crash delegate was freed before it could even be used.

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -281,6 +281,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
   // Enabling.
   if (isEnabled) {
     id<MSCrashHandlerSetupDelegate> crashSetupDelegate = [MSWrapperCrashesHelper getCrashHandlerSetupDelegate];
+    MSLogDebug([MSCrashes logTag], @"Crashes setup delegate object=%@", crashSetupDelegate);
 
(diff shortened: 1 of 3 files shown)


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -281,6 +281,7 @@ - (void)applyEnabledState:(BOOL)isEnabled {
   // Enabling.
   if (isEnabled) {
     id<MSCrashHandlerSetupDelegate> crashSetupDelegate = [MSWrapperCrashesHelper getCrashHandlerSetupDelegate];
+    MSLogDebug([MSCrashes logTag], @"Crashes setup delegate object=%@", crashSetupDelegate);
```
MatkovIvan: "Setup crash handler with delegate"?
To highlight **the action** we're doing at the moment.

jaelim-ms: I think @guperrot wants to see if `crashSetupDelegate` is `nil` (freed) or not at the moment. What about
```
MSLogDebug([MSCrashes logTag], @"Got crashSetupDelegate for wrapper SDK, delegate=%@", crashSetupDelegate);
```

guperrot:
https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/825 can replace this one so we can keep using the pattern of using weak delegates.

PR #1293, opened 2019-02-01
Fix deadlock on start from background thread

MatkovIvan:
* [ ] Did you add unit tests?

## Description

A deadlock may occur if the AppCenter is started from a background thread.

### DIFF
--- a/AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
+++ b/AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
@@ -121,15 +121,9 @@ static void ReachabilityCallback(SCNetworkReachabilityRef target,
   return [self reachabilityWithAddress:(const struct sockaddr *)&zeroAddress];
 }
 
-#pragma mark reachabilityForLocalWiFi
-// reachabilityForLocalWiFi has been removed from the sample.  See ReadMe.md for
-// more information.
-//+ (instancetype)reachabilityForLocalWiFi
-
 #pragma mark - Start and stop notifier
 
-- (BOOL)startNotifier {
-  __block BOOL returnValue = NO;
+- (void)startNotifier {
   dispatch_block_t block = ^{
     SCNetworkReachabilityContext context = {0, (__bridge void *)(self), NULL,
                                             NULL, NULL};
@@ -138,16 +132,16 @@ static void ReachabilityCallback(SCNetworkReachabilityRef target,
       if (SCNetworkReachabilityScheduleWithRunLoop(self.reachabilityRef,
                                                    CFRunLoopGetCurrent(),
                                                    kCFRunLoopDefaultMode)) {
-        returnValue = YES;
+        [[NSNotificationCenter defaultCenter] postNotificationName:kMSReachabilityChangedNotification
+                                                            object:self];
       }
     }
   };
   if ([NSThread isMainThread]) {
     block();
   } else {
-    dispatch_sync(dispatch_get_main_queue(), block);
+    dispatch_async(dispatch_get_main_queue(), block);
   }
-  return returnValue;
 }
 
 - (void)stopNotifier {
@@ -160,7 +154,7 @@ static void ReachabilityCallback(SCNetworkReachabilityRef target,
   if ([NSThread isMainThread]) {
     block();
   } else {
-    dispatch_sync(dispatch_get_main_queue(), block);
+    dispatch_async(dispatch_get_main_queue(), block);
   }
 }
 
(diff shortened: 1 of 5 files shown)


AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
```diff
@@ -81,17 +81,11 @@ @implementation MS_Reachability {
...
       SCNetworkReachabilityCreateWithName(NULL, [hostName UTF8String]);
   if (reachability != NULL) {
-    if ([NSThread isMainThread]) {
-      returnValue = [[MS_Reachability alloc] init];
-    } else {
-      dispatch_sync(dispatch_get_main_queue(), ^{
-        returnValue = [[MS_Reachability alloc] init];
-      });
-    }
+    returnValue = [[MS_Reachability alloc] init];
```
clpolet: Yep I'd love do delete this code too 👍 . Though we might need some other sort of synchronization here, I have a vague memory of reachability not being thread safe (cc: @jaelim-ms).

ElektrojungeMS: Yep, Reachability is not thread safe.

ElektrojungeMS: Talked to @MatkovIvan and this is only for the init, all other reachability code is still on main thread. We should be good.

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
```diff
@@ -150,16 +132,16 @@ - (BOOL)startNotifier {
...
+        [[NSNotificationCenter defaultCenter] postNotificationName:kMSReachabilityChangedNotification
+                                                            object:self];
       }
     }
   };
   if ([NSThread isMainThread]) {
     block();
   } else {
-    dispatch_sync(dispatch_get_main_queue(), block);
+    dispatch_async(dispatch_get_main_queue(), block);
```
jaelim-ms: I remember we make a *SYNC* call in main thread due to a deadlock issue. When `startNotifier` and `stopNotifier` calls come together it might cause the issue. Are you sure this is okay?

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
clpolet: I'll unblock it seems like it's not using any shared resources at this time.

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
clpolet: Thinking about it, I don't like using the main thread I would assign reachability it's own serial queue and dispatch async the blocks there. That way we are sure not to bloat the main thread + everyone will use the same serial queue.

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
clpolet: Do we know if `SCNetworkReachabilityCreateWithName(NULL, [hostName UTF8String])` returns a different reference every time? If not is it thread safe? @MatkovIvan

ElektrojungeMS: According to the docs, it returns a new, immutable reference:
`A new immutable reachability reference. You must release the returned value.`

MatkovIvan: Thanks @ElektrojungeMS. Anyway, this isn't changed in the PR.

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
MatkovIvan: @clpolet It was already implemented and then reverted.

@jaelim-ms **sync** cause deadlock because it LOCKS the current thread.

> When `startNotifier` and `stopNotifier` calls come together it might cause the issue.

Nope. It's queued in the main thread.

AppCenter/AppCenter/Internals/Vendor/Reachability/MS_Reachability.m
clpolet: Yeah that looks good.

clpolet:
@ElektrojungeMS about:
> What got me confused was the fact that we were calling `[self networkStateChanged]` explicitly in the init and the `setEnabled:andDeleteDataOnDisabled:` but I don't see a reason for this.

That's because when we initiate the ingestion object or enable it we don't know the state of the network, if at this time the network is down there is no point in sending data.

PR #1305, opened 2019-02-13
Feature/send auth token

achocron:

* [x] Did you add unit tests?

## Description

Sends auth token header with logs.

## Misc

This does not address re-sending push logs or storing encrypted auth token in storage. Those will be addressed in separate PRs.


AppCenter/AppCenter/Internals/Ingestion/MSAppCenterIngestion.m
```diff
@@ -62,6 +64,13 @@ - (NSURLRequest *)createRequest:(NSObject *)data eTag:(NSString *)__unused eTag
   request.allHTTPHeaderFields = self.httpHeaders;
   [request setValue:self.appSecret forHTTPHeaderField:kMSHeaderAppSecretKey];
 
+  // Copy self.authToken into a local variable to avoid a race condition.
+  NSString *authTokenCopy = self.authToken;
+  if (authTokenCopy) {
```
dhei: Probably reasonable to check for empty string of `authToken`

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -0,0 +1,28 @@
+#import <Foundation/Foundation.h>
```
dhei: This import is not necessary.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
```diff
@@ -0,0 +1,19 @@
+#import <Foundation/Foundation.h>
```
dhei: This import is not necessary.

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -0,0 +1,50 @@
+#import <Foundation/Foundation.h>
```
dhei: This import is not necessary.

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -64,6 +64,7 @@ - (void)testKeyIsRestoredFromKeychain {
 - (void)testKeyIsNull {
 
   // If
+  [MSEncrypter deleteKeyWithTag:self.keyTag];
```
dhei: @annakocheshkova We don't need this because it's already called in `tearDown` method.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
achocron: This contains an `NSString` property, so I believe we need the import. Unless we forward declare the `NSString` class. But that's pretty unusual.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
achocron: (same as above)

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -231,4 +236,10 @@ - (void)setMaxStorageSize:(long)sizeInBytes completionHandler:(nullable void (^)
   });
 }
 
+- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(NSString *)authToken {
+  if (self.ingestion) {
```
clpolet: I think I wouldn't assign the authToken to ingestion. It's a tied to a batch of logs but which batch of logs needs what token is a channel unit logic. The current design would work at the moment but I am concerned about the near future.

achocron: But we give all other headers to ingestion and channels do not know about headers. What problem do you foresee? And the solution for making sure the logs get the right headers will still be managed by channel since it decides when to flush, right?

I also think that we can do that change if we need to, but not necessary to do that now. Doing it later wouldn't be a big change, and we may not even need to.

achocron: (I think the order could be flush, and then once ingestion makes the request we update its auth token)

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -231,4 +236,10 @@ - (void)setMaxStorageSize:(long)sizeInBytes completionHandler:(nullable void (^)
   });
 }
 
+- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(NSString *)authToken {
```
jwallra: I'm a bit confused by the parameter name.  Does it mean, "did recieve _this_ auth token"?  To me it looks more like a method name which returns a bool

AppCenter/AppCenterTests/MSAppCenterIngestionTests.m
```diff
@@ -13,6 +13,8 @@
 static NSTimeInterval const kMSTestTimeout = 5.0;
 static NSString *const kMSBaseUrl = @"https://test.com";
 static NSString *const kMSTestAppSecret = @"TestAppSecret";
+static NSString *const kMSAuthorizationHeaderKey = @"Authorization";
```
jwallra: Can you use the same ones as in `MSAppCenterIngestion.m` to avoid duplication?

AppCenter/AppCenterTests/MSAppCenterIngestionTests.m
```diff
@@ -574,6 +576,47 @@ - (void)testCompressHTTPBodyWhenNeeded {
...
+
+  // Then
+  XCTAssertNil([request.allHTTPHeaderFields valueForKey:kMSAuthorizationHeaderKey]);
+}
+
+- (void)testDoesNotSendAuthHeaderWithEmptyAuthToken {
+
+  // If
+  MSLogContainer *logContainer = [[MSLogContainer alloc] initWithBatchId:@"whatever" andLogs:(NSArray<id<MSLog>> *)@ [[MSMockLog new]]];
+  self.sut.authToken  = @"";
```
jwallra:
```suggestion
  self.sut.authToken = @"";
```

jwallra: Probably would be fixed by running formatter

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -92,7 +92,6 @@ - (void)testInitWithSchema {
 
 - (void)testTableExists {
   [self.sut executeQueryUsingBlock:^int(void *db) {
-
```
jwallra: Why are all the blank lines before comments removed?  Anyways you should probably just revert this file since there are no functional changes

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
dhei: The `didRecieveXXX` naming pattern comes from iOS APIs from Apple. 

For example, `application:didReceiveRemoteNotification:fetchCompletionHandler:` https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc.

achocron: I think it's a convention to name callbacks like that in objective c, but I agree that it can be confusing

AppCenter/AppCenterTests/MSAppCenterIngestionTests.m
achocron: Yes I forgot to do that on latest commit. Good catch.

AppCenter/AppCenterTests/MSDBStorageTests.m
achocron: That may have been the formatter. I propose that we leave it because if we don't, this will come up every time

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
clpolet: OK let's go with this for now, keeping it simple and we'll adjust later as needed.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
clpolet: `willAction` and `didAction` is a classic in objective c.

clpolet: And yes you can see the first parameter as the sender of the event.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -23,8 +24,12 @@ - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion {
     _logsDispatchQueue = serialQueue;
     _channels = [NSMutableArray<id<MSChannelUnitProtocol>> new];
     _delegates = [NSHashTable weakObjectsHashTable];
-    _ingestion = ingestion;
     _storage = [MSLogDBStorage new];
+    if (ingestion) {
+      _ingestion = ingestion;
+      _ingestion.authToken = [MSAuthTokenContext sharedInstance].authToken;
+    }
+    [[MSAuthTokenContext sharedInstance] addDelegate:self];
```
clpolet: :thinking: Is it worth adding the delegate if ingestion is nil?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -0,0 +1,28 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol MSAuthTokenContextDelegate;
+
+@interface MSAuthTokenContext : NSObject
```
clpolet: Could you document the class itself too?

AppCenter/AppCenter/Internals/Ingestion/MSAppCenterIngestion.m
```diff
@@ -62,6 +64,13 @@ - (NSURLRequest *)createRequest:(NSObject *)data eTag:(NSString *)__unused eTag
   request.allHTTPHeaderFields = self.httpHeaders;
   [request setValue:self.appSecret forHTTPHeaderField:kMSHeaderAppSecretKey];
 
+  // Copy self.authToken into a local variable to avoid a race condition.
+  NSString *authTokenCopy = self.authToken;
```
guperrot: is the property marked as atomic then? if not need to access it via synchronized?

guperrot: Ah yes it's atomic in the header, all good

AppCenter/AppCenter/Internals/Ingestion/MSAppCenterIngestion.m
```diff
@@ -62,6 +64,13 @@ - (NSURLRequest *)createRequest:(NSObject *)data eTag:(NSString *)__unused eTag
...
+  if (authTokenCopy && ![authTokenCopy isEqualToString:@""]) {
```
clpolet: `if (!([authTokenCopy length] == 0))` should do the trick.

AppCenter/AppCenterTests/MSAppCenterIngestionTests.m
clpolet: Can probably reuse the ingestion ones.

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -29,14 +30,15 @@ - (void)setUp {
...
+  [MSAuthTokenContext sharedInstance].authToken = nil;
 }
 
 #pragma mark - Tests
 
 - (void)testNewInstanceWasInitialisedCorrectly {
 
   // If
-  id ingestionMock = OCMProtocolMock(@protocol(MSIngestionProtocol));
+  MSAppCenterIngestion *ingestionMock = OCMPartialMock([MSAppCenterIngestion new]);
```
clpolet: Any reason to change to a partial mock?

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
guperrot: Since the token is not used by One Collector it would not be useful according to me.

jaelim-ms: I agree with you guys but I think we don't need to logically couple auth token context with ingestion. These are independent operations even though token will be sent to Ingestion.

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
jaelim-ms: Agreed, why?

AppCenter/AppCenter/Internals/Ingestion/MSAppCenterIngestion.m
jaelim-ms: Maybe `if ([authTokenCopy length] > 0)`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -0,0 +1,28 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol MSAuthTokenContextDelegate;
```
jaelim-ms: Consider using internal header for methods only used internally.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -0,0 +1,70 @@
...
+  sharedInstance = nil;
+}
+
+- (NSString *)authToken {
+  @synchronized(self) {
+    return _authToken;
+  }
+}
+
+- (void)setAuthToken:(NSString *)authToken {
```
jaelim-ms: Is it okay to run this in main thread? Since browser is associated in login process, seems it will always be called in main thread. If we do flush during this time then it will block UI thread until it initiates flush.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
jaelim-ms: I agree with you both guys but we need to have a discussion about it before moving forward. This is like a design concern so we should have agreement and be on the same page.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
clpolet: Isn't that ` _ingestion.authToken = [MSAuthTokenContext sharedInstance].authToken;` logically coupled already? Either way not a blocker to me.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
clpolet: It's internal already since part of the "Internals" folder. When there is no "public" header for the class we don't bother explicitly mark the header filename with "Internal".

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
achocron: Yes there is a reason - technically, we were passing `id<MSIngestionProtocol>` to a method that expects `MSAppCenterIngestion *`, and this was not noticed until now because we are now actually using a method in the latter that doesn't exist in the former.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
achocron: That is true. We can run on a background thread or expect `setAuthToken` to be invoked on a background thread. I prefer the latter because the caller has the responsibility to run this on a background thread, not the callee.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
achocron: Ingestion doesn't know about `MSAuthTokenContext`, and `MSAuthTokenContext` doesn't know about Ingestion. In my opinion, they are already decoupled.

AppCenter/AppCenter/Internals/Ingestion/MSAppCenterIngestion.m
clpolet: Even better Jae :)

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
clpolet: :thinking: good point, my guess is it's probably multithreaded and synchronized but yes we should double check.

PR #1309, opened 2019-02-14
Resend push log after logged in

dhei:

* [x] Did you add unit tests?

## Description

Implement resend push log after a new auth token is received.

### DIFF
--- a/AppCenter/AppCenterTests/MSAuthTokenContextTests.m
+++ b/AppCenter/AppCenterTests/MSAuthTokenContextTests.m
@@ -45,4 +45,18 @@
   OCMVerify([delegateMock authTokenContext:self.sut didReceiveAuthToken:expectedAuthToken]);
 }
 
+- (void)testRemoveDelegate {
+
+  // If
+  id<MSAuthTokenContextDelegate> delegateMock = OCMProtocolMock(@protocol(MSAuthTokenContextDelegate));
+  [self.sut addDelegate:delegateMock];
+
+  // Then
+  OCMReject([delegateMock authTokenContext:self.sut didReceiveAuthToken:OCMOCK_ANY]);
+
+  // When
+  [self.sut removeDelegate:delegateMock];
+  self.sut.authToken = @"something";
+}
+
 @end
--- a/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
+++ b/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
@@ -25,7 +25,7 @@ static NSString *const kMSPushNotificationOldCustomDataKey = @"mobile_center";
 
 @property(nonatomic) id<MSPushDelegate> delegate;
 
-@property(nonatomic) NSString *pushToken;
+@property(atomic) NSString *pushToken;
 
 #if TARGET_OS_OSX
 @property(nonatomic) id<NSUserNotificationCenterDelegate> originalUserNotificationCenterDelegate;
--- a/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
+++ b/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
@@ -1,7 +1,7 @@
+#import "MSAuthTokenContextDelegate.h"
 #import "MSPush.h"
 #import "MSPushDelegate.h"
 #import "MSServiceInternal.h"
-#import "MSAuthTokenContextDelegate.h"
 
 /**
  * Keys for payload in push notification.
--- a/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
+++ b/AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
@@ -25,7 +25,7 @@ static NSString *const kMSPushNotificationOldCustomDataKey = @"mobile_center";
 
 @property(nonatomic) id<MSPushDelegate> delegate;
 
-@property(atomic) NSString *pushToken;
+@property(atomic, copy) NSString *pushToken;
 
 #if TARGET_OS_OSX
 @property(nonatomic) id<NSUserNotificationCenterDelegate> originalUserNotificationCenterDelegate;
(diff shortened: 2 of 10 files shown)


AppCenterPush/AppCenterPush/MSPush.m
```diff
@@ -389,4 +392,10 @@ - (BOOL)didReceiveRemoteNotification:(NSDictionary *)userInfo fromUserNotificati
   return NO;
 }
 
+- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(__unused NSString *)authToken {
+  if (self.pushToken) {
```
guperrot: Unless you can guarantee it's always same thread accessing pushToken (and that would deserve comments), accessing self.pushToken after a if is not thread safe.
Either mark pushToken atomic and use a local variable before the if or use synchronized.

annakocheshkova: This will execute if Push is disabled.

annakocheshkova: We decided that the best place to add / remove delegate is in `applyEnabledState`

AppCenterPush/AppCenterPushTests/MSPushTests.m
```diff
@@ -60,6 +62,9 @@ @implementation MSPushTests
 - (void)setUp {
   [super setUp];
   [MSUserIdContext resetSharedInstance];
+
+  // Need to reset defaults because enabled state gets persisted on disk otherwise, causing test inconsistencies.
```
dhei: This comment is out of date, should be removed.

AppCenterPush/AppCenterPush/MSPush.m
```diff
@@ -389,4 +392,13 @@ - (BOOL)didReceiveRemoteNotification:(NSDictionary *)userInfo fromUserNotificati
   return NO;
 }
 
+- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(__unused NSString *)authToken {
+
+  // Make a copy of push token so that this code is thread safe.
```
guperrot: pointer copy, not value (we don't the copy flag on the property), I think the comment is confusing.
we could be more implicit like use a local reference to be thread safe?

dhei: Fixed by adding `copy` attribute to `pushToken` property.

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -45,4 +45,18 @@ - (void)testSetAuthToken {
...
+  id<MSAuthTokenContextDelegate> delegateMock = OCMProtocolMock(@protocol(MSAuthTokenContextDelegate));
+  [self.sut addDelegate:delegateMock];
+
+  // Then
+  OCMReject([delegateMock authTokenContext:self.sut didReceiveAuthToken:OCMOCK_ANY]);
+
+  // When
+  [self.sut removeDelegate:delegateMock];
+  self.sut.authToken = @"something";
+}
```
ElektrojungeMS: it's not obvious to me why we are setting the auth token after removing a delegate. I also don't see an actual test after `// When` or maybe I'm missing something?

AppCenterPush/AppCenterPush/Internal/Model/MSPushLog.h
```diff
@@ -7,6 +7,6 @@
 /**
  * Push token for push service
  */
-@property(nonatomic) NSString *pushToken;
+@property(nonatomic, copy) NSString *pushToken;
```
ElektrojungeMS: Nice!

200w_d

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
ElektrojungeMS: Gosh, I just realized how the test works! Sorry about my comment.

AppCenterPush/AppCenterPush/MSPush.m
guperrot: Ok I was thinking of just fixing the comment instead actually, not sure you need a value copy here.

AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
```diff
@@ -17,14 +18,14 @@ static NSString *const kMSPushNotificationOldCustomDataKey = @"mobile_center";
...
+@interface MSPush () <NSUserNotificationCenterDelegate, MSAuthTokenContextDelegate>
 #else
-@interface MSPush ()
+@interface MSPush () <MSAuthTokenContextDelegate>
 #endif
 
 @property(nonatomic) id<MSPushDelegate> delegate;
 
-@property(nonatomic) NSString *pushToken;
+@property(atomic) NSString *pushToken;
```
guperrot: Then according to https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1309/files#r258166953 this one should be copy too.

AppCenterPush/AppCenterPush/MSPush.m
guperrot: Synced offline, should be copy for best practice.

AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
dhei: Good catch, thanks a lot.

AppCenterPush/AppCenterPushTests/MSPushTests.m
```diff
@@ -60,6 +62,8 @@ @implementation MSPushTests
 - (void)setUp {
   [super setUp];
   [MSUserIdContext resetSharedInstance];
+
```
jwallra: Remove extra new line

AppCenterPush/AppCenterPush/Internal/MSPushPrivate.h
ElektrojungeMS: yes, it should! Sorry for missing this!

PR #1326, opened 2019-02-22
Acquire an authentication token silently

ElektrojungeMS:

* ~[ ] Has `CHANGELOG.md` been updated?~
* [x] Did you add unit tests?
* ~[] Did you check UI tests on the sample app? They are not executed on CI.~

## Description

This PR tries to authenticate silently if possible to avoid the user having to login every time. It will fall back to an "interactive" login if silent login fails.

### DIFF
--- a/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
+++ b/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
@@ -1,3 +1,5 @@
+#import <MSAL/MSALPublicClientApplication.h>
+
 #import "MSChannelDelegate.h"
 #import "MSCustomApplicationDelegate.h"
 #import "MSIdentity.h"
@@ -70,6 +72,18 @@ static NSString *const kMSIdentityMSALAccountEnvironmentKey = @"MSIdentityMSALAc
  * Config MSAL client.
  */
 - (void)configAuthenticationClient;
+
+/**
+ * Retrieve an updated token without user interaction.
+ *
+ * @param account The MSALAccount that is used to retrieve an authentication token.
+ */
+- (void)acquireTokenSilentlyWithMSALAccount:(MSALAccount *)account;
+
+/**
+ * Retrieve an updated token with user interaction.
+ */
+- (void)acquireTokenInteractively;
 
 @end
 
--- a/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
+++ b/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
@@ -37,7 +37,7 @@ static NSString *const kMSIdentityMSALAccountEnvironmentKey = @"MSIdentityMSALAc
 /**
  * The flag that indicates a user requested signIn before it is configured.
  */
-@property(nonatomic) BOOL signInDelayed;
+@property(nonatomic) BOOL signInDelayedAndRetryLater;
 
 /**
  * Custom application delegate dedicated to Identity.
--- a/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
+++ b/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
@@ -17,9 +17,6 @@ static NSString *const kMSIdentityAuthTokenKey = @"MSIdentityAuthToken";
 // The key for the MSALAccount homeAccountId stored in user defaults.
 static NSString *const kMSIdentityMSALAccountHomeAccountKey = @"MSIdentityMSALAccountHomeAccount";
 
-// The key for the MSALAccount environment stored in user defaults.
-static NSString *const kMSIdentityMSALAccountEnvironmentKey = @"MSIdentityMSALAccountEnvironment";
-
 @class MSALPublicClientApplication;
 
 @interface MSIdentity () <MSServiceInternal, MSChannelDelegate>
--- a/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
+++ b/AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
@@ -12,6 +12,12 @@ static NSString *const kMSIdentityETagKey = @"MSIdentityETagKey";
 // The key for Identity auth token stored in keychain.
 static NSString *const kMSIdentityAuthTokenKey = @"MSIdentityAuthToken";
 
+// The key for the MSALAccount homeAccountId stored in user defaults.
+static NSString *const kMSIdentityMSALAccountHomeAccountKey = @"MSIdentityMSALAccountHomeAccount";
+
+// The key for the MSALAccount environment stored in user defaults.
+static NSString *const kMSIdentityMSALAccountEnvironmentKey = @"MSIdentityMSALAccountEnvironment";
+
 @class MSALPublicClientApplication;
 
 @interface MSIdentity () <MSServiceInternal, MSChannelDelegate>
--- a/AppCenterIdentity/AppCenterIdentity/MSIdentity.m
+++ b/AppCenterIdentity/AppCenterIdentity/MSIdentity.m
@@ -161,24 +161,9 @@ static dispatch_once_t onceToken;
     return;
   }
   self.loginDelayed = NO;
-  __weak typeof(self) weakSelf = self;
   MSALAccount *account = [self retrieveAccount];
   if (account) {
-    [self.clientApplication
-        acquireTokenSilentForScopes:@[ (NSString * _Nonnull) self.identityConfig.identityScope ]
-                            account:account
-                    completionBlock:^(MSALResult *result, NSError *e) {
-                      typeof(self) strongSelf = weakSelf;
-                      if (e) {
-                        MSLogWarning([MSIdentity logTag],
-                                     @"Silent acquisition of token failed with error: %@. Triggering interactive acquisition", e);
-                        [strongSelf acquireTokenInteractively];
-                      } else {
-                        [MSAuthTokenContext sharedInstance].authToken = result.idToken;
-                        [strongSelf saveAuthToken:result.idToken];
-                        [strongSelf saveAccountId:(NSString * _Nonnull) result.account.homeAccountId];
-                      }
-                    }];
+    [self acquireTokenSilentlyWithMSALAccount:account];
   } else {
     [self acquireTokenInteractively];
   }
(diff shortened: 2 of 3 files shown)


AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -310,4 +305,60 @@ - (void)removeAuthToken {
...
+  __weak typeof(self) weakSelf = self;
+  [self.clientApplication
+      acquireTokenSilentForScopes:@[ (NSString * _Nonnull) self.identityConfig.identityScope ]
+                          account:account
+                  completionBlock:^(MSALResult *result, NSError *e) {
+                    typeof(self) strongSelf = weakSelf;
+                    if (e) {
+                      MSLogWarning([MSIdentity logTag],
+                                   @"Silent acquisition of token failed with error: %@. Triggering interactive acquisition", e);
+                      [strongSelf acquireTokenInteractively];
```
achocron: Should we really trigger interactive token acquisition here, or is it better to have the app call login again?

AppCenterIdentity/AppCenterIdentityTests/MSIdentityTests.m
```diff
@@ -455,7 +441,119 @@ - (void)testSignInDelayedWhenNoIdentityConfig {
...
+  OCMStub([identityMock sharedInstance]).andReturn(identityMock);
+  OCMStub([identityMock canBeUsed]).andReturn(YES);
+
+  // Then
+  OCMReject([self.clientApplicationMock acquireTokenForScopes:OCMOCK_ANY completionBlock:OCMOCK_ANY]);
+
+  // When
+  [MSIdentity login];
+
+  //  // Then
```
dhei: Clean the extra `//`

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -310,4 +305,60 @@ - (void)removeAuthToken {
...
+                                }];
+}
+
+- (MSALAccount *)retrieveAccount {
+  NSString *homeAccountId = [[MSUserDefaults shared] objectForKey:kMSIdentityMSALAccountHomeAccountKey];
+  if (homeAccountId) {
+    NSError *error;
+    MSALAccount *account = [self.clientApplication accountForHomeAccountId:homeAccountId error:&error];
+    if (error) {
+      MSLogWarning([MSIdentity logTag], @"Could not get MSALAccount for homeAccountId.");
```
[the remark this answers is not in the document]
dhei: @achocron If `accountForHomeAccountId:error:` API returns an error, the `account` should be `nil`, hence it's not necessary to do explicitly return `nil`.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -310,4 +305,60 @@ - (void)removeAuthToken {
...
+  __weak typeof(self) weakSelf = self;
+  [self.clientApplication acquireTokenForScopes:@[ (NSString * _Nonnull) self.identityConfig.identityScope ]
+                                completionBlock:^(MSALResult *result, NSError *e) {
+                                  if (e) {
+                                    MSLogError([MSIdentity logTag], @"User sign-in failed. Error: %@", e);
+                                  } else {
+                                    typeof(self) strongSelf = weakSelf;
+                                    [MSAuthTokenContext sharedInstance].authToken = result.idToken;
+                                    [strongSelf saveAuthToken:result.idToken];
+                                    [strongSelf saveAccountId:(NSString * _Nonnull) result.account.homeAccountId];
```
annakocheshkova: Are you sure this conversion is ok? `homeAccountId` has a type with three different properties.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -310,4 +305,60 @@ - (void)removeAuthToken {
...
+                                  }
+                                }];
+}
+
+- (MSALAccount *)retrieveAccount {
```
annakocheshkova: Please add a method similar to `saveAccountId` / `removeAccountId` for consistency (and I would also pass account id as a parameter here - due to single responsibility principle).
The account id getter is also going to be useful in Push implementation (we need to retrieve account id there).

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
annakocheshkova: I guess you should use `homeAccountId.identifier` here, it is declared as `nonnull` so you also won't have to do this `nullable` to `nonnull` conversion which looks super-unsafe to me.

dhei: @annakocheshkova good catch, will fix this.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
dhei: make sense, will do

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
annakocheshkova: @dhei I partly fixed this in my branch too, (to be able to test), but I guess it will not produce serious merge conflicts.

annakocheshkova:
@dhei @ElektrojungeMS I have some free time today so I resolved  mu two comments on PR, hope it's ok 🙂

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -353,6 +352,10 @@ - (MSALAccount *)retrieveAccount {
   }
```
dhei: nit: the if-else can be simplified by removing else and directly returning nil.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -353,6 +352,10 @@ - (MSALAccount *)retrieveAccount {
...
 }
 
+- (NSString *)getAccountId {
```
dhei: We should call it `retrieveAccountId:` to align with `(MSALAccount *)retrieveAccountWithAccountId:(NSString *)homeAccountId` API.

AppCenterIdentity/AppCenterIdentityTests/MSIdentityTests.m
```diff
@@ -67,59 +70,66 @@ - (void)testApplyEnabledStateWorks {
...
+  NSString *expectedHomeAccountId = @"fakeHomeAccountId";
+  MSALAccountId *mockAccountId = OCMPartialMock([MSALAccountId new]);
+  OCMStub(mockAccountId.identifier).andReturn(expectedHomeAccountId);
+  id accountMock = OCMPartialMock([MSALAccount new]);
+  OCMStub([accountMock homeAccountId]).andReturn(mockAccountId);
+  [self.settingsMock setObject:expectedHomeAccountId forKey:kMSIdentityMSALAccountHomeAccountKey];
   [MSMockKeychainUtil storeString:expectedToken forKey:kMSIdentityAuthTokenKey];
+  OCMStub([self.clientApplicationMock accountForHomeAccountId:expectedHomeAccountId error:[OCMArg setTo:nil]]).andReturn(accountMock);
+  MSIdentity *identityMock = OCMPartialMock(self.sut);
+  OCMStub([identityMock configAuthenticationClient]).andReturn(nil);
```
dhei: Looks like the partial mock of MSIdentity is not necessary, and `[identityMock configAuthenticationClient]` is not needed. You can use `self.sut` instead  of the partial mock of MSIdentity, which is the pattern in this class.

dhei: @annakocheshkova I reverted your change in `testEnablingReadsAuthTokenFromKeychainAndSetsAuthContext:` to keep it simple. As the name suggests, this test case only covers the scenario that MSAuthTokenContext authtoken is correctly set after enabling. IMO you don't need to cover home accountId because there're other tests covering that.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -310,4 +305,62 @@ - (void)removeAuthToken {
...
+}
+
+- (MSALAccount *)retrieveAccountWithAccountId:(NSString *)homeAccountId {
+  if (!homeAccountId) {
+    return nil;
+  }
+  NSError *error;
+  MSALAccount *account = [self.clientApplication accountForHomeAccountId:homeAccountId error:&error];
+  if (error) {
+    MSLogWarning([MSIdentity logTag], @"Could not get MSALAccount for homeAccountId.");
```
clpolet: Should we print the error message from the error?

AppCenterIdentity/AppCenterIdentity/Internals/MSIdentityPrivate.h
```diff
@@ -12,6 +14,12 @@ static NSString *const kMSIdentityETagKey = @"MSIdentityETagKey";
 // The key for Identity auth token stored in keychain.
 static NSString *const kMSIdentityAuthTokenKey = @"MSIdentityAuthToken";
 
+// The key for the MSALAccount homeAccountId stored in user defaults.
+static NSString *const kMSIdentityMSALAccountHomeAccountKey = @"MSIdentityMSALAccountHomeAccount";
+
+// The key for the MSALAccount environment stored in user defaults.
+static NSString *const kMSIdentityMSALAccountEnvironmentKey = @"MSIdentityMSALAccountEnvironment";
```
clpolet: What is this one used for? Looks like we can get rid of it.

dhei: good catch.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
dhei: The scenario is if we get an error from `accountForHomeAccountId:homeAccountId:` from MSAL API, which means we won't be able to use the persisted `accountId` to acquire token silently from MSAL, we will "fall back" to interactive login.

It is not an error/failure case, therefore it shouldn't be an error log. Warning level should be more appropriate.

clpolet: True I agree with that but I was more talking about should we print the message from the error object that we get from MSAL inside our warning log :sweat_smile:. That wasn't clear sorry fro the confusion.

dhei: Sorry I misread that. It's fixed now.

PR #1328, opened 2019-02-24
[Identity] Don't resend push log if refreshing token while running

annakocheshkova:

* [X] Did you add unit tests?

## Description
- Don't resend push log if refreshing token while running (i.e. if userId does not change).
- Send anonymous Push log when Identity is disabled.

**_One important detail:_**
I decided not to use synthesized getter/setter on `authToken` in `AuthTokenContext`, becasue we have one more required paramenter on setter (which is `homeAccountId`). If somebody makes a mistake and uses a default setter (without account id), it would lead to confusion and being unable to distinguish users.

## Related PRs or issues


AppCenterPush/AppCenterPushTests/MSPushTests.m
```diff
@@ -222,14 +222,47 @@ - (void)testSendsPushTokenWhenNewAuthTokenReceived {
...
   // When
-  [MSAuthTokenContext sharedInstance].authToken = @"something";
+  [[MSAuthTokenContext sharedInstance] setAuthToken:@"token1" withAccountId:account1];
+  [[MSAuthTokenContext sharedInstance] setAuthToken:@"token1" withAccountId:account2];
 
   // Then
   [pushMock verify];
 }
 
+- (void)testSendsPushTokenAnonimouslyWhenClearsAuthToken {
```
jwallra:
```suggestion
- (void)testSendsPushTokenAnonymouslyWhenClearsAuthToken {
```

jwallra:
```suggestion
- (void)testSendsPushTokenAnonymouslyWhenClearsAuthToken {
```

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -27,7 +27,7 @@ - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion {
     _storage = [MSLogDBStorage new];
     if (ingestion) {
       _ingestion = ingestion;
-      _ingestion.authToken = [MSAuthTokenContext sharedInstance].authToken;
+      _ingestion.authToken = [[MSAuthTokenContext sharedInstance] getAuthToken];
```
ElektrojungeMS: The "getter" for the token should be named just `authToken` as per ObjC convention.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -34,6 +29,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ * Sets current auth token and account id.
+ */
+- (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId;
+
+/**
+ * Retrieves auth token.
+ */
+- (nullable NSString *)getAuthToken;
```
dhei: method name should be `authToken:` instead of `getAuthToken` per Objective C convention.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -34,6 +29,21 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)removeDelegate:(id<MSAuthTokenContextDelegate>)delegate;
 
+/**
+ * Clears cached token and account id.
```
dhei: nit: We don't add `s` to verbs in docs with third person singular:

`Clear`, `Set` and `Retrieve`.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -9,6 +9,16 @@
 
 @interface MSAuthTokenContext ()
 
+/**
+ * Authorization token cached value.
+ */
+@property(nonatomic) NSString *authToken;
```
dhei: `copy` for NSString properties.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -9,6 +9,16 @@
...
+
+/**
+ * The last value of user account id.
+ */
+@property(nonatomic) NSString *lastHomeAccountId;
```
dhei: `copy` for NSString properties.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
dhei: `authToken` should be `nullable`, right?

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -236,7 +236,8 @@ - (void)setMaxStorageSize:(long)sizeInBytes completionHandler:(nullable void (^)
   });
 }
 
-- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(NSString *)authToken {
+- (void)authTokenContext:(__unused MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(NSString *)authToken
+              forNewUser:(__unused BOOL)isNewUser {
```
dhei: @clpolet @ElektrojungeMS @jaelim-ms What do you think about this method naming? The `forNewUser:` part sounds strange to me.

clpolet: "forNewUser" I think we are expecting a user object not really a Boolean.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -94,10 +94,12 @@ - (void)applyEnabledState:(BOOL)isEnabled {
       eTag = [MS_USER_DEFAULTS objectForKey:kMSIdentityETagKey];
     }
     NSString *authToken = [self retrieveAuthToken];
+    MSALAccountId *accountId = (MSALAccountId *_Nonnull)[self retrieveAccount].homeAccountId;
+    NSString *lastAccountId = accountId.identifier;
 
     // Only set the auth token if it is not nil to avoid triggering callbacks.
```
dhei: Update comment

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
dhei: `lastHomeAccountId` naming is a little inconsistent with getter/setter of the auth token. Maybe we just use `homeAccountId` for consistency purpose?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -42,22 +50,39 @@ + (void)resetSharedInstance {
...
+  @synchronized(self) {
+    self.authToken = nil;
+    self.lastHomeAccountId = nil;
+    
     // Don't invoke the delegate while locking; it might be locking too and deadlock ourselves.
     synchronizedDelegates = [self.delegates allObjects];
   }
   for (id<MSAuthTokenContextDelegate> delegate in synchronizedDelegates) {
-    [delegate authTokenContext:self didReceiveAuthToken:authToken];
+    [delegate authTokenContext:self didReceiveAuthToken:nil forNewUser:YES];
```
dhei: Calling `YES` on `forNewUser:` delegate to trigger Push module send a new push token sounds like a confusing and misleading method name. It's very confusing without context of the problem as `forNewUser:YES` tells us it's calling delegate because of a new user, but that's not right.

Consider rename this method to make it less confusing.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
dhei: `lastHomeAccountId` should be `nullable`, right?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
```diff
@@ -11,8 +11,9 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param authTokenContext The auth token context.
  * @param authToken The auth token.
+ * @param isNewUser True if the user has changed, false otherwise.
```
dhei: Per my previous comment (https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1328/files#r260102518), "is New User" != "has User changed".

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -34,15 +34,43 @@ - (void)testSetAuthToken {
 
   // If
   NSString *expectedAuthToken = @"authToken1";
+  NSString *expectedAccountId = @"account1";
+  NSString *expectedAccountId1 = @"account2";
```
dhei: I know it's only a unit test, please consider make the variable name and it's value consistent, not self-contradicting. 😄

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -10,11 +10,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface MSAuthTokenContext : NSObject
 
-/**
- * Auth token.
- */
-@property(nonatomic, nullable) NSString *authToken;
```
jaelim-ms: What's the reason of the change? `atomic` isn't enough?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -9,6 +9,16 @@
 
 @interface MSAuthTokenContext ()
 
+/**
+ * Authorization token cached value.
+ */
+@property(nullable, nonatomic, copy) NSString *authToken;
```
jaelim-ms: Can it be `atomic`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -42,22 +50,49 @@ + (void)resetSharedInstance {
...
 }
 
-- (void)setAuthToken:(NSString *)authToken {
+- (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId {
   NSArray *synchronizedDelegates;
+  BOOL isNewUser = NO;
   @synchronized(self) {
-    _authToken = authToken;
+    self.authToken = authToken;
+    isNewUser = self.homeAccountId == nil || ![self.homeAccountId isEqualToString:accountId];
```
jaelim-ms: Is there a case that both `homeAccountId` and `accountId` are `nil`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
```diff
@@ -6,13 +6,23 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol MSAuthTokenContextDelegate <NSObject>
 
+@optional
```
jaelim-ms: Should it be an optional? Any use case that we only implement one method?

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -94,10 +94,11 @@ - (void)applyEnabledState:(BOOL)isEnabled {
       eTag = [MS_USER_DEFAULTS objectForKey:kMSIdentityETagKey];
     }
     NSString *authToken = [self retrieveAuthToken];
+    NSString *accountId = [self retrieveAccountId];
 
-    // Only set the auth token if it is not nil to avoid triggering callbacks.
-    if (authToken) {
-      [MSAuthTokenContext sharedInstance].authToken = authToken;
+    // Only set the auth token if it and account id are not nil to avoid triggering callbacks.
```
jaelim-ms: Since you are changing the comment,
`if it and account id are` -> `if auth token and account id are`

AppCenterPush/AppCenterPushTests/MSPushTests.m
```diff
@@ -222,14 +222,47 @@ - (void)testSendsPushTokenWhenNewAuthTokenReceived {
...
+  NSString *pushToken = @"ConvertedPushToken";
+  OCMStub([pushMock convertTokenToString:deviceToken]).andReturn(pushToken);
+  __block MSPushLog *log;
+  id channelUnitMock = OCMProtocolMock(@protocol(MSChannelUnitProtocol));
+  id channelGroupMock = OCMProtocolMock(@protocol(MSChannelGroupProtocol));
+  OCMStub([channelGroupMock addChannelUnitWithConfiguration:OCMOCK_ANY]).andReturn(channelUnitMock);
+  OCMStub([channelUnitMock enqueueItem:[OCMArg isKindOfClass:[MSPushLog class]] flags:MSFlagsDefault]).andDo(^(NSInvocation *invocation) {
+    [invocation getArgument:&log atIndex:2];
+  });
+  [[pushMock expect] sendPushToken:pushToken];
```
jaelim-ms: Why not using `OCMExpect`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
dhei: @jaelim-ms `authToken` is changed from a property to our own getter/setter methods.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
dhei: @jaelim-ms `accountId` shouldn't be nil if this setter is called correctly. In all the 3 places  it get called we check for nil before calling, so it's not a problem now.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
dhei: @jaelim-ms MSPush.m only implements the `didUpdateUserWithAuthToken:` one.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: No. `Writable atomic property cannot pair a synthesized setter with a user defined getter`
Plus, it isn't required there: it's accessed only from `@synchronized` block.

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -34,15 +34,44 @@ - (void)testSetAuthToken {
...
 
   // Then
-  XCTAssertEqualObjects(self.sut.authToken, expectedAuthToken);
+  XCTAssertEqualObjects([self.sut authToken], expectedAuthToken);
+  OCMVerify([delegateMock authTokenContext:self.sut didUpdateUserWithAuthToken:expectedAuthToken]);
+  
+  // When
+  [self.sut setAuthToken:expectedAuthToken withAccountId:expectedAccountId];
+  
+  //Then
```
roland-chernov-akvelon: Could you please run ./clang-format-changed-files.sh utility to fix formatting.

MatkovIvan: It looks like you reviewed not updated version. It's already fixed in https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1328/commits/9d43dc302e791419bbf128ed9597e6759e08b7e5

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: The reason of asking this is you can simply the line like below.
`isNewUser = ![self.homeAccountId isEqualToString:accountId];`

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: Overall, it doesn't need to have customized getter/setter. The getter only synchronize the property, there is no custom setter anymore.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
jaelim-ms: My original question was "why do you need to cumstomize getter/setter for the property?".
The getter only synchronize the property, there is no customer setter.
You can keep this line (and remove the property declaration in .m file) and change the property to atomic will do exactly same thing that you changed here.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -27,7 +27,7 @@ - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion {
     _storage = [MSLogDBStorage new];
     if (ingestion) {
       _ingestion = ingestion;
-      _ingestion.authToken = [MSAuthTokenContext sharedInstance].authToken;
+      _ingestion.authToken = [[MSAuthTokenContext sharedInstance] authToken];
```
clpolet: If we want to still use the `authToken` as a property we can declare it as readonly property in the "MSAuthTokenContext.h" but leave it as read/write in "MSAuthTokenContextPrivate.h".

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
dhei: @jaelim-ms unless it's an Objective-C convention, I'd prefer readability over micro-optimization. 😄

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: There is a setter. It's generated, do things like `copy` and is used only as a private setter.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: I'm saying `self.homeAccountId == nil` is redundant. It will be compared with `
[self.homeAccountId isEqualToString:accountId]`

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: It didn’t mean atomic will solve the synchronization block in getter. I didn’t see any reason of getting authToken requires locking the whole instance. What’s why I thought getter can be removed, then authToken property can become an atomic property. authToken seems it can be used as soon as it is set. If it is not correct then it totally makes sense.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -42,22 +50,40 @@ + (void)resetSharedInstance {
...
-    _authToken = authToken;
+    self.authToken = nil;
+    self.homeAccountId = nil;
 
     // Don't invoke the delegate while locking; it might be locking too and deadlock ourselves.
     synchronizedDelegates = [self.delegates allObjects];
   }
-  for (id<MSAuthTokenContextDelegate> delegate in synchronizedDelegates) {
-    [delegate authTokenContext:self didReceiveAuthToken:authToken];
+  [self invokeAuthTokenEvents:synchronizedDelegates withToken:nil isNewUser:YES];
```
dhei: Per my original comment (https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1328#discussion_r260102518), I think it's confusing to call `isNewUser:YES` inside clear auth token method because it's not a new user. Due to this reason, I think we can keep them separate instead of refactoring into the `(void)invokeAuthTokenEvents:(NSArray *)delegates withToken:(NSString *)token isNewUser:(BOOL)newUser` method.

clpolet: Yeah it might not be a new user and I think we don't need to trigger the `authTokenContext:didUpdateUserWithAuthToken` if the current user is already an anonymous user. It's supposed to be cleared out already.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: Seems everyone wants to lock the instance when `authToken` is set. Let's keep it.

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -34,15 +34,45 @@ - (void)testSetAuthToken {
 
   // If
```
clpolet: You should divide into more tests for more readability. This test should stay very basic, no previous token/account then new token/account: both callback called. 

Second test: Already a token/account then new token/account: both callbacks triggered.

Third test: Already a token/account then same token but different account: only didReceiveAuthToken triggered.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegate.h
```diff
@@ -6,13 +6,23 @@ NS_ASSUME_NONNULL_BEGIN
...
+
 /**
  * A callback that is called when an auth token is received.
  *
  * @param authTokenContext The auth token context.
  * @param authToken The auth token.
  */
-- (void)authTokenContext:(MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(NSString *)authToken;
+- (void)authTokenContext:(MSAuthTokenContext *)authTokenContext didReceiveAuthToken:(nullable NSString *)authToken;
```
clpolet: I think we might want to rename a bit these callbacks to reflect the action they are triggered for. Can be `didSetNewAuthToken` and `didSetNewAccountIdWithAuthToken`?

dhei: agree with @clpolet

PR #1329, opened 2019-02-26
[iOS] Sasquatch app doesn't display userId value in Transmission Target tab when it switches to another target

Zongrenfeng:

* [ ] Did you add unit tests?

## Description
In transmission tab, when editing in textfields using virtual keyboard, tap on another target will lead to value to be assigned to the new target.

The original design has below issues:
-- the call to propertyValueChanged() always happened after user switch out of a textfield to another target; so when the code inside the propertyValueChanged() try to get the current target supposed to be update its data with, it will get the one target just switch to, not the original one.
-- bind the call to textfields' resignFirstResponder() in editingDidEndOnExit event, which only happened when user press enter key in keyboard;

Fix:
-- make cell will note down current target when any key/value textfield control got focus;
-- in propertyValueChanged(), first compare the noted target with current target, if different, change to use original one;

### DIFF
--- a/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
@@ -3,11 +3,10 @@ import UIKit
 @objc(MSAnalyticsPropertyTableViewCell) class MSAnalyticsPropertyTableViewCell: UITableViewCell {
   @IBOutlet weak var keyField: UITextField!
   @IBOutlet weak var valueField: UITextField!
+  var currentTarget: String = ""
 
   @IBAction func dismissKeyboard(_ sender: UITextField!) {
     sender.resignFirstResponder()
   }
-    
-  // MARK:variable
-  var curTarget: String = ""
+
 }
--- a/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
@@ -7,4 +7,7 @@ import UIKit
   @IBAction func dismissKeyboard(_ sender: UITextField!) {
     sender.resignFirstResponder()
   }
+    
+    // MARK:variable
+    var curTarget: String = "Child 1"
 }
--- a/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
@@ -8,6 +8,6 @@ import UIKit
     sender.resignFirstResponder()
   }
     
-    // MARK:variable
-    var curTarget: String = "Child 1"
+  // MARK:variable
+  var curTarget: String = ""
 }
--- a/Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
@@ -61,19 +61,17 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
   }
   
   // Note down current target for using in End edit events
-  override func recordCurrentTarget(sender: UITextField!){
+  override func recordCurrentTarget(sender: UITextField!) {
     let curCell = sender.superview!.superview as! MSAnalyticsPropertyTableViewCell
     curCell.curTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget() ?? ""
-    }
+  }
     
-  // Add below method since tap out of the key textField will crash the app, that's a bad experience
-  override func propertyKeyChanged(sender: UITextField!){    
-    }
+// Add below method since tap out of the key textField will crash the app, that's a bad experience
+  override func propertyKeyChanged(sender: UITextField!) {    
+  }
     
   override func propertyValueChanged(sender: UITextField!) {
-    // when user changed value inside textfield by using virtual keyboard, then click a different target like 'Child 2'
-    // the textfield's events always happen after the switch target event, this will lead to set values to wrong target error
-    // Add below lines to check out whether the target have been changed, if so, change to use original target
+    // Check out whether the target have been changed, if so, change to use original target
     let curCell = sender.superview!.superview as! MSAnalyticsPropertyTableViewCell
     let curTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget()
     let selectedTarget = curCell.curTarget != curTarget ? curCell.curTarget : curTarget
--- a/Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
@@ -59,22 +59,23 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
       return cell
     }
   }
-  
-  // Note down current target for using in End edit events
+
+  // Note down current target for using in End edit events.
   override func recordCurrentTarget(sender: UITextField!) {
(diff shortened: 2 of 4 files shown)


Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
```diff
@@ -7,4 +7,7 @@ import UIKit
   @IBAction func dismissKeyboard(_ sender: UITextField!) {
     sender.resignFirstResponder()
   }
+    
+    // MARK:variable
+    var curTarget: String = "Child 1"
```
jaelim-ms: What's this one? Is it a default selection?

Zongrenfeng:
> Minor comments. BTW there is a fix for macOS for the same issue. Can you also look at this PR and see which one could be better? If we can use same logic, it would be good for maintenance.

I cannot launch the macOS version Sasquatch App on our local mac machine because of App id out of date issue. 

By examine through and compare with two version's code, I found that the text fields box with this issue are initialized by different way: 
- in macOS version, the text fields objects are directly created in TransmissionViewController.swift;
- but in iOS version, they are created dynamically so that have to be bind with event handling code at run time -- which limited the events we can utilized, and objects the code can see.

So I think that fixed two app by different ways maybe reasonable. Feel free to correct me if I am wrong. Thanks~

jaelim-ms:
@Zongrenfeng Thanks for the answer.

Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
Zongrenfeng: At first, i was expected that the target retrive from code will be a name like "Child 1" or "Child 2", so i set "Child 1" as default, but later found what i got is a Id...

Missed clean up it, will remove from new commit. Thanks~

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -60,8 +60,24 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
     }
   }
   
+  // Note down current target for using in End edit events
+  override func recordCurrentTarget(sender: UITextField!){
```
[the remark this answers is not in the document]
Zongrenfeng: fixed in new commit.

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -60,8 +60,24 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
...
+    let curCell = sender.superview!.superview as! MSAnalyticsPropertyTableViewCell
+    curCell.curTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget() ?? ""
+    }
+    
+  // Add below method since tap out of the key textField will crash the app, that's a bad experience
+  override func propertyKeyChanged(sender: UITextField!){
```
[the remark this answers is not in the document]
Zongrenfeng: fixed in new commit.

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -60,8 +60,22 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
     }
   }
   
+  // Note down current target for using in End edit events
+  override func recordCurrentTarget(sender: UITextField!) {
+    let curCell = sender.superview!.superview as! MSAnalyticsPropertyTableViewCell
+    curCell.curTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget() ?? ""
+  }
+    
+// Add below method since tap out of the key textField will crash the app, that's a bad experience
```
MatkovIvan: Please format the code.

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -60,8 +60,22 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
     }
   }
   
+  // Note down current target for using in End edit events
```
clpolet: Another nit: Please add `.` at the end of your sentences in comments.

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
```diff
@@ -60,8 +60,22 @@ class CommonSchemaPropertiesTableSection : SimplePropertiesTableSection {
...
+    curCell.curTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget() ?? ""
+  }
+    
+// Add below method since tap out of the key textField will crash the app, that's a bad experience
+  override func propertyKeyChanged(sender: UITextField!) {    
+  }
+    
   override func propertyValueChanged(sender: UITextField!) {
-    let selectedTarget = transmissionTargetSelectorCell?.selectedTransmissionTarget()
+    // Check out whether the target have been changed, if so, change to use original target
```
clpolet: You need to add a new line before a comment.

Sasquatch/Sasquatch/ViewControllers/Sections/SimplePropertiesTableSection.swift
```diff
@@ -15,8 +15,10 @@ class SimplePropertiesTableSection : PropertiesTableSection {
 
     // Set cell to respond to being edited.
     cell.keyField.addTarget(self, action: #selector(propertyKeyChanged), for: .editingDidEnd)
+    cell.keyField.addTarget(self, action: #selector(recordCurrentTarget), for: .editingDidBegin)
+
```
clpolet: Can remove this line, we usually have empty lines within code blocks only before a comment.

Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
```diff
@@ -7,4 +7,7 @@ import UIKit
   @IBAction func dismissKeyboard(_ sender: UITextField!) {
     sender.resignFirstResponder()
   }
+    
+  // MARK:variable
```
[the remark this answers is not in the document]
Zongrenfeng: removed comment and moved the code line to above

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
Zongrenfeng: Ok, added the dot

Sasquatch/Sasquatch/ViewControllers/Sections/SimplePropertiesTableSection.swift
Zongrenfeng: removed

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
Zongrenfeng: added a new line

Sasquatch/Sasquatch/ViewControllers/Cells/MSAnalyticsPropertyTableViewCell.swift
```diff
@@ -7,4 +7,7 @@ import UIKit
...
+  var curTarget: String = ""
```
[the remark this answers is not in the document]
Zongrenfeng: changed several names from curTarget -> currentTarget, curCell --> cell

Sasquatch/Sasquatch/ViewControllers/Sections/CommonSchemaPropertiesTableSection.swift
Zongrenfeng: Hum I checked it and made some format changes, please tell me if you found not meet your expectations, thanks~

PR #1330, opened 2019-02-26
[macOS] Sasquatch app doesn't display userId value in Transmission Target tab when it switches to another target

charmian01:
## Description

1. Fix bug 57374 for macOS:

    Before: The  propertyValueChanged() is called when user press enter key to end textfield editing . This      cause the textfield value will be add to another target if you edit textfield and directly change target.

    After: Make propertyValueChanged() is called when textfield happens any change.

2. By the way, fix tiny truncation issue for textfield. 

## Related PRs or issues
https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/57374


jaelim-ms:
I tested it again but I couldn't repro the issue. I assume Xcode was in bad state.

PR #1320, opened 2019-02-21
Use the correct object for @synchonized blocks and make them consistent

roland-chernov-akvelon:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [ ] ~Did you add unit tests?~
* [x] Did you check UI tests on the sample app? They are not executed on CI.
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

This PR:
- Makes sure that the correct object for `@synchonized` blocks is used in services and internal singletons.
- Makes `@synchonized` blocks consistent in `MSService` implementations.

## Misc

- :warning: `[MSAppCenter isEnabled]` static method was left as is.
  Static (publicly accessible) `isEnabled` is different from the instance method and it also checks `canBeUsed` in addition to `isEnabled` (usually we're checking them separately). So technically it's `isEnabledAndCanBeUsed`.
  If we move `canBeUsed` from the static method to the `isEnabled` instance method, we'll have chicken-egg problem: SDK initialization method checks `isEnabled` instance method and `canBeUsed` checks if SDK is initialized.
- `[MSCrashes generateTestCrash]` static method was left as is because it doesn't have corresponding instance method.
- You can hide whitespace changes to simplify review:
screenshot_2019-03-04 use the correct object for synchonized blocks and make them consistent by roland-chernov-akvelon pu

PR #1347, opened 2019-03-04
Use atomic properties where possible and implement addition @synchronized in MSAppCenter

roland-chernov-akvelon:

## Description

This PR:
1. Updates properties referenced in `MSAppCenter` to use `atomic` where possible;
2. Implements additional `@synchronized` in `MSAppCenter`.

## Related PRs or issues

This refactoring was made after discussion in https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1320#issuecomment-468010448 .

### DIFF
--- a/AppCenter/AppCenter/MSAppCenter.m
+++ b/AppCenter/AppCenter/MSAppCenter.m
@@ -413,8 +413,11 @@ static const long kMSMinUpperSizeLimitInBytes = 24 * 1024;
 
 - (void)setLogUrl:(NSString *)logUrl {
   _logUrl = logUrl;
-  if (self.channelGroup) {
-    [self.channelGroup setLogUrl:logUrl];
+
+  // Retain using a local var just in case to ensure it is not set to `nil` in between checking for `nil` & the call to `setLogUrl:`.
+  id<MSChannelGroupProtocol> localChannelGroup = self.channelGroup;
+  if (localChannelGroup) {
+    [localChannelGroup setLogUrl:logUrl];
   }
 }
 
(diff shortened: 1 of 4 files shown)


ElektrojungeMS:
This looks good to me, one question: did we see any perf issues using the `atomic` properties?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -411,12 +407,14 @@ - (BOOL)startService:(Class)clazz
...
+}
+
 - (void)setLogUrl:(NSString *)logUrl {
-  @synchronized(self) {
-    _logUrl = logUrl;
-    if (self.channelGroup) {
-      [self.channelGroup setLogUrl:logUrl];
-    }
+  _logUrl = logUrl;
+  if (self.channelGroup) {
```
achocron: This probably needs to be stored in a local variable and then tested + set since it isn't in a synchronized block

annakocheshkova:
@ElektrojungeMS I don't think this would cause serious performance drop.
F. e., this is `testSetLogUrl` with `@synchronized`:
image
This is `testSetLogUrl` with `atomic`:
image

AppCenter/AppCenter/MSAppCenter.m
MatkovIvan: @ElektrojungeMS @achocron Do we really need this null-check?
From the apple [docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithObjects/WorkingwithObjects.html#//apple_ref/doc/uid/TP40011210-CH4-SW22)
> it’s perfectly acceptable in Objective-C to send a message to nil. If you do send a message to nil, obviously nothing happens.

roland-chernov-akvelon: Thank you. I tested this method with `channelGroup = nil` (without `if` block) and can confirm that it doesn't cause issues.

In addition I verified calling this method from the test app — looks fine.

achocron: Makes sense. Then I think the check is not needed.

PR #1356, opened 2019-03-06
[macOS] Implement Sasquatch test apps to change the log URL for both App Center and One Collector

charmian01:
## Description
[macOS] Implement Sasquatch test apps to change the log URL for both App Center and One Collector:
1. Add a button on App Center tab that displays popup dialog to take a URL for Ingestion endpoint.
2. The dialog  has a TextView with current URL + Reset button + Save button + Cancel button.

## Related PRs or issues
https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/57974

screen shot 2019-03-06 at 5 56 32 pm

### DIFF
--- a/SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
+++ b/SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
@@ -152,13 +152,13 @@ class AppCenterViewController : NSViewController, NSTextFieldDelegate, NSTextVie
     switch(alert.runModal()) {
     case NSAlertFirstButtonReturn:
       UserDefaults.standard.set(prodLogUrl, forKey: kMSLogUrl)
-      MSAppCenter.setLogUrl(prodLogUrl)
+      appCenter.setLogUrl(prodLogUrl)
       break
     case NSAlertSecondButtonReturn:
       let text = textView.string ?? ""
       let logUrl = !text.isEmpty ? text : nil
       UserDefaults.standard.set(logUrl, forKey: kMSLogUrl)
-      MSAppCenter.setLogUrl(logUrl)
+      appCenter.setLogUrl(logUrl)
       break
     case NSAlertThirdButtonReturn:
       break
--- a/SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
+++ b/SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
@@ -9,7 +9,6 @@ class AppCenterViewController : NSViewController, NSTextFieldDelegate, NSTextVie
   var currentAction = AuthenticationViewController.AuthAction.signin
 
   let kMSAppCenterBundleIdentifier = "com.microsoft.appcenter";
-  let prodLogUrl = "https://in.appcenter.ms"
 
   @IBOutlet var installIdLabel : NSTextField?
   @IBOutlet var appSecretLabel : NSTextField?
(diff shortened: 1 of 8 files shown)


SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
```diff
@@ -3,12 +3,13 @@ import Cocoa
...
 let kMSDefaultDatabaseSize = 10 * 1024 * 1024
 
-class AppCenterViewController : NSViewController, NSTextFieldDelegate {
+class AppCenterViewController : NSViewController, NSTextFieldDelegate, NSTextViewDelegate {
 
   var appCenter: AppCenterDelegate = AppCenterProvider.shared().appCenter!
   var currentAction = AuthenticationViewController.AuthAction.signin
 
   let kMSAppCenterBundleIdentifier = "com.microsoft.appcenter";
+  let prodLogUrl = "https://in.appcenter.ms"
```
MatkovIvan: You can use empty value to reset, and then get the current URL from SDK to display.

jaelim-ms: Are you talking about exposing logUrl in `MSAppCenter` class? There is only setter available as a public API.

jaelim-ms:
I'd prefer to realign the button like this.
image

SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
MatkovIvan: Yeah, you are right. This is not a puppet, and this API not available (but due to objective c specifics it can be called 😉).
I just wanted to avoid having this constant here.

charmian01: I use "nil" to set value in UserDefaults when reset, and if restarting app, it will skip setLogUrl call and continue AppCenter start. So the SDK will still use default log url (PROD URL: https://in.appcenter.ms ).
For the UI display, if it don't need to display PROD Log URL,  I use text "Log URL currently set to the production"  to point out current log url is PROD url.  Do you think this is ok?

I pushed a new commit for above, and fix "Set Log URL" button display issue, please help to check.

SasquatchMac/SasquatchMac/ViewControllers/AppCenterViewController.swift
```diff
@@ -133,4 +134,35 @@ class AppCenterViewController : NSViewController, NSTextFieldDelegate {
...
+    let defaultText = "Log URL currently set to the production"
+    textView.string = UserDefaults.standard.string(forKey: kMSLogUrl) ?? defaultText
+    scrollView.documentView = textView
+    scrollView.hasVerticalScroller = true
+    scrollView.contentView.scroll(NSPoint(x: 0, y: textView.frame.size.height))
+    alert.accessoryView = scrollView
+    alert.alertStyle = NSWarningAlertStyle
+    switch(alert.runModal()) {
+    case NSAlertFirstButtonReturn:
+      UserDefaults.standard.set(nil, forKey: kMSLogUrl)
```
jinquan2019: It should also call the setLogUrl right after persist the URL in UserDefaults. Missing calling code.

charmian01: This feature need restart app to work, when restart app, it will call setLogUrl before AppCenter start in AppDetegate. So I think it don't need call setLogUrl after press button.

jinquan2019: Can app refresh the log URL‘s value immediately after changing it？

jaelim-ms: Good point @jinquan2019. We have two scenarios to test `setLogUrl` that is called before `start` and after `start`. Calling `setLogUrl` when an endpoint is changed will cover after `start` scenario.

charmian01: Thanks for above point, I missed the scenario: Calling `setLogUrl `when an endpoint is changed will cover after `start`. Added it in the new commit.

jinquan2019:
When I test the test app on macOS, I found a new bug:
1. Start app with App center mode and save a log URL
2. change startup mode to one collector and restart the app
3. The log url will be the saved log url for app center mode. 
I think that it doesn't make sense.
So I suggest that we need to remove the logurl key from user default when the startup mode is changed. I will send a commit for it.

jaelim-ms:
@jinquan2019 That's a very good point!

PR #1366, opened 2019-03-08
Feature/add set log url button for sasquatch test apps

JasoWu:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [ ] Did you add unit tests?
* [x] Did you check UI tests on the sample app? They are not executed on CI.
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?
## Description

Add new button 'Set Log Url' in Sasquatch test app's AppCenter Settings page.

PR #1372, opened 2019-03-11
[iOS]Add set log url button for test app

jinquan2019:

* [x] Did you add unit tests?

## Description

1. Add and update unit tests for testing default log url
2. Add set log url button for iOS test app
(close the previous PR: https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1366 because it is based on develop)
3. Reset the log url after change the startup mode and restart the app

## Related PRs or issues

Reopen from PR https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1366

image
image

### DIFF
--- a/AppCenter/AppCenter/MSAppCenter.m
+++ b/AppCenter/AppCenter/MSAppCenter.m
@@ -573,12 +573,12 @@ static const long kMSMinUpperSizeLimitInBytes = 24 * 1024;
   @synchronized(self) {
 
     // Construct channel group.
-    if (self.appSecret) {
-      self.oneCollectorChannelDelegate =
-          self.oneCollectorChannelDelegate ?: [[MSOneCollectorChannelDelegate alloc] initWithInstallId:self.installId baseUrl:nil];
-    } else {
+    if (!self.appSecret && self.logUrl != kMSAppCenterBaseUrl) {
       self.oneCollectorChannelDelegate =
           self.oneCollectorChannelDelegate ?: [[MSOneCollectorChannelDelegate alloc] initWithInstallId:self.installId baseUrl:self.logUrl];
+    } else {
+      self.oneCollectorChannelDelegate =
+          self.oneCollectorChannelDelegate ?: [[MSOneCollectorChannelDelegate alloc] initWithInstallId:self.installId baseUrl:nil];
     }
     if (!self.channelGroup) {
       self.channelGroup = [[MSChannelGroupDefault alloc] initWithInstallId:self.installId logUrl:self.logUrl];
--- a/AppCenter/AppCenter/MSAppCenter.m
+++ b/AppCenter/AppCenter/MSAppCenter.m
@@ -224,7 +224,6 @@ static const long kMSMinUpperSizeLimitInBytes = 24 * 1024;
 - (instancetype)init {
   if ((self = [super init])) {
     _services = [NSMutableArray new];
-    _logUrl = kMSAppCenterBaseUrl;
     _enabledStateUpdating = NO;
   }
   return self;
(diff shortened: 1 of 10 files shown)


Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -51,7 +56,7 @@ class MSMainViewController: UITableViewController, AppCenterProtocol, CLLocation
     self.startupModePicker = MSEnumPicker<StartupMode>(
       textField: self.startupModeField,
       allValues: StartupMode.allValues,
-      onChange: {(index) in UserDefaults.standard.set(index, forKey: kMSStartTargetKey)})
+      onChange: {(index) in UserDefaults.standard.set(index, forKey: kMSStartTargetKey); UserDefaults.standard.removeObject(forKey: kMSLogUrl)})
```
MatkovIvan: Please format

Sasquatch/Sasquatch/Base.lproj/Main.storyboard
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="12121" systemVersion="17G65" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="5VQ-Mp-xic">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14460.31" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="5VQ-Mp-xic">
```
MatkovIvan: Does it work on XCode 8?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -577,12 +577,12 @@ - (void)initializeChannelGroup {
   @synchronized(self) {
 
     // Construct channel group.
-    if (self.appSecret) {
+    if (!self.appSecret && self.logUrl != kMSAppCenterBaseUrl) {
```
jaelim-ms: Why do you check `kMSAppCenterBaseUrl`? What if the log URL is still for App Center but not PROD?

jinquan2019: App Center will always have the appSecret, so it will not execute this statement when the log url is for app center. 
When it initialize the app center instance, the log URL will always set as `kMSAppCenterBaseUrl`. 
If we only check the appSecret and don't check the `kMSAppCenterBaseUrl`, the default log URL will be changed to `kMSAppCenterBaseUrl` when app startups with one collector mode in original code.
I think it doesn't mase sence

Sasquatch/Sasquatch/Base.lproj/Main.storyboard
jinquan2019: We only have the XCode 10, not test it on XCode 8. Perhaps I can revert this change.

annakocheshkova:
Verify works with **XCode 8** + **iOS 10.3.3**.

AppCenter/AppCenter/MSAppCenter.m
jaelim-ms: So, this is the scenario.
1. Call `setLogUrl` with valid App Center endpoint (not PROD one).
2. Start App Center.
In this case, we couldn't prevent the App Center custom endpoint being set to One Collector endpoint. So, it isn't necessary since it doesn't cover all scenarios. What do you think?

guperrot: Remove this extra check as discussed with @jaelim-ms

PR #1382, opened 2019-03-15
Add button to update the appsecret for macos test apps

jinquan2019:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [ ] Are tests passing locally?
* [ ] Are the files formatted correctly?
* [ ] Did you add unit tests?
* [x] Did you check UI tests on the sample app? They are not executed on CI.
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Add a button to update the app secret for macOS test apps
1. Need to restart the app after updating the app secret.
2. Set app secret button will be disabled when app startups with one collector mode.

App Center mode:
image

One collector mode:
image

PR #1395, opened 2019-03-18
Add button to update app secret for iOS test apps

jinquan2019:

* [ ] Did you add unit tests?

## Description

Add a button to update the app secret for iOS test apps, need to restart the app after updating the app secret.

App center mode
image

One collector mode
image


Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -92,7 +93,7 @@ class MSMainViewController: UITableViewController, AppCenterProtocol, CLLocation
 
     // Miscellaneous section.
     self.installId.text = appCenter.installId()
-    self.appSecret.text = appCenter.appSecret()
+    self.appSecret.text = prodAppSecret()
```
t-nafle: Similar to #1382 did we want to keep it as appCenter.appSecret() or is prodAppSecret() really necessary?

Sasquatch/SasquatchSwift/AppCenterDelegateSwift.swift
```diff
@@ -39,7 +39,7 @@ class AppCenterDelegateSwift: AppCenterDelegate {
   func appSecret() -> String {
     // TODO: Uncomment when appSecret is moved from internal to public
```
jaelim-ms: Remove TODO.

Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -232,6 +259,10 @@ class MSMainViewController: UITableViewController, AppCenterProtocol, CLLocation
   func prodLogUrl() -> String {
     return startUpModeForCurrentSession == 1 ? ocProdLogUrl : acProdLogUrl
   }
+  
+  func prodAppSecret() -> String {
```
jaelim-ms: This isn't getting PROD app secret. The func name can jsut be `appSecret`.

jaelim-ms: However I think you can just remove this func.

Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
jaelim-ms: Revert your change. One Collector doesn't have app secret concept, you can just leave the app secret there.

Sasquatch/Sasquatch/AppCenterDelegate.swift
```diff
@@ -16,7 +16,6 @@ import AppCenterAnalytics
   func setCustomProperties(_ customProperties: MSCustomProperties)
   func installId() -> String
   func appSecret() -> String
-  func logUrl() -> String
```
ElektrojungeMS: Is this change intentionally part of the PR?

Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -109,6 +110,7 @@ class MSMainViewController: UITableViewController, AppCenterProtocol, CLLocation
 
   override func viewWillAppear(_ animated: Bool) {
     super.viewWillAppear(animated)
+    self.setAppSecretButton.isEnabled = startUpModeForCurrentSession == 1 ? false : true
```
ElektrojungeMS: `1` is a magic number. Can we refactor this to `startUpModeForCurrentSession == startUpMode$WHATEVEROURDESIREDVALUEUS ? false : true`?

clpolet: Yes we should have a constant or enum if multiple values to explicitly define the startUpModes because just 1 is not really explicit.

Sasquatch/Sasquatch/AppCenterDelegate.swift
jinquan2019: No use in the test app. We don't read the log url from SDK in the test app now.

PR #1396, opened 2019-03-18
Add license for storage files

jinquan2019:
1. Merge the feature/identity branch
2. Add license for storage files


jaelim-ms:
All looks good but seems files in `CrashLib` still have an old format. You can submit a separate PR for that change.

PR #1397, opened 2019-03-18
Add history support for auth tokens

roland-chernov-akvelon:

* [x] Did you add unit tests?

## Description

This PR implements `MSAuthTokenStorage` similar to Android and adds history support for Identity auth tokens. Tokens history will be used in order to send logs to ingestion with the correct user.

## Misc

Token removal will be implemented in #1418.


AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -36,37 +36,44 @@ - (MSAuthTokenInfo *)oldestAuthToken {
...
 
-  if (authToken) {
-    BOOL success = [MSKeychainUtil storeString:(NSString *)authToken forKey:kMSIdentityAuthTokenKey];
-    if (success) {
-      MSLogDebug([MSIdentity logTag], @"Saved auth token in keychain.");
+  /*
+ TODO:
+ 1. [ ] read history
+ 2. [x] deserialize (lazy cache?)
+ 3. [ ] don't add duplicates to story
```
MatkovIvan: Remove this point.
Revert
> if there is no history at all, add null entry

instead

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.h
```diff
@@ -7,15 +7,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MSAuthTokenInfo : NSObject
 
-@property(nonatomic, nullable,  copy, readonly) NSString *authToken;
+@property(nonatomic, nullable, copy, readonly) NSString *authToken;
 
-@property(nonatomic, nullable, readonly) NSDate *startTime;
-
-@property(nonatomic, nullable, readonly) NSDate *endTime;
+@property(nonatomic, readonly) int timestamp;
```
MatkovIvan: This class was planned for returning result and requires 2 time-fields. Please revert

AppCenterIdentity/AppCenterIdentity/Internals/Util/MSIdentityConstants.h
```diff
@@ -31,6 +31,16 @@ static NSString *const kMSIdentityETagKey = @"MSIdentityETagKey";
  */
 static NSString *const kMSIdentityAuthTokenKey = @"MSIdentityAuthToken";
 
+/**
+ *The key for Identity auth token array stored in keychain.
```
MatkovIvan:
```suggestion
 * The key for Identity auth token array stored in the keychain.
```

AppCenterIdentity/AppCenterIdentity/Internals/Util/MSIdentityConstants.h
```diff
@@ -31,6 +31,16 @@ static NSString *const kMSIdentityETagKey = @"MSIdentityETagKey";
...
+ */
+static NSString *const kMSIdentityAuthTokenArrayKey = @"MSIdentityAuthTokenArray";
+
+/**
+ *Maximum amount of avaliable token stored in keychain.
```
MatkovIvan:
```suggestion
 * Maximum amount of available token stored in the keychain.
```

AppCenterIdentity/AppCenterIdentity/Internals/Util/MSIdentityConstants.h
```diff
@@ -31,6 +31,16 @@ static NSString *const kMSIdentityETagKey = @"MSIdentityETagKey";
...
+ */
+static int const kMSIdentityMaxAuthTokenArraySize = 10;
```
MatkovIvan:
```suggestion
static int const kMSIdentityMaxAuthTokenArraySize = 5;
```

AppCenterIdentity/AppCenterIdentityTests/MSIdentityTests.m
```diff
@@ -51,6 +49,7 @@ @implementation MSIdentityTests
 
 - (void)setUp {
   [super setUp];
+  [MSAuthTokenContext sharedInstance].init;
```
MatkovIvan: What is this for?

AppCenterIdentity/AppCenterIdentityTests/MSKeychainAuthTokenStorageTests.m
```diff
@@ -0,0 +1,83 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "MSIdentityConstants.h"
+#import "MSKeychainAuthTokenStorage.h"
+#import "MSMockKeychainUtil.h"
+#import "MSMockUserDefaults.h"
+#import "MSTestFrameworks.h"
+#import "MSUtility+File.h"
+#import <OCMock/OCMock.h>
```
MatkovIvan:
```suggestion
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -47,7 +58,12 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Set current auth token and account id.
  */
-- (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId;
+- (void)setAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn;
```
annakocheshkova: Please add comment about parameters, I would write this is a little bit more details, especially since we have new parameter here

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.m
```diff
@@ -0,0 +1,42 @@
...
+
+#import "MSAuthTokenInfo.h"
+
+static NSString *const kMSAuthTokenKey = @"authTokenKey";
+static NSString *const kMSStartTimeKey = @"startTimeKey";
+static NSString *const kMSEndTimeKey = @"endTimeKey";
+
+@implementation MSAuthTokenInfo
+
+- (instancetype)initWithAuthToken:(nullable NSString *)authToken
```
annakocheshkova: So it all can be `nil`? 🤔

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+}
+
+- (nullable NSString *)retrieveAccountId {
+  return [MS_USER_DEFAULTS objectForKey:kMSIdentityMSALAccountHomeAccountKey];
+}
+
+- (MSAuthTokenInfo *)oldestAuthToken {
+
+  // Read token array from storage.
+  NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
```
annakocheshkova: tokenArray can be nil

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+  }
+  return authTokenInfo;
+}
+
+- (void)saveAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
+    if (tokenArray.count == 0) {
```
annakocheshkova: tokenArray can be `nil`

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+    if (![tokenArray.lastObject.authToken isEqual:authToken]) {
+
+      // If new token differs from the last token of array - add it to array.
+      MSAuthTokenInfo *newAuthToken = [[MSAuthTokenInfo alloc] initWithAuthToken:authToken andStartTime:[NSDate date] andEndTime:expiresOn];
+      [tokenArray addObject:newAuthToken];
+    }
+
+    // Cap array size at max available size const (deleting from beginning).
+    if ([tokenArray count] > kMSIdentityMaxAuthTokenArraySize) {
+      [tokenArray removeObjectAtIndex:0];
```
annakocheshkova: Can you explain why at index 0? Have you tested how it works in iOS? I'm just not sure if the array items will move or the position will stay empty.. Maybe we should store oldest tokens at the end of array?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+}
+
+- (void)removeAuthToken:(nullable NSString *)authToken {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
+    for (NSUInteger i = 0; i < tokenArray.count; i++) {
```
annakocheshkova: again, will it fail if tokenArray is nil?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+
+  // Read token array from storage.
+  NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
+  if (tokenArray.count == 0) {
+    return nil;
+  }
+
+  MSAuthTokenInfo *authTokenInfo = tokenArray.firstObject;
+  NSDate *nextChangeTime = tokenArray.count > 1 ? tokenArray[1].startTime : nil;
+  if ([authTokenInfo.endTime laterDate:nextChangeTime]) {
```
annakocheshkova: Will this fail if `endTime` is `nil`?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+
+- (void)saveAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
+    if (tokenArray.count == 0) {
+
+      // Add nil token if the entire story is empty.
+      [tokenArray addObject:[MSAuthTokenInfo new]];
```
annakocheshkova: Can you explain this? Will there just always be nil token at the first place?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,116 @@
...
+    if ([tokenArray count] > kMSIdentityMaxAuthTokenArraySize) {
+      [tokenArray removeObjectAtIndex:0];
+    }
+
+    // Save new array.
+    [MSKeychainUtil storeArray:tokenArray forKey:kMSIdentityAuthTokenArrayKey];
+    if (authToken) {
+      [self saveTokenToKeychain:authToken];
+    } else {
+      [self deleteTokenFromKeychain];
```
annakocheshkova: I'm not completely get it, so if `authToken` is `nil` it means we should delete it, but why we add it to token history array then (as `nil`)?

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -332,22 +306,24 @@ - (MSIdentityConfig *)deserializeData:(NSData *)data {
 }
 
 - (BOOL)clearAuthData {
+  BOOL result = YES;
+  if (![self removeAccount]) {
+    MSLogWarning([MSIdentity logTag], @"Couldn't remove account data.");
+    result = NO;
```
annakocheshkova: I would suggest using two flags here and `return accountSuccessfullyDeleted && authTokenSuccessfullyCleared;` at the end. But that's up to you - just would look more clear and we could assign the reult of the call directly to the flag.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.m
MatkovIvan: Yes

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
MatkovIvan: No

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
Jamminroot: Filling gaps works as subtracting 1 from every index beyond (according to apple's [dev portal](https://developer.apple.com/documentation/foundation/nsmutablearray/1416616-removeobjectatindex?language=objc)) so no, they will not stay empty.
As for why at 0 - we need to delete oldest entry to maintain array size, and since those are in the beginning (newest are written to the end), we delete element at 0.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
Jamminroot: We have to store history state when we were not authorized at all (```nil``` value)

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
roland-chernov-akvelon: Yes, right. We're adding a `nil` token if it's a first element in the history (only occurs once during the first initialization of history). This will be used later for querying data.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
roland-chernov-akvelon: I'm not sure about using 2 flags. Because in that case these flags will be used only once in the return statement and won't be used separately. For me it looks not so much different from the current implementation. I would keep it as is. Thank you.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,130 @@
...
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
+
+    // Do nothing if there's just one entry in the history or no history at all.
+    if (!tokenArray || tokenArray.count == 1) {
+      return;
+    }
+
+    // Find, delete the oldest entry. Do not delete the most recent entry.
+    for (NSUInteger i = 0; i < tokenArray.count - 1; i++) {
```
MatkovIvan: Please allow only the oldest token to be removed.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,130 @@
...
+- (void)deleteTokenFromKeychain {
+  if ([MSKeychainUtil deleteStringForKey:kMSIdentityAuthTokenKey]) {
+    MSLogDebug([MSIdentity logTag], @"Deleted auth token from keychain.");
+  } else {
+    MSLogWarning([MSIdentity logTag], @"Failed to delete auth token from keychain.");
+  };
+}
+
+- (void)saveTokenToKeychain:(nullable NSString *)authToken {
+  if ([MSKeychainUtil storeString:(NSString *)authToken forKey:kMSIdentityAuthTokenKey]) {
```
MatkovIvan: Remove `kMSIdentityAuthTokenKey` and use the last token from history

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Time and date from which the token began to act.
+ */
+@property(nonatomic, nullable, readonly) NSDate *startTime;
+
+/**
+ * Time and date to which token acted.
+ */
+@property(nonatomic, nullable, readonly) NSDate *endTime;
```
roland-chernov-akvelon: Add a comment on the `endTime` property to clarify that in the keychain it's used to store the expiration date.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
roland-chernov-akvelon: Sure. But I'd rather add this check to the PR #1418 about tokens deletion, not into this PR. In this PR `removeAuthToken` method is unused and we only changed tokens storing logic. I think it would be better to apply our changes iteratively and finish the implementation of tokens removal as part of the separate story. I'll add a `TODO` so we don't forget to implement this.

I think we shouldn't have implemented tokens removal as part of this PR because it makes PR too huge and hard to review. But since we already implemented some removal logic here then I'd better keep it as is and fix in the next PR.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
roland-chernov-akvelon: Removed, thanks. There are 9 tests failing after this change. We will fix them and remove the unused constant a bit later.

AppCenter/AppCenter/Internals/Storage/MSAuthTokenStorage.h
```diff
@@ -26,10 +26,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * @return auth token info.
  */
 - (MSAuthTokenInfo *)oldestAuthToken;
 
+/**
+ * Returns auth token info for the latest entity in the the history.
+ *
+ * @return auth token info.
+ */
+- (MSAuthTokenInfo *)latestAuthToken;
```
MatkovIvan: Remove it from protocol at least

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -12,7 +12,8 @@
 @implementation MSKeychainAuthTokenStorage
 
 - (nullable NSString *)retrieveAuthToken {
-  MSAuthTokenInfo *latestAuthTokenInfo = [self latestAuthToken];
+  NSMutableArray<MSAuthTokenInfo *> *authTokensHistory = [self authTokensHistoryState];
+  MSAuthTokenInfo *latestAuthTokenInfo = authTokensHistory.firstObject;
```
MatkovIvan: last, not first

AppCenter/AppCenterTests/MSAuthTokenInfoTests.m
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAuthTokenInfo.h"
+#import "MSKeychainUtil.h"
+#import "MSKeychainUtilPrivate.h"
+#import "MSTestFrameworks.h"
+
+@interface MSAuthTokenInfoTests : XCTestCase
+@property(nonatomic) id keychainUtilMock;
```
MatkovIvan:
```suggestion

@property(nonatomic) id keychainUtilMock;
```

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,128 @@
...
+  if (authTokensHistory) {
+    MSLogDebug([MSIdentity logTag], @"Retrieved history state from the keychain.");
+  } else {
+    MSLogWarning([MSIdentity logTag], @"Failed to retrieve history state from the keychain or none was found.");
+    authTokensHistory = [NSMutableArray<MSAuthTokenInfo *> new];
+  }
+  if (authTokensHistory.count == 0) {
+
+    // Add nil token if the entire story is empty.
+    [authTokensHistory addObject:[MSAuthTokenInfo new]];
```
MatkovIvan: Add it only on save operation

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // Save new array after changes.
+    [self storeAuthTokensHistoryState:tokenArray];
+  }
+   */
+}
+
+- (NSMutableArray<MSAuthTokenInfo *> *)authTokensHistoryState {
+  NSMutableArray<MSAuthTokenInfo *> *authTokensHistory = [MSKeychainUtil arrayForKey:kMSIdentityAuthTokenArrayKey];
```
MatkovIvan: Cache it to avoid deserialization each time

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,128 @@
...
+}
+
+- (void)saveAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *authTokensHistory = [self authTokensHistoryState];
+    NSString *latestAuthToken = [authTokensHistory lastObject].authToken;
+    if ((latestAuthToken != authToken) &&
+        !(latestAuthToken && authToken && [latestAuthToken isEqualToString:(NSString * _Nonnull) authToken])) {
```
MatkovIvan:
```suggestion
        ![latestAuthToken isEqualToString:authToken]) {
```

roland-chernov-akvelon: Unfortunately, your suggestion doesn't work. There are 2 issues since both `latestAuthToken` and `authToken` are nullables:
1. We also need to check that `latestAuthToken != nil` to avoid situations like this: `[nil isEqualToString:nil] == NO`.
2. Compilation fails if we're passing nullable argument to `isEqualToString`: `Error:(52, 61) implicit conversion from nullable pointer 'NSString * _Nullable' to non-nullable pointer type 'NSString * _Nonnull'`.

But I agree that this check isn't really clear so I did some refactoring and simplified it using a ternary operator. I hope it looks better now. Thanks.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
roland-chernov-akvelon: Sure. I've implemented it using property getter/setter. Please let me know if it looks good to you. Thanks.

roland-chernov-akvelon:
@MatkovIvan Sure, added.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -20,7 +21,7 @@ @interface MSAuthTokenContext ()
 /**
  * The last value of user account id.
  */
-@property(nullable, nonatomic, copy) NSString *homeAccountId;
+@property(nullable, atomic, copy) NSString *accountId;
```
guperrot: Let's not rename if homeAccountId and accountId are different strings.

The Android client keeps the 2 terms distinct (accountId is only the guid, while homeaccountid includes tenant and policy). In Android MSAL get account requires homeAccountId while we expose only accountId in signIn method result.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.m
jaelim-ms: This covers anonymous when it is `nil`. We should know from the history if a user sign-out.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -20,7 +21,7 @@ @interface MSAuthTokenContext ()
 /**
  * The last value of user account id.
  */
-@property(nullable, nonatomic, copy) NSString *homeAccountId;
+@property(nullable, atomic, copy) NSString *homeAccountId;
```
jaelim-ms: You moved this to header, shouldn't it be removed?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -112,4 +102,11 @@ - (void)removeDelegate:(id<MSAuthTokenContextDelegate>)delegate {
   }
 }
 
+- (void)cacheAuthToken {
```
jaelim-ms: What's the purpose of this method? Seems it contains current `authToken` and `homeAccountId` that can be handled in `setAuthToken` I guess.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAuthTokenStorage.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSKeychainAuthTokenStorage : NSObject <MSAuthTokenStorage>
+
+@property(nullable, nonatomic) NSMutableArray<MSAuthTokenInfo *> *authTokensHistoryState;
```
jaelim-ms: Add a doc for this property. Why does the name contain `State`?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,137 @@
...
+
+- (void)saveAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *authTokensHistory = [self authTokensHistoryState];
+    if (authTokensHistory.count == 0) {
+
+      /*
+       * Adding a nil entry is required during the first initialization to differentiate
```
jaelim-ms: If we need to add a `nil` entry to history, it can be done by core (like session history). I still don't see any reason of doing this one because anonymous usage could be covered if it isn't in the history. Am I missing anything here?

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+
+    // Save new array after changes.
+    [self storeAuthTokensHistoryState:tokenArray];
+  }
+   */
+}
+
+- (NSMutableArray<MSAuthTokenInfo *> *)authTokensHistoryState {
```
jaelim-ms: Why do you return mutable array? Do you want it to be changed outside of *MSKeychainAuthTokenStorage.m*?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: No, it's readonly in header

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
MatkovIvan: Disagree.

> because anonymous usage could be covered if it isn't in the history

No, it can't.

> Am I missing anything here?

Probably the history size limit. We shouldn't send logs related to deleted token as anonymous.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: Yes, it can be replaced by `setAuthToken`.
The purposes:
- Harmonize with Android
- wrap almost operations with token storage into context

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.h
roland-chernov-akvelon: > Add a doc for this property.
Sure, added.

> Why does the name contain State?
It seems `State` suffix is too confusing and actually doesn't reflect the purpose of the property so I removed it. Thanks.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
roland-chernov-akvelon: Agree, it's better to make it immutable. Fixed, thanks.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: Ah.. OK.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
jaelim-ms: Well. I still think `nil` should be added at the time the service starts. This is a valid data in between the SDK start and sign-in especially for crashes.
BTW, are we dropping logs that cannot be found an identity for the logs? I think we still send it as anonymous. Please double check.

AppCenter/AppCenter/Internals/Util/MSKeychainUtil.m
```diff
@@ -17,6 +17,33 @@ @implementation MSKeychainUtil
   return serviceName;
 }
 
++ (BOOL)storeArray:(NSArray *)array forKey:(NSString *)key {
```
clpolet: I don't think we should store arrays of objects in the keychain. One token is OK but for multiple tokens I would do something similar to what we are already doing for 1DS tokens: Use the MSEncrypter to encrypt the token and store it. 1DS tokens are stored with logs in the DB but auth tokens history can be stored in the user defaults with the actual token being encrypted by the encrypter of course.

roland-chernov-akvelon: I just wonder if there are any limitations on the keychain item size or if there are some performance issues. Currently, we're limiting the history by just 5 items.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
Jamminroot: @MatkovIvan why would we have a parameter to search authToken then, if we only need to delete the oldest token? I think it might be good to refactor this to always delete only the oldest auth token, without parameters.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -75,28 +81,12 @@ - (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId {
...
-  }
-  for (id<MSAuthTokenContextDelegate> delegate in synchronizedDelegates) {
-    if ([delegate respondsToSelector:@selector(authTokenContext:didSetNewAuthToken:)]) {
-      [delegate authTokenContext:self didSetNewAuthToken:nil];
-    }
-    if (clearedExistingUser && [delegate respondsToSelector:@selector(authTokenContext:didSetNewAccountIdWithAuthToken:)]) {
-      [delegate authTokenContext:self didSetNewAccountIdWithAuthToken:nil];
     }
   }
+  [self setAuthToken:nil withAccountId:nil expiresOn:nil];
```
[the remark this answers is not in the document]
MatkovIvan: @clpolet there is `synchronized` block inside setter implementation

AppCenter/AppCenter/Internals/Util/MSKeychainUtil.m
MatkovIvan: @clpolet Disagree.

This is a list of tokens that can still be valid. Also, additional information is just some integers.
Plus we have a size limit of history and cache decrypted value. I believe that's not an issue.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
MatkovIvan: Disagree again. The key point here is the start time. In your case logs from previous runs/older SDKs version are out of range. This check fixes it.

MatkovIvan: Regarding dropping logs, it's in the separate ongoing PR

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
MatkovIvan: Already discussed. See the comment for this method.

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
jaelim-ms: I assume it drops logs if it cannot find identity from history, this implementation will drop some of anonymous logs.
Here is a scenario.
1. App launched at 1:00 PM
2. App Crashed at 1:01 PM
3. App launched again at 1:02 PM, at this time previous crashes will be processed. Assume first and second tries failed and it is waiting for the last retry.
4. User sign-in at 1:05 PM
5. App Crashed at 1:06 PM. At this time the history will contain "nil/anonymous" at 1:05 PM and one user identity at 1:05 PM
6. App relaunched at 1:07 PM

The first crash happened at 1:01 PM and it couldn't find any identities in the history so the log will be dropped.

If you add "nil/anonymous" identity to the history at the time Identity starts, the history will contain "nil/anonymous" with the time of 1:00 PM so it will use this anonymous identity (not dropping logs).

jaelim-ms: And putting an anonymous identity when history is empty isn't necessary anymore.

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -92,13 +90,7 @@ - (void)applyEnabledState:(BOOL)isEnabled {
...
       eTag = [MS_USER_DEFAULTS objectForKey:kMSIdentityETagKey];
     }
-    NSString *authToken = [self retrieveAuthToken];
-    NSString *accountId = [self retrieveAccountId];
-
-    // Only set the auth token if auth token and account id are not nil to avoid triggering callbacks.
-    if (authToken && accountId) {
-      [[MSAuthTokenContext sharedInstance] setAuthToken:authToken withAccountId:accountId];
-    }
+    [[MSAuthTokenContext sharedInstance] cacheAuthToken];
```
jaelim-ms: Is it loading history from cache? Please take a look at Device and Session history. It can be loaded at the time AuthTokenContext is initialized.

MatkovIvan: It can't be loaded without `Identity` module

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: The case is the default `readwrite` attribute outside of the header. The setter should stay private

AppCenterIdentity/AppCenterIdentity/Internals/Storage/MSKeychainAuthTokenStorage.m
MatkovIvan: Anyway dropping logs is not a part of this PR. It's not a blocker. It will be fixed in next one.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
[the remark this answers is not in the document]
MatkovIvan: @jaelim-ms It shouldn't place the calls of listeners inside block - it's wrong. But I agree it should be fixed, but NOT by moving this line inside block

AppCenterIdentity/AppCenterIdentity/MSIdentity.m
jaelim-ms: We probably need to think about this one because I'm not sure if it can be loaded without Identity module or we don't actually care. Session and Device histories are also loaded by core and these are not used by Distribute and Push modules. I don't see any reasons of coupling AuthTokenContext with Identity because it can be used without Identity in the future.

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -481,4 +481,32 @@ - (void)testDelegateCalledWhenChannelUnitShouldFilterLog {
   [channelUnitMock stopMocking];
 }
 
+- (void)testUpdatesIngestionAuthTokenWhenAuthTokenIsReceived {
```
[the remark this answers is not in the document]
annakocheshkova: This test was added here a month ago, and I guess is shown as "green" in diff by mistake :) 
I don't think this is related to this PR but we can fix it later in a new PR.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenHistoryState.h
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Represents token data entity.
+ */
+@interface MSAuthTokenHistoryState : NSObject
```
MatkovIvan: Add account id to history

AppCenter/AppCenter/Internals/Storage/MSAuthTokenStorage.h
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+@class MSAuthTokenInfo;
+@class MSAuthTokenHistoryState;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol MSAuthTokenStorage <NSObject>
```
MatkovIvan: Remove this class

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAuthTokenStorage.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSKeychainAuthTokenStorage : NSObject <MSAuthTokenStorage>
```
MatkovIvan: Remove this and place the code directly in context.

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,145 @@
...
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *authTokenHistory = [[self authTokenHistory] mutableCopy];
+    if (authTokenHistory.count == 0) {
+
+      /*
+       * Adding a nil entry is required during the first initialization to differentiate
+       * anonymous usage before the moment and situation when we don't have a token
+       * in history because of the size limit for example.
+       */
+      [authTokenHistory addObject:[MSAuthTokenInfo new]];
```
MatkovIvan: Remove it (as discussed)

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,145 @@
...
+      MSAuthTokenInfo *newAuthToken = [[MSAuthTokenInfo alloc] initWithAuthToken:authToken
+                                                                    andAccountId:accountId
+                                                                    andStartTime:[NSDate date]
+                                                                      andEndTime:expiresOn];
+      [authTokenHistory addObject:newAuthToken];
+    }
+
+    // Cap array size at max available size const (deleting from beginning).
+    if ([authTokenHistory count] > kMSMaxAuthTokenArraySize) {
+      [authTokenHistory removeObjectAtIndex:0];
```
MatkovIvan: Remove `count - max` elements

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,145 @@
...
+       */
+      [authTokenHistory addObject:[MSAuthTokenInfo new]];
+    }
+
+    // If new token differs from the last token of array - add it to array.
+    NSString *latestAuthToken = [authTokenHistory lastObject].authToken;
+    if (latestAuthToken ? ![latestAuthToken isEqualToString:(NSString * _Nonnull) authToken] : authToken != nil) {
+      MSAuthTokenInfo *newAuthToken = [[MSAuthTokenInfo alloc] initWithAuthToken:authToken
+                                                                    andAccountId:accountId
+                                                                    andStartTime:[NSDate date]
```
MatkovIvan: Copy logic from android (based on account id)

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,145 @@
...
+
+    // Cap array size at max available size const (deleting from beginning).
+    if ([authTokenHistory count] > kMSMaxAuthTokenArraySize) {
+      [authTokenHistory removeObjectAtIndex:0];
+    }
+
+    // Save new array.
+    [self setAuthTokenHistory:authTokenHistory];
+    if (authToken && accountId) {
+      [MS_USER_DEFAULTS setObject:(NSString *)accountId forKey:kMSHomeAccountKey];
```
MatkovIvan: Remove it and get account id from history

AppCenter/AppCenter/Internals/Storage/MSKeychainAuthTokenStorage.m
```diff
@@ -0,0 +1,145 @@
...
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [[self authTokenHistory] mutableCopy];
+
+    // Do nothing if there's just one entry in the history or no history at all.
+    if (!tokenArray || tokenArray.count == 1) {
+      return;
+    }
+
+    // Find, delete the oldest entry. Do not delete the most recent entry.
+    for (NSUInteger i = 0; i < tokenArray.count - 1; i++) {
```
MatkovIvan: Only the oldest allowed to be deleted. No need for loop

AppCenter/AppCenter/Internals/Storage/MSLogDBStorage.m
```diff
@@ -211,6 +221,10 @@ - (BOOL)loadLogsWithGroupId:(NSString *)groupId
 
 #pragma mark - Delete logs
 
+- (void)deleteLogsWithDateBefore:(nonnull NSDate *)dateBefore {
```
MatkovIvan: Unused. Remove

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenHistoryState.h
Jamminroot: We have it in history (`MSAuthTokenInfo` class); Even though the name (`MSAuthTokenHistoryState`) stands for history, it is actually a representation of token validity information and account id is not to be placed here. I'd rather suggest renaming class to smth like `MSAuthTokenValidityInfo`

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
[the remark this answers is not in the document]
clpolet: You are right because `setAuthToken:withAccountId:expiresOn` is calling the listener. It will have to be another fix.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -74,41 +82,145 @@ - (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId {
...
+}
+
+- (NSMutableArray<MSAuthTokenValidityInfo *> *)authTokenValidityArray {
+  NSMutableArray<MSAuthTokenInfo *> *__nullable tokenArray =
+      (NSMutableArray<MSAuthTokenInfo *> * __nullable)[MSKeychainUtil arrayForKey:kMSAuthTokenArrayKey];
+  NSMutableArray<MSAuthTokenValidityInfo *> *resultArray = [NSMutableArray<MSAuthTokenValidityInfo *> new];
+  if (!tokenArray || tokenArray.count == 0) {
+    return nil;
+  }
+  for (NSUInteger i = 0; i < tokenArray.count; i++) {
```
MatkovIvan: Add null token as the first element to correctly handle old logs

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.m
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAuthTokenInfo.h"
+#import "MSAuthTokenValidityInfo.h"
+
+static NSString *const kMSAuthTokenKey = @"authTokenKey";
+static NSString *const kMSAccountIdKey = @"accountIdKey";
+static NSString *const kMSStartTimeKey = @"startTimeKey";
+static NSString *const kMSEndTimeKey = @"endTimeKey";
```
MatkovIvan: Rename to expire

AppCenter/AppCenter/Internals/Storage/MSAuthTokenStorage.h
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MatkovIvan: Delete this file

AppCenter/AppCenter/Internals/Storage/MSStorage.h
```diff
@@ -73,9 +73,19 @@ typedef void (^MSLoadDataCompletionHandler)(NSArray<id<MSLog>> *_Nullable logArr
 - (BOOL)loadLogsWithGroupId:(NSString *)groupId
                       limit:(NSUInteger)limit
          excludedTargetKeys:(nullable NSArray<NSString *> *)excludedTargetKeys
+                  afterDate:(nullable NSDate *)dateAfter
                  beforeDate:(nullable NSDate *)date
```
MatkovIvan:
```suggestion
                  beforeDate:(nullable NSDate *)dateBefore
```

AppCenter/AppCenter/MSConstants.h
```diff
@@ -55,6 +55,21 @@ typedef void (^MSLogHandler)(MSLogMessageProvider messageProvider, MSLogLevel lo
 typedef NS_ENUM(NSInteger, MSPriority) { MSPriorityBackground, MSPriorityDefault, MSPriorityHigh };
 static short const kMSPriorityCount = MSPriorityHigh + 1;
 
+/**
+ * The key for Identity auth token array stored in keychain.
+ */
+static NSString *const kMSAuthTokenArrayKey = @"MSAppCenterAuthTokenArray";
```
MatkovIvan: Array -> History

AppCenter/AppCenter/MSConstants.h
```diff
@@ -55,6 +55,21 @@ typedef void (^MSLogHandler)(MSLogMessageProvider messageProvider, MSLogLevel lo
...
+
+/**
+ * Maximum amount of available token stored in the keychain.
+ */
+static int const kMSMaxAuthTokenArraySize = 5;
+
+/**
+ * The key for the account homeAccountId stored in user defaults.
+ */
+static NSString *const kMSHomeAccountKey = @"MSAppCenterHomeAccount";
```
MatkovIvan: Remove

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -33,7 +33,7 @@ - (void)setUp {
                                                               flushInterval:flushInterval
                                                              batchSizeLimit:batchSizeLimit
                                                         pendingBatchesLimit:pendingBatchesLimit];
-  [[MSAuthTokenContext sharedInstance] clearAuthToken];
+  (void)[[MSAuthTokenContext sharedInstance] init];
```
MatkovIvan: WAT. Just remove it

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.h
```diff
@@ -38,21 +45,53 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (NSMutableArray<MSAuthTokenValidityInfo *> *)authTokenValidityArray;
+
+/**
+ * Stores auth token and account ID to settings and keychain.
+ *
+ * @param authToken Auth token.
+ * @param accountId Account identifier.
  */
-- (void)setAuthToken:(NSString *)authToken withAccountId:(NSString *)accountId;
+- (void)saveAuthToken:(nullable NSString *)authToken withAccountId:(nullable NSString *)accountId expiresOn:(nullable NSDate *)expiresOn;
```
MatkovIvan: It should be private

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -33,7 +33,7 @@ - (void)setUp {
                                                               flushInterval:flushInterval
                                                              batchSizeLimit:batchSizeLimit
                                                         pendingBatchesLimit:pendingBatchesLimit];
-  [[MSAuthTokenContext sharedInstance] clearAuthToken];
+  [[MSAuthTokenContext sharedInstance] init];
```
MatkovIvan:
```suggestion
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -13,14 +12,14 @@
 @interface MSAuthTokenContext ()
 
 /**
- * Cached authorization token.
+ * Private field used to get and set auth tokens history array.
```
clpolet: Please move these declarations to a MSAuthTokenContextPrivate.h header

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,62 +52,168 @@ + (void)resetSharedInstance {
...
+    if (lastEntry != nil && [latestAuthToken isEqual:(NSString * _Nonnull) authToken]) {
+      return;
+    }
+    BOOL isNewUser = authTokenHistory.lastObject == nil || ![accountId isEqualToString:(NSString * __nonnull) latestAccountId];
+    NSDate *newTokenStartDate = [NSDate date];
+
+    // If there is a gap between tokens.
+    if (latestTokenEndTime && [newTokenStartDate laterDate:(NSDate * __nonnull) latestTokenEndTime]) {
+
+      // If the account the same or become anonymous.
```
clpolet:
```suggestion
      // If the account is the same or becomes anonymous.
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,62 +52,168 @@ + (void)resetSharedInstance {
...
+  return latestAuthTokenInfo.authToken;
+}
+
+- (nullable NSString *)accountId {
+  NSArray<MSAuthTokenInfo *> *authTokenHistory = [self authTokenHistory];
+  MSAuthTokenInfo *latestAuthTokenInfo = authTokenHistory.lastObject;
+  return latestAuthTokenInfo.accountId;
+}
+
+- (NSMutableArray<MSAuthTokenValidityInfo *> *)authTokenValidityArray {
```
clpolet: It feels like this method would need some comments.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Time and date from which the token began to act.
+ * It can be nil if it applies to all logs before endTime.
+ */
+@property(nonatomic, nullable, readonly) NSDate *startTime;
+
+/**
+ * Time and date to which token acted.
+ * It can be nil if it's still valid.
+ * In the keychain encoded data used to store expiration time.
```
clpolet: Please rephrase this line.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenInfo.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Initialize a token info with required parameters.
+ *
+ * @param authToken Auth token.
+ * @param startTime Start time.
+ * @param expiresOn End time.
+ *
+ * @return Token info instance.
+ */
+- (instancetype)initWithAuthToken:(nullable NSString *)authToken
```
clpolet: no `and` in the API name please. Unless what follows it is an action but there is no actions here.

clpolet: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
Jamminroot: Its out there already, in a header file:
```Objective-C
/*
 * Returns array of auth tokens validity info.
 *
 * @return Array of MSAuthTokenValidityInfo
 */
- (NSMutableArray<MSAuthTokenValidityInfo *> *)authTokenValidityArray;
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,51 +48,75 @@ + (void)resetSharedInstance {
...
-      return NO;
-    } else if (self.homeAccountId) {
-      clearedExistingUser = YES;
+    MSAuthTokenInfo *lastEntry = authTokenHistory.lastObject;
+    NSString *__nullable latestAuthToken = lastEntry.authToken;
+    NSString *__nullable latestAccountId = lastEntry.accountId;
+    NSDate *__nullable latestTokenEndTime = lastEntry.expiresOn;
+
+    // If new token doesn't differ from the last token of array - no need to add it to array.
+    if (lastEntry != nil && [latestAuthToken isEqual:(NSString * _Nonnull) authToken]) {
```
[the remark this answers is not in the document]
jaelim-ms: What if `latestAuthToken` and `authToken` are `nil`? The scenario would be when user sign-out twice.
`[(NSString *)nil isEqual:(NSString *)nil]` returns `NO`.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,51 +48,75 @@ + (void)resetSharedInstance {
...
+      return;
+    }
+    BOOL isNewUser = authTokenHistory.lastObject == nil || ![accountId isEqualToString:(NSString * __nonnull) latestAccountId];
```
jaelim-ms: Reconsider this condition too.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,51 +48,75 @@ + (void)resetSharedInstance {
...
+    NSDate *newTokenStartDate = [NSDate date];
+
+    // If there is a gap between tokens.
+    if (latestTokenEndTime && [newTokenStartDate laterDate:(NSDate * __nonnull) latestTokenEndTime]) {
+
+      // If the account is the same or becomes anonymous.
+      if (!isNewUser || authToken == nil) {
+
+        // Apply the new token to this time.
+        newTokenStartDate = latestTokenEndTime;
```
jaelim-ms: Should it be
```suggestion
        newTokenStartDate = [NSDate new];
```

Start time for the new token is not the end time of previous token.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,51 +48,75 @@ + (void)resetSharedInstance {
...
+    MSAuthTokenInfo *newAuthToken = [[MSAuthTokenInfo alloc] initWithAuthToken:authToken
+                                                                     accountId:accountId
+                                                                     startTime:newTokenStartDate
+                                                                     expiresOn:expiresOn];
+    [authTokenHistory addObject:newAuthToken];
+
+    // Cap array size at max available size const (deleting from beginning).
+    if ([authTokenHistory count] > kMSMaxAuthTokenArraySize) {
+      [authTokenHistory removeObjectsInRange:(NSRange){0, [authTokenHistory count] - kMSMaxAuthTokenArraySize}];
+      MSLogWarning([MSAppCenter logTag], @"Size of the token history is exceeded. The oldest token has been removed.");
```
jaelim-ms: This is an expected behavior and how we maintain history. Why is it *Warning*?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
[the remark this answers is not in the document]
jaelim-ms: I think you should review all conditions for string `isEqual` or `isEqualToString`.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -112,4 +131,89 @@ - (void)removeDelegate:(id<MSAuthTokenContextDelegate>)delegate {
...
+
+- (void)removeAuthToken:(nullable NSString *)authToken {
+  @synchronized(self) {
+
+    // Read token array from storage.
+    NSMutableArray<MSAuthTokenInfo *> *tokenArray = [[self authTokenHistory] mutableCopy];
+
+    // Do nothing if there's just one entry in the history or no history at all.
+    if (!tokenArray || tokenArray.count <= 1) {
+      MSLogWarning([MSAppCenter logTag], @"Couldn't remove token from history; token history is empty or contains only current one.");
```
jaelim-ms: Doesn't seem *Warning*.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -112,4 +131,89 @@ - (void)removeDelegate:(id<MSAuthTokenContextDelegate>)delegate {
...
+      return;
+    }
+
+    // Check oldest entry, delete if it matches.
+    if ([authToken isEqual:tokenArray[0].authToken]) {
```
jaelim-ms: Channel is flushing log sequentially. Why do you need to check the token is oldest here? Or, why do you take `authToken` value to remove tokens? Seems it is complicated to determine whether a token could be safely removed or not, I would suggest not to delete auth token anywhere except when it needs to add a new one on history full.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -112,4 +131,89 @@ - (void)removeDelegate:(id<MSAuthTokenContextDelegate>)delegate {
...
+      [tokenArray removeObjectAtIndex:0];
+    } else {
+      MSLogWarning([MSAppCenter logTag], @"Couldn't remove token from history; the token isn't oldest or is already removed.");
```
jaelim-ms: Is it *Warning* level if the token isn't oldest? We should think about simplifying "remove" logic.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: Discussed offline. Seems correct.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenValidityInfo.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Represents token data entity.
+ */
+@interface MSAuthTokenValidityInfo : NSObject
+
+/**
+ * Auth token string.
+ */
+@property(nonatomic, nullable, copy, readonly) NSString *authToken;
```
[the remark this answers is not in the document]
annakocheshkova: Readonly cause it shouldn't be rewritten after initialized

AppCenter/AppCenter/Internals/Storage/MSStorage.h
```diff
@@ -65,17 +65,28 @@ typedef void (^MSLoadDataCompletionHandler)(NSArray<id<MSLog>> *_Nullable logArr
...
          excludedTargetKeys:(nullable NSArray<NSString *> *)excludedTargetKeys
-                 beforeDate:(nullable NSDate *)date
+                  afterDate:(nullable NSDate *)dateAfter
+                 beforeDate:(nullable NSDate *)dateBefore
           completionHandler:(nullable MSLoadDataCompletionHandler)completionHandler;
 
+/**
+ * Return amount of log entries before date.
+ *
+ * @param date Date to count logs until.
```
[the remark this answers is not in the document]
annakocheshkova: @jaelim-ms sorry, didn't quite get this one, any actions from my side needed?

AppCenter/AppCenter/Internals/Util/MSKeychainUtil.m
annakocheshkova: @jaelim-ms for now, yes, we can refactor this with further pull requests. This one is already big enough and has too many changes :(

AppCenter/AppCenterTests/MSAuthTokenInfoTests.m
```diff
@@ -0,0 +1,62 @@
...
+@property(nonatomic, copy) NSString *acServiceName;
+
+@end
+
+@implementation MSAuthTokenInfoTests
+
+- (void)setUp {
+  [super setUp];
+  self.keychainUtilMock = OCMClassMock([MSKeychainUtil class]);
```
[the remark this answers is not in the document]
annakocheshkova: I guess no, we need custom stubbing logic further in this class.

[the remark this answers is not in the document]
jaelim-ms: Why do you need custom stubbing logic? Mock Keychain util should be working as a real one. Are you changing behavior of Keychain?

AppCenter/AppCenter/Internals/Storage/MSStorage.h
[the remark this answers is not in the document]
jaelim-ms: It actually is on a wrong line. I meant
`Return amount of log entries before date.` => `Return amount of log entries on or before date.`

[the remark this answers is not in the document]
jaelim-ms: It would be better
`Return the number of logs on or before the given date`.

AppCenter/AppCenter/Internals/Storage/MSStorage.h
```diff
@@ -65,17 +65,28 @@ typedef void (^MSLoadDataCompletionHandler)(NSArray<id<MSLog>> *_Nullable logArr
...
 - (BOOL)loadLogsWithGroupId:(NSString *)groupId
                       limit:(NSUInteger)limit
          excludedTargetKeys:(nullable NSArray<NSString *> *)excludedTargetKeys
-                 beforeDate:(nullable NSDate *)date
+                  afterDate:(nullable NSDate *)dateAfter
+                 beforeDate:(nullable NSDate *)dateBefore
           completionHandler:(nullable MSLoadDataCompletionHandler)completionHandler;
 
+/**
+ * Return amount of log entries before date.
```
jaelim-ms:
```suggestion
 * Return the number of logs on or before the given date.
```

AppCenter/AppCenter/Internals/Storage/MSStorage.h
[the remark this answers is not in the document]
jaelim-ms: Suggested that in the correct line.

AppCenter/AppCenterTests/MSAuthTokenInfoTests.m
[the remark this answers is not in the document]
clpolet: It's better to improve our existing shared mock whenever possible I think

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -223,106 +348,8 @@ - (void)flushQueue {
...
-                    }];
-          }
-        }];
-
-  // Flush again if there is another batch to send.
-  if (self.availableBatchFromStorage && !self.pendingBatchQueueFull) {
-    [self flushQueue];
-  }
+  NSMutableArray<MSAuthTokenValidityInfo *> *tokenArray = [MSAuthTokenContext sharedInstance].authTokenValidityArray;
+  [self flushQueueForTokenArray:tokenArray withTokenIndex:0];
```
jaelim-ms: Is it safe if `tokenArray` is empty?

jaelim-ms: Is it safe if `tokenArray` is empty?

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -223,106 +348,8 @@ - (void)flushQueue {
...
-                      });
-                    }];
-          }
-        }];
-
-  // Flush again if there is another batch to send.
-  if (self.availableBatchFromStorage && !self.pendingBatchQueueFull) {
-    [self flushQueue];
-  }
+  NSMutableArray<MSAuthTokenValidityInfo *> *tokenArray = [MSAuthTokenContext sharedInstance].authTokenValidityArray;
```
jaelim-ms: Can it be `[[MSAuthTokenContext sharedInstance] authTokenValidityArray]`? There is a custom getter, it can be accessed like that.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,16 +48,67 @@ + (void)resetSharedInstance {
...
+    }
+    NSMutableArray<MSAuthTokenInfo *> *authTokenHistory = [[self authTokenHistory] mutableCopy];
+
+    MSAuthTokenInfo *lastEntry = authTokenHistory.lastObject;
+    NSString *__nullable latestAuthToken = lastEntry.authToken;
+    NSString *__nullable latestAccountId = lastEntry.accountId;
+    NSDate *__nullable latestTokenEndTime = lastEntry.expiresOn;
+
+    // If new token doesn't differ from the last token of array - no need to add it to array.
+    if (latestAuthToken != nil && authToken != nil && [latestAuthToken isEqualToString:(NSString * __nonnull) authToken]) {
```
jaelim-ms: I think you can simplify the condition like below.
```suggestion
    if (latestAuthToken != nil && [latestAuthToken isEqualToString:(NSString * __nonnull) authToken]) {
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -53,16 +48,67 @@ + (void)resetSharedInstance {
...
+      return;
+    }
+    BOOL oneOfIdsIsNil = ((accountId == nil && latestAccountId != nil) || (accountId != nil && latestAccountId == nil));
+    isNewUser = !(accountId == nil && latestAccountId == nil) &&
+                (oneOfIdsIsNil || ![accountId isEqualToString:(NSString * __nonnull) latestAccountId]);
```
jaelim-ms: This condition could be rewritten as below.
```
isNewUser = ![accountId ?: @"" isEqualToString:(lastestAccountId ?: @"")];
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: BTW don't you also want to skip when both are `nil`?

MatkovIvan:
Agree with https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1397#pullrequestreview-221453880
Let's merge it and do some polishing separately.

PS There are some conflicts, so @clpolet 's approve will be reset. I'm going to approve & merge it after solving conflicts.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
MatkovIvan: https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1397/files#diff-80d0c935d334ac8e361cbfaeea30eeebR152

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
MatkovIvan: @annakocheshkova This one is important, please fix it before merge

PR #1407, opened 2019-03-20
CosmosDB Read API

MehrdadMzfr:

* [ ] Did you add unit tests?

## Description

Added Read Document API from CosmosDB


AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,6 +183,81 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
 #pragma mark - MSDataStore Implementation
+
+- (void)readWithPartition:(NSString *)partition
+               documentId:(NSString *)documentId
+             documentType:(Class)documentType
+              readOptions:(MSReadOptions *)__unused readOptions
+        completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
+
+  // TODO consume writeOptions
+  [MSTokenExchange
```
achocron: Can this token acquisition step be moved into a different method or into the cosmos db class? It seems like we are doing the same thing in several places (and will need to do this in a lot of other methods in this class)

AppCenterDataStorage/AppCenterDataStorage/MSDataStoreErrors.h
```diff
@@ -0,0 +1,36 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - Domain
+
+// Error documentation here: https://docs.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb
+static NSString *const kMSACDataStoreErrorDomain = MS_APP_CENTER_BASE_DOMAIN @"DataStoreErrorDomain";
+
+#pragma mark - Error Codes
+
+// Error codes
```
achocron: nit: Do we need this comment?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -225,8 +300,8 @@ - (void)createWithPartition:(NSString *)partition
                                                                             body:body
                                                                completionHandler:^(NSData *_Nonnull data, NSError *_Nonnull cosmosDbError) {
                                                                  // If not created.
```
achocron: nit: Line break before comments

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,6 +183,81 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+
+                                 // Call CosmosDb.
+                                 [MSCosmosDb
+                                     performCosmosDbAsyncOperationWithHttpClient:cosmosDbIngestion
+                                                                     tokenResult:tokenResponses.tokens[0]
+                                                                      documentId:documentId
+                                                                      httpMethod:@"GET"
+                                                                            body:nil
+                                                               completionHandler:^(NSData *_Nonnull data, NSError *_Nonnull cosmosDbError) {
+                                                                 // If not created.
```
achocron: nit: Line break before comments

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,6 +183,81 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+                               completionHandler:^(MSTokensResponse *_Nonnull tokenResponses, NSError *_Nonnull error) {
+                                 // If error getting token.
+                                 if (error || !tokenResponses) {
+                                   MSLogError([MSDataStore logTag], @"Can't get CosmosDb token:%@", [error description]);
+                                   completionHandler([[MSDocumentWrapper alloc] initWithError:error documentId:documentId]);
+                                   return;
+                                 }
+
+                                 // Create http client.
+                                 MSCosmosDbIngestion *cosmosDbIngestion = [[MSCosmosDbIngestion alloc] init];
```
achocron: It looks like there is already an ingestion property for this class. Why not use it? (same question about the code this is copied from)

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,6 +183,81 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+                                                                 NSDictionary *json =
+                                                                     [NSJSONSerialization JSONObjectWithData:data
+                                                                                                     options:0
+                                                                                                       error:&deserializeError];
+                                                                 if (deserializeError) {
+                                                                   MSLogError([MSDataStore logTag], @"Error deserializing data:%@",
+                                                                              [deserializeError description]);
+                                                                 }
+                                                                 MSLogDebug([MSDataStore logTag], @"Document json:%@", json);
+
```
achocron: Can we also log the document id here?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -257,7 +332,8 @@ - (void)createWithPartition:(NSString *)partition
                                                                                                                documentId:documentId
                                                                                                                      eTag:eTag
                                                                                                           lastUpdatedDate:date];
-                                                                 MSLogDebug([MSDataStore logTag], @"Document created:%@", data);
+                                                                 MSLogDebug([MSDataStore logTag], @"Document created with Id:%@",
```
achocron: nit: Id -> ID

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.m
```diff
@@ -2,16 +2,70 @@
...
 }
 
++ (NSInteger)errorCodeWithError:(NSError *)error {
+  if (error == nil) {
+    return kMSACDocumentSucceededErrorCode;
+  }
+
+  // Get error code.
+  NSNumber *errorCode = [error userInfo][kMSCosmosDbHttpCodeKey];
+  switch ([errorCode integerValue]) {
```
achocron: What is this mapping? Can the int itself just be returned?

ElektrojungeAtWork: the user info is defined as `@property(readonly, copy) NSDictionary<NSErrorUserInfoKey, id> *userInfo;` so it's really up to us. I don't think there's an effect in practice. From a consistency standpoint, we don't have a large enough dataset in our SDK ;) there's only a handful of cases where we return an `int` or a `NSInteger`.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
MehrdadMzfr: -1 for moving token retrieval into CosmoDB, I prefer to keep them loosely coupled.

moving to a separate method won't save much, current implementation asks for token and calls completionHandler in case of error. Also, considering the Async nature of token retrieval and cosmosDB call, it requires nested calls from completion handler.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
MehrdadMzfr: this will change into a single http client after the http refactor work, currently we need 2 separate http clients(ingestion) for token and cosmos(since requests are different)

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
MehrdadMzfr: document id is as part of the json

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.m
MehrdadMzfr: user will be able to use this enumeration to identify errors. it's also good practice to provide custom message along with this error code(as opposed to generic error message coming from CosmosDB)

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
achocron: I'm fine with them being in a separate method instead of the CosmosDB class, and I think it's worth doing. performCosmosDbAsyncOperationWithHttpClient and performDbTokenAsyncOperationWithHttpClient can be wrapped into one method, thereby reducing the nesting level, improving readability, and removing code that is or will be duplicated many times over.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
achocron: But there is a property that isn't being used, and I fear that if we don't use it, it will be left as dead code. It shouldn't  be more than a couple of lines to change this now, right?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,13 +183,89 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+             documentType:(Class)documentType
+              readOptions:(MSReadOptions *)__unused readOptions
+        completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
+
+  // TODO consume writeOptions
+  [MSTokenExchange
+      performDbTokenAsyncOperationWithHttpClient:(MSStorageIngestion *)self.ingestion
+                                       partition:partition
+                               completionHandler:^(MSTokensResponse *_Nonnull tokenResponses, NSError *_Nonnull error) {
+                                 // If error getting token.
```
achocron: nit: line break

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -183,13 +183,89 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+                                                                     [NSJSONSerialization JSONObjectWithData:data
+                                                                                                     options:0
+                                                                                                       error:&deserializeError];
+                                                                 if (deserializeError) {
+                                                                   MSLogError([MSDataStore logTag], @"Error deserializing data:%@",
+                                                                              [deserializeError description]);
+                                                                 }
+                                                                 MSLogDebug([MSDataStore logTag], @"Document json:%@", json);
+
+                                                                 // Create document.
```
achocron: This comment is basically repeated on line 245. Is that a mistake?

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.m
achocron: Perhaps I am missing something, or I wasn't specific enough. The return type of this method is `NSInteger`. We do a switch on the int, and return an enum with the same value as the code we already have, and the enum type is `NSInteger`. Further, from the callee's point of view, the return type is an `NSInteger`. So there would be no perceivable difference if we just returned `[errorCode integerValue]`. So why do we need to switch? I am under the impression that this essentially is saying `case 200: return 200`, etc.

achocron: Discussed offline; I didn't realize it was a common practice to make sure that all the error codes are handled.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -187,91 +192,186 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+     
+     // Create http client.
+     MSCosmosDbIngestion *cosmosDbIngestion = [[MSCosmosDbIngestion alloc] init];
+     
+     // Call CosmosDb.
+     [MSCosmosDb
+      performCosmosDbAsyncOperationWithHttpClient:cosmosDbIngestion
+      tokenResult:tokenResponses.tokens[0]
+      documentId:documentId
+      httpMethod:@"GET"
```
achocron: Can you use a constant for GET?

jwallra: @yochaude and I have added one in our PR, it's probably OK for now to avoid conflicts

AppCenterDataStorage/AppCenterDataStorage/Internal/Client/MSCosmosDb.h
```diff
@@ -30,7 +30,26 @@ typedef void (^MSCosmosDbCompletionHandler)(NSData *data, NSError *error);
...
+
+/**
+ * Call CosmosDb Api and perform db actions(read, write, delete, list, etc).
+ *
+ * @param httpClient Http client to call perform http calls .
+ * @param tokenResult Token result object containing token value used to call CosmosDb Api.
+ * @param documentId Document Id.
+ * @param httpMethod Http method.
+ * @param body Http body.
+ 8 @param additionalHeaders Additional http headers.
```
jwallra:
```suggestion
 * @param additionalHeaders Additional http headers.
```

AppCenterDataStorage/AppCenterDataStorage/Internal/Client/MSCosmosDb.m
```diff
@@ -124,20 +130,30 @@ + (void)performCosmosDbAsyncOperationWithHttpClient:(MSCosmosDbIngestion *)httpC
...
   httpClient.httpMethod = httpMethod;
-  httpClient.httpHeaders = [MSCosmosDb defaultHeaderWithPartition:tokenResult.partition dbToken:tokenResult.token];
+  httpClient.httpHeaders = [MSCosmosDb defaultHeaderWithPartition:tokenResult.partition dbToken:tokenResult.token additionalHeaders:additionalHeaders];
   httpClient.sendURL = (NSURL *)[NSURL URLWithString:[MSCosmosDb documentUrlWithTokenResult:tokenResult documentId:documentId]];
   [httpClient sendAsync:body
       completionHandler:^(NSString *callId, NSHTTPURLResponse *response, NSData *data, NSError *error) {
         MSLogVerbose([MSDataStore logTag], @"CosmosDb HttpClient callback, request Id %@ with status code: %lu and description: %@", callId,
                      (unsigned long)response.statusCode, [error description]);
-
+
```
jwallra:
```suggestion

```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -187,91 +192,186 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+             documentType:(Class)documentType
+              readOptions:(MSReadOptions *)__unused readOptions
+        completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
+  
+  // TODO consume writeOptions.
+  [MSTokenExchange
+   performDbTokenAsyncOperationWithHttpClient:(MSStorageIngestion *)self.ingestion
+   partition:partition
+   completionHandler:^(MSTokensResponse *_Nonnull tokenResponses, NSError *_Nonnull error) {
+     // If error getting token.
```
jwallra:
```suggestion

     // If error getting token.
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -187,91 +192,186 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+        }
+        MSLogDebug([MSDataStore logTag], @"Document json:%@", json);
+        
+        // Create document.
+        id<MSSerializableDocument> deserializedDocument = [(id<MSSerializableDocument>)[documentType alloc]
+                                                           initFromDictionary:(NSDictionary *)json[kMSDocumentKey]];
+        
+        // Create a document.
+        NSTimeInterval interval =
+        [(NSString *)json[kMSDocumentTimestampKey] doubleValue];
```
jwallra: formatting is wrong, please run the script

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -187,91 +192,186 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+   }];
+}
+
 - (void)createWithPartition:(NSString *)partition
                  documentId:(NSString *)documentId
                    document:(id<MSSerializableDocument>)document
                writeOptions:(MSWriteOptions *)__unused writeOptions
           completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
-
+
```
jwallra:
```suggestion

```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -187,91 +192,186 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
-                                                                                                          lastUpdatedDate:date];
-                                                                 MSLogDebug([MSDataStore logTag], @"Document created:%@", data);
-                                                                 completionHandler(docWrapper);
-                                                                 return;
-                                                               }];
-                               }];
+   performDbTokenAsyncOperationWithHttpClient:(MSStorageIngestion *)self.ingestion
+   partition:partition
+   completionHandler:^(MSTokensResponse *_Nonnull tokenResponses, NSError *_Nonnull error) {
+     // If error getting token.
```
jwallra:
```suggestion

     // If error getting token.
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -331,7 +429,7 @@ - (void)startWithChannelGroup:(id<MSChannelGroupProtocol>)channelGroup
       transmissionTargetToken:(nullable NSString *)token
               fromApplication:(BOOL)fromApplication {
   [super startWithChannelGroup:channelGroup appSecret:appSecret transmissionTargetToken:token fromApplication:fromApplication];
-
+
```
jwallra:
```suggestion

```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
yochaude: fixed

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.h
```diff
@@ -19,4 +24,23 @@
...
+- (instancetype)initWithError:(NSError *)error errorCode:(NSInteger)errorCode;
+
+/**
+ * Get the error code from the error.
+ *
+ * @param error An error object.
+ *
+ * @return The error code.
+ */
++ (NSInteger)errorCodeWithError:(NSError *)error;
```
dhei: `errorCodeFromError:error` naming makes more sense to me. What do you think?

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.m
```diff
@@ -2,16 +2,34 @@
...
 
 @implementation MSDataSourceError
 
 @synthesize error = _error;
+@synthesize errorCode = _errorCode;
 
 - (instancetype)initWithError:(NSError *)error {
+  return [self initWithError:error errorCode:[MSDataSourceError errorCodeWithError:error]];
+}
+- (instancetype)initWithError:(NSError *)error errorCode:(NSInteger)errorCode {
```
dhei: nit: line break

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -272,40 +367,41 @@ - (void)deleteDocumentWithPartition:(NSString *)partition
...
-
                                  // If error getting token.
                                  if (tokenExchangeError || [tokenResponses.tokens count] == 0) {
-                                   NSNumber *httpStatusCode = [tokenExchangeError userInfo][kMSCosmosDbHttpCodeKey];
-                                     MSLogError([MSDataStore logTag], @"Can't get CosmosDb token. Error: %@;  HTTP status code: %d; Partition: %@", tokenExchangeError.localizedDescription,
-                                              [httpStatusCode intValue], partition);
+                                   NSInteger httpStatusCode = [MSDataSourceError errorCodeWithError:tokenExchangeError];
+                                   MSLogError([MSDataStore logTag],
+                                              @"Can't get CosmosDb token. Error: %@;  HTTP status code: %ld; Partition: %@",
+                                              tokenExchangeError.localizedDescription, (long)httpStatusCode, partition);
```
dhei: What is the reason casting `httpStatusCode` to `long` instead of `int`?

AppCenterDataStorage/AppCenterDataStorage/MSDataStoreErrors.h
```diff
@@ -0,0 +1,33 @@
...
+                   kMSACDocumentNoContentErrorCode = 204,
+                   kMSACDocumentBadRequestErrorCode = 400,
+                   kMSACDocumentUnauthorizedErrorCode = 401,
+                   kMSACDocumentForbiddenErrorCode = 403,
+                   kMSACDocumentNotFoundErrorCode = 404,
+                   kMSACDocumentRequestTimeoutErrorCode = 408,
+                   kMSACDocumentConflictErrorCode = 409,
+                   kMSACDocumentPreconditionFailedErrorCode = 412,
+                   kMSACDocumentEntityTooLargeErrorCode = 413,
+                   kMSACDocumentTooManyRequestErrorCode = 429,
```
dhei: `TooManyRequest` => `TooManyRequests`

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -272,40 +367,41 @@ - (void)deleteDocumentWithPartition:(NSString *)partition
...
-                                                                              @"Document: %@/%@", cosmosDbError.localizedDescription,
-                                                                              [httpStatusCode intValue], partition, documentId);
+                                                                 NSInteger httpStatusCode =
+                                                                     [MSDataSourceError errorCodeWithError:tokenExchangeError];
+                                                                 if (httpStatusCode != MSHTTPCodesNo204NoContent) {
+                                                                   MSLogError(
+                                                                       [MSDataStore logTag],
+                                                                       @"Not able to delete document. Error: %@; HTTP status code: %ld; "
+                                                                       @"Document: %@/%@",
+                                                                       cosmosDbError.localizedDescription, (long)httpStatusCode, partition,
```
dhei: Same question here.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
dhei: Per https://stackoverflow.com/a/16076218/873234, it's recommended to cast `NSInteger` to `long` to be formatted correctly on 32-bit/64-bit system.
> Since NSInteger is 32-bit or 64-bit, depending on the platform, the compiler recommends to add a cast to long generally.

AppCenterDataStorage/AppCenterDataStorage/MSDataSourceError.h
MehrdadMzfr: convention is to provide methodNameWithFirstArgument:

dhei: @MehrdadMzfr While we stick to that convention most of the time, but in this case, at least it makes sense to me to make it more readable as the return type "errorCode" is something to extract from the first parameter "error".

Anyway, I don't have a strong preference and it's not a big deal either way, just sharing my thoughts.

PR #1415, opened 2019-03-21
Fix the build errors due to missing some h files in datastorage framework

jinquan2019:
Fix: Change missing h files as public than they can export to framework 
Issue: 
Some h files are not exported to data storage framework, such as `MSDataStore.h`
This cause the sasquatchObjc and sasquatchSwift app fail to build.
image
image


jinquan2019:
> Can you give me more details of the issue?

Updated

jaelim-ms:
Totally makes sense. I think it wasn't detected because we didn't build SasquatchObjC and SasquatchSwift in feature branches. Good catch!

PR #1416, opened 2019-03-21
Fix unit test [MSChannelUnitDefaultTests testResumeWhenOnlyPausedObjectIsDeallocated] intermittently fails in CI

charmian01:

* [ ] Did you add unit tests?

## Description
Fix unit test [MSChannelUnitDefaultTests testResumeWhenOnlyPausedObjectIsDeallocated] intermittently fails.
Before:
It's because given parameters both are new object when call `pauseWithIdentifyingObjectSync` and   `resumeWithIdentifyingObjectSync` method in `testResumeWhenOnlyPausedObjectIsDeallocated` test.
In`pauseWithIdentifyingObjectSync` method, add the income parameter for `pausedIdentifyingObjects` .
In`resumeWithIdentifyingObjectSync` method, remove the income parameter for `pausedIdentifyingObjects`. 
So the removed object is sometimes different from the added object. It cause the added object can't be removed for `pausedIdentifyingObjects` and the test finally failed. 
After:
  Make sure given a same parameter when call `pauseWithIdentifyingObjectSync` and `resumeWithIdentifyingObjectSync` method, so added object will always be removed for `pausedIdentifyingObjects`.

## Related PRs or issues
[58813](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/58813/?triage=true)

### DIFF
--- a/AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
+++ b/AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
@@ -1098,8 +1098,12 @@ static NSString *const kMSTestGroupId = @"GroupId";
   NSObject *object = [NSObject new];
   [self.sut pauseWithIdentifyingObjectSync:object];
 
+  // Then
+  XCTAssertTrue([self.sut paused]);
+
   // When
-  [[self.sut pausedIdentifyingObjects] removeObject:object];
+  // To use `dealloc`, `-fno-objc-arc` flag should be enabled for this class in Build Phases > Compile Sources.
+  [object dealloc];
   [self.sut resumeWithIdentifyingObjectSync:[NSObject new]];
 
   // Then
--- a/AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
+++ b/AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
@@ -1095,15 +1095,21 @@ static NSString *const kMSTestGroupId = @"GroupId";
 - (void)testResumeWhenOnlyPausedObjectIsDeallocated {
 
   // If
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -1095,10 +1095,11 @@ - (void)testResumesWhenAllPauseObjectsResumed {
 - (void)testResumeWhenOnlyPausedObjectIsDeallocated {
 
   // If
-  [self.sut pauseWithIdentifyingObjectSync:[NSObject new]];
+  NSObject *object = [NSObject new];
+  [self.sut pauseWithIdentifyingObjectSync:object];
 
   // When
-  [self.sut resumeWithIdentifyingObjectSync:[NSObject new]];
+  [self.sut resumeWithIdentifyingObjectSync:object];
```
achocron: I don't think this is the right fix. Although this change will make the test check a different valid scenario, the test is specifically for when the "pause" object is deallocated. Maybe instead of calling "resume" we can force the object to be deallocated in the "When" section.

clpolet: Try `object = nil` after pause. If that doesn't work try the whole pause new object inside a block to force switching scope.

charmian01: @achocron I'm not clear how to force the "pause" object to be deallocated in the "When" section. Do you mean remove the object form "pausedIdentifyingObjects" in the "When" section? I pushed a new commit for it, please help to take a look.

PR #1419, opened 2019-03-21
Feature/add helper method

achocron:

* [x] Did you add unit tests?

## Description

Combine token acquisition and cosmos db operation in a helper method for MSDataStore.m.

### DIFF
--- a/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
+++ b/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
@@ -220,7 +220,6 @@ static dispatch_once_t onceToken;
                                 body:nil
                    additionalHeaders:nil
                    completionHandler:^(NSData *data, NSError *_Nonnull cosmosDbError) {
-
                      // If not created.
                      if (!data || [MSDataSourceError errorCodeWithError:cosmosDbError] != kMSACDocumentSucceededErrorCode) {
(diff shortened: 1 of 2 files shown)


AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -217,207 +214,154 @@ - (void)readWithPartition:(NSString *)partition
...
                          documentId:(NSString *)documentId
                        writeOptions:(MSWriteOptions *)__unused writeOptions
                   completionHandler:(MSDataSourceErrorCompletionHandler)completionHandler {
+  [self performOperationForPartition:partition
+                          documentId:documentId
+                          httpMethod:kMSHttpMethodDelete
+                                body:[NSData data]
+                   additionalHeaders:nil
+                   completionHandler:^(NSData *__unused data, NSError *_Nonnull cosmosDbError) {
+                     // Body returned from call (data) is empty.
```
MehrdadMzfr: nit: empty line

achocron: Actually this is one of the unfortunate side effects of our formatter. When I add an empty line, the formatter removes it. Thus, if we commit it with the empty line, this will just come up the next time someone changes the file. In this case, we usually just keep the comment with no empty line before it.

PR #1423, opened 2019-03-22
Skip test and import validation on lint

jaelim-ms:
Skip test and import validation on pod lint temporarily. CocoaPods is broken when it runs with Xcode 10.1.


jaelim-ms:
Do not merge until the workaround is verified.

PR #1431, opened 2019-03-22
Feature/add storage to podspec

clpolet:

## Description

This is adding DataStorage to podspec and update AzDO build scripts. 
Validated podspec and tested on an application locally.

### DIFF
--- a/AppCenter.podspec
+++ b/AppCenter.podspec
@@ -84,4 +84,11 @@ Pod::Spec.new do |s|
     ss.osx.vendored_frameworks = "AppCenter-SDK-Apple/macOS/AppCenterPush.framework"
  end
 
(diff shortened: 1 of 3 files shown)


AppCenter.podspec
```diff
@@ -21,6 +21,9 @@ Pod::Spec.new do |s|
                       4. App Center Push (iOS and macOS):
                       App Center Push enables you to send push notifications to users of your app from the App Center portal. You can also segment your user base based on a set of properties and send them targeted notifications.
 
+                      5. App Center DataStorage (iOS):
+                      TODO: Add description for Datastorage
```
clpolet: We created a work item to update the description when we'll have the right wording.

PR #1432, opened 2019-03-23
Refactor Http Client (without touching ingestion for now)

guperrot:

* [x] Did you add unit tests?

## Description

New HTTP Client aiming at replacing direct http usage in ingestion classes.
This PR does not actually replace usages in ingestion yet, just adding the new http library.
It also adds the x-ms-retry-after header in retry management.

[AB#58169](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=58169)


AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,327 @@
...
+  NSDictionary *headers = nil;
+  NSData *payload = nil;
+
+  // When
+  [self.sut sendAsync:url
+                 method:method
+                headers:headers
+                   data:payload
+      completionHandler:^(NSData *responseBody, NSHTTPURLResponse *response, NSError *error) {
+        // Then
```
jwallra:
```suggestion

        // Then
```

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,321 @@
...
+  XCTAssertEqualObjects(actualRequest.URL, url);
+  XCTAssertEqualObjects(actualRequest.HTTPMethod, method);
+  XCTAssertEqualObjects(actualRequest.allHTTPHeaderFields, headers);
+}
+
+- (void)testGetWithHeadersWhileNetworkError {
+
+  // If
+  __block NSURLRequest *actualRequest;
+  // Stub HTTP response.
```
jwallra:
```suggestion

  // Stub HTTP response.
```

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,321 @@
...
+  [OHHTTPStubs
+      stubRequestsPassingTest:^BOOL(__attribute__((unused)) NSURLRequest *request) {
+        ++numRequests;
+        actualRequest = request;
+        return YES;
+      }
+      withStubResponse:^OHHTTPStubsResponse *(__attribute__((unused)) NSURLRequest *request) {
+        return [OHHTTPStubsResponse responseWithData:[NSData data] statusCode:MSHTTPCodesNo204NoContent headers:nil];
+      }];
+  self.sut = [[MSHttpClient alloc] initWithRetryIntervals:@[ @1 ] reachability:self.reachabilityMock];
```
jwallra: All other retry intervals are doubles (or whatever the default is for numbers with fractional parts), any reason this one is `@1` instead of `@1.0`?

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,321 @@
...
+  NSDictionary *headers = [NSDictionary new];
+  NSData *payload = nil;
+
+  // When
+  [self.sut sendAsync:url
+                 method:method
+                headers:headers
+                   data:payload
+      completionHandler:^(NSData *responseBody, NSHTTPURLResponse *response, NSError *error) {
+        // Then
```
jwallra:
```suggestion

        // Then
```

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
```diff
@@ -2,20 +2,45 @@
...
+#import "MSUtility+StringFormatting.h"
+#import "MS_Reachability.h"
 
 @implementation MSHttpClient
 
 - (instancetype)init {
-
-  // TODO implement this properly.
-  return self = [super init];
+  if ((self = [super init])) {
```
clpolet: The initialization pattern here is not really conventional. The initializers in objective c are usually chained as explained in [the Apple documentation](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html) see _Multiple Initializers and the Designated Initializer_.

That means you need to add a designated initializer with max coverage and chain the inits `initWithMaxHttpConnectionsPerHost:retryIntervals:` :
- `init` would call `initWithMaxHttpConnectionsPerHost:retryIntervals:` with default values.
- `initWithMaxHttpConnectionsPerHost:` would call `initWithMaxHttpConnectionsPerHost:retryIntervals:` with default value for retry interval but custom value for max http connexion.
- `initWithRetryIntervals:` would call `initWithMaxHttpConnectionsPerHost:retryIntervals:` with default value for max http connexion but custom value for retry interval.

The main purpose is to enforce the initialization of every properties and not leave one behind when an object is initializing this class or its subclass whatever the initializer used.

guperrot: Oh yeah I had the same comment while pairing and we already fixed it. Should be addressed by 8de53d2

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,327 @@
...
+  __weak XCTestExpectation *expectation = [self expectationWithDescription:@""];
+  self.sut = [MSHttpClient new];
+  NSURL *url = [NSURL URLWithString:@"https://mock/something?a=b"];
+  NSString *method = @"DELETE";
+  NSDictionary *headers = nil;
+  NSData *payload = nil;
+
+  // When
+  [self.sut sendAsync:url
+                 method:method
```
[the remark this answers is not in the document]
guperrot: The formatter does that, when I align it reverts to that.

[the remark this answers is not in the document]
guperrot: Yeah even when I put everything in 1 line it reverts to that... No way to fix it.

AppCenter/AppCenterTests/MSHttpClientTests.m
guperrot: AppCode refactor inline did that. Will restore.

guperrot: Actually should be ints.

guperrot: Removing `.0` instead where used.

AppCenter/AppCenter/Internals/HttpClient/MSHttpCall.m
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSHttpCall.h"
+#import "MSAppCenterInternal.h"
+#import "MSCompression.h"
+#import "MSConstants+Internal.h"
+#import "MSLogger.h"
+#import <Foundation/Foundation.h>
```
jaelim-ms: We put `<>` at the top of import. Please sort.

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -145,6 +145,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   // Set delegates.
   [self crashes];
   [self setAppCenterDelegate];
+
```
jaelim-ms: ??

AppCenter/AppCenterTests/MSHttpClientTests.m
jaelim-ms: I think this is done by formatter.

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,552 @@
...
+}
+
+- (void)tearDown {
+  [super tearDown];
+
+  [MSHttpTestUtil removeAllStubs];
+  [self.reachabilityMock stopMocking];
+
+  /*
+   * Setting the variable to nil. We are experiencing test failure on Xcode 9 beta because the instance that was used for previous test
```
jaelim-ms: Is it a good time to consider removing `self.sut` in this file and use local variable in each test? This will allow us to run all unit tests in parallel to speed up.

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,552 @@
...
+#import "MSDevice.h"
+#import "MSDeviceInternal.h"
+#import "MSHttpClient.h"
+#import "MSHttpIngestionPrivate.h"
+#import "MSHttpTestUtil.h"
+#import "MSIngestionCall.h"
+#import "MSIngestionDelegate.h"
+#import "MSMockLog.h"
+#import "MSTestFrameworks.h"
+#import <OHHTTPStubs/NSURLRequest+HTTPBodyTesting.h>
```
jaelim-ms: Sort. Our script doesn't sort `<>` import.

AppCenter/AppCenterTests/MSHttpCallTests.m
```diff
@@ -0,0 +1,71 @@
...
+  XCTAssertEqualObjects(call.data, expectedData);
+  XCTAssertEqualObjects(call.headers, expectedHeaders);
+}
+
+- (void)testDoesNotCompressHTTPBodyWhenSmall {
+
+  // If
+
+  // HTTP body is big enough to be compressed.
+  NSData *shortData = [NSData dataWithBytes:"h" length:1];
```
jaelim-ms: Is this big enough? You should update above comment.

AppCenter/AppCenterTests/MSHttpCallTests.m
```diff
@@ -0,0 +1,71 @@
...
+#import "MSDevice.h"
+#import "MSDeviceInternal.h"
+#import "MSHttpCall.h"
+#import "MSHttpIngestionPrivate.h"
+#import "MSHttpTestUtil.h"
+#import "MSIngestionCall.h"
+#import "MSIngestionDelegate.h"
+#import "MSMockLog.h"
+#import "MSTestFrameworks.h"
+#import <OHHTTPStubs/NSURLRequest+HTTPBodyTesting.h>
```
jaelim-ms: Sort.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
```diff
@@ -24,14 +54,206 @@ - (void)sendAsync:(NSURL *)url
...
+                                           userInfo:@{NSLocalizedDescriptionKey : kMSACCanceledErrorDesc}];
+          call.completionHandler(nil, nil, error);
+        }
+        [self.pendingCalls removeAllObjects];
+      }
+    }
+  }
+}
+
+- (NSString *)obfuscateHeaderValue:(NSString *)value forKey:(NSString *)key {
```
jaelim-ms: Please also consider to refactor this one. Obfuscating app secret is only for Ingestion, not a good fit to stay in generic HttpClient.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.h
```diff
@@ -22,4 +22,11 @@
  */
 - (instancetype)initWithMaxHttpConnectionsPerHost:(int)maxHttpConnectionsPerHost;
 
+/**
+ * Enables or disables the client. All pending requests are canceled and discarded upon disabling.
+ *
+ * @param isEnabled The desired enabled state of the client - pass `YES` to enable, `NO` to disable.
+ */
+- (void)setEnabled:(BOOL)isEnabled;
```
jaelim-ms: What's the purpose of enable/disable HttpClient?

AppCenter/AppCenter/Internals/HttpClient/MSHttpClientPrivate.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSHttpClient.h"
+#import <Foundation/Foundation.h>
```
jaelim-ms: Sort.

AppCenter/AppCenter/Internals/HttpClient/Util/MSHttpUtil.m
```diff
@@ -23,10 +23,6 @@ + (BOOL)isSSLConnectionError:(NSError *)error {
           ((error.code >= NSURLErrorCannotLoadFromNetwork) && (error.code <= NSURLErrorSecureConnectionFailed)));
 }
 
-+ (BOOL)isRequestCanceledError:(NSError *)error {
```
jaelim-ms: Good catch. It is dead code.

AppCenter/AppCenter/Internals/HttpClient/MSHttpCall.h
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSHttpClientProtocol.h"
+#import <Foundation/Foundation.h>
```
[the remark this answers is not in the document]
guperrot: If I move the formatter reverts it.

AppCenter/AppCenter/Internals/HttpClient/MSHttpCall.m
guperrot: The formatter reverts that if I try to address the feedback.

Sasquatch/SasquatchObjC/AppDelegate.m
guperrot: Please clarify what you meant here.

AppCenter/AppCenterTests/MSHttpClientTests.m
guperrot: No it does not.

AppCenter/AppCenterTests/MSHttpCallTests.m
guperrot: No the formatter reverts that.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.h
guperrot: To actually cancel the http calls while pause just says don't retry but keep calls in flight.
Pause is called on network down. disable is called on module disabled.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClientPrivate.h
guperrot: No the formatter reverts that.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
guperrot: We actually still need that otherwise we can't log any headers in httpclient.

AppCenter/AppCenter/Internals/HttpClient/MSHttpCall.h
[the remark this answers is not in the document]
jaelim-ms: Seems new toolchain has a different behavior.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
jaelim-ms: My suggestion was synced offline. We can do this later since we only have appSecret that needs to be obfuscated.

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,676 @@
...
+        completionHandlerCalled = YES;
+        XCTAssertNotNil(responseBody);
+        XCTAssertNotNil(response);
+        XCTAssertEqual(response.statusCode, MSHTTPCodesNo204NoContent);
+        XCTAssertNil(error);
+        [expectation fulfill];
+      }];
+
+  // Wait a little to ensure that the completion handler is not invoked yet.
+  sleep(1);
```
jaelim-ms: Do you really need to sleep here? You can move L245-249 within expectation handler after L256.
It would be `else` case of the `if`.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
```diff
@@ -2,20 +2,50 @@
...
 
-  // TODO implement this properly.
-  return self = [super init];
+- (instancetype)init {
+  return [self initWithMaxHttpConnectionsPerHost:nil
+                                  retryIntervals:DEFAULT_RETRY_INTERVALS
+                                    reachability:[MS_Reachability reachabilityForInternetConnection]];
 }
 
 - (instancetype)initWithMaxHttpConnectionsPerHost:(int)maxHttpConnectionsPerHost {
```
clpolet: Why the choice of `int` here vs `NSInteger` before?

AppCenter/AppCenterTests/MSHttpClientTests.m
achocron: I don't think that would work. We need L245-249 to execute before the completion handler is invoked (in fact, unless line 249 occurs, the completion handler will not be invoked). Since the expectation is fulfilled in the completion handler, the expectation handler is too late for that code to execute.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
```diff
@@ -2,20 +2,50 @@
...
+                                   retryIntervals:(NSArray *)retryIntervals
+                                     reachability:(MS_Reachability *)reachability {
   if ((self = [super init])) {
+    _sessionConfiguration = [NSURLSessionConfiguration defaultSessionConfiguration];
+    if (maxHttpConnectionsPerHost) {
+      _sessionConfiguration.HTTPMaximumConnectionsPerHost = [maxHttpConnectionsPerHost intValue];
+    }
+    _session = [NSURLSession sessionWithConfiguration:_sessionConfiguration];
+    _pendingCalls = [NSMutableSet new];
+    _retryIntervals = [NSArray arrayWithArray:retryIntervals];
```
clpolet: Can we make this property a `copy` instead of using `arrayWithArray` or that wasn't the only reason?

achocron: But does making it `copy` copy the original parameter? I thought `copy` means it will be copied if it's on the right-hand side of the assignment, but nothing if it's on the left.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
```diff
@@ -2,20 +2,50 @@
...
+#import "MSHttpClientPrivate.h"
+#import "MSHttpUtil.h"
+#import "MSLoggerInternal.h"
+#import "MSUtility+StringFormatting.h"
+#import "MS_Reachability.h"
 
 @implementation MSHttpClient
 
-- (instancetype)init {
+#define DEFAULT_RETRY_INTERVALS @[ @10, @(5 * 60), @(20 * 60) ]
```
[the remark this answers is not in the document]
achocron: No reason. Can move.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
achocron: Do you mean because the next init uses `NSNumber`? That is because we need it to be nullable.

guperrot: The default value is different on iOS (4) and macos (6), so we want the type to be nullable to use the default. The NSNumber constructor is not public.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
achocron: i think i had it backwards. will fix

AppCenter/AppCenterTests/MSHttpClientTests.m
```diff
@@ -0,0 +1,676 @@
...
+
+  // Then
+  XCTAssertFalse(completionHandlerCalled);
+
+  // Restore the network and wait for completion handler to be called.
```
jaelim-ms: Please add `// When` here.

AppCenter/AppCenter/Internals/HttpClient/MSHttpClient.m
clpolet: Wait actually forget about this comment that only works when using the accessor for the property so outside of the init.
But yeah copy would call NSCopying on the right object to assign the copy to the left object.

PR #1433, opened 2019-03-23
Enable verbose log on pod spec lint

jaelim-ms:


jaelim-ms:
Bypass CI since this is not a code change and there are several tests intermittently fail. All test failures are captured as bugs.

PR #1436, opened 2019-03-26
Feature/storage list documents

yochaude:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [X] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Implementing the list documents API.

PR #1438, opened 2019-03-27
Add offlineMode to Storage

jaelim-ms:
* [X] Did you add unit tests?

## Description

Adding offlineMode API to Storage so that customers can simulate local cache access when network is disconnected.


AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -373,6 +383,12 @@ - (void)performOperationForPartition:(NSString *)partition
                                             }];
 }
 
+- (BOOL)isOfflineMode {
```
clpolet: Do we need to override the getter here?

AppCenterDataStorage/AppCenterDataStorage/Internal/MSDataStorePrivate.h
```diff
@@ -13,6 +13,11 @@ static NSString *const kMSDefaultApiUrl = @"https://api.appcenter.ms/v0.1";
 
 @interface MSDataStore () <MSAuthTokenContextDelegate>
 
+/**
+ * A flag that indicates offline mode is on or off.
+ */
+@property(nonatomic) BOOL offlineMode;
```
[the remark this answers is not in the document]
jaelim-ms: This is the one that was decided for both platform.

AppCenterDataStorage/AppCenterDataStorage/Internal/Ingestion/MSCosmosDbIngestion.h
```diff
@@ -6,7 +6,18 @@
...
+ * Error domain for Storage.
+ */
+static NSString *const MSDataStorageErrorDomain = @"MSDataStorageErrorDomain";
+
 @interface MSCosmosDbIngestion : MSHttpIngestion
+
+/**
+ * A flag that indicates offline mode is on or off.
+ */
+@property(nonatomic) BOOL offlineMode;
```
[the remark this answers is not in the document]
jaelim-ms: It has a custom setter below and synchronized there. Looking both comment here and there, we can remove the custom setter and make it atomic.

AppCenterDataStorage/AppCenterDataStorage/Internal/Ingestion/MSCosmosDbIngestion.m
```diff
@@ -9,6 +9,8 @@
 
 @implementation MSCosmosDbIngestion
 
+@synthesize offlineMode = _offlineMode;
```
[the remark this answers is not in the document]
jaelim-ms: This was because there is a custom setter. It is gone now.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
```diff
@@ -182,5 +182,18 @@ typedef void (^MSDataSourceErrorCompletionHandler)(MSDataSourceError *error);
                        writeOptions:(MSWriteOptions *)writeOptions
                   completionHandler:(MSDataSourceErrorCompletionHandler)completionHandler;
 
+/**
+ * Enable or disable offline mode.
```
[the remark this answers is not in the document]
jaelim-ms: Yeah.. We can improve the doc but it has same doc in Android. I totally agree with you but we can improve this later. For the usage in prod env, there is no spec for the one. We can talk about this one but it shouldn't be a problem .

AppCenterDataStorage/AppCenterDataStorage/Internal/Ingestion/MSCosmosDbIngestion.h
```diff
@@ -6,7 +6,18 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ * Error domain for Storage.
+ */
+static NSString *const MSDataStorageErrorDomain = @"MSDataStorageErrorDomain";
```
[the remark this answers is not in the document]
jaelim-ms: It depends on what we expose to customers.
Current implementation only expose error code to customers so it could be in the header but it might be changed soon. Related to [this](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1436/files#r269817126)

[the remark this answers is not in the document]
jaelim-ms: Waiting for final decision.

AppCenterDataStorage/AppCenterDataStorage/Internal/Ingestion/MSCosmosDbIngestion.m
```diff
@@ -17,10 +17,24 @@ - (instancetype)init {
...
 
+- (void)sendAsync:(nullable NSObject *)data
+                 eTag:(nullable NSString *)eTag
+            authToken:(nullable NSString *)authToken
+               callId:(NSString *)callId
+    completionHandler:(MSSendAsyncCompletionHandler)completionHandler {
+  if (self.offlineMode) {
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey : @"Storage offline simulation mode is enabled."};
+    completionHandler(callId, nil, nil,
+                      [NSError errorWithDomain:MSDataStorageErrorDomain code:NSURLErrorNotConnectedToInternet userInfo:userInfo]);
```
clpolet:
```suggestion
                      [NSError errorWithDomain:kMSDataStorageErrorDomain code:NSURLErrorNotConnectedToInternet userInfo:userInfo]);
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
```diff
@@ -16,6 +16,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ * Error domain for Storage.
+ */
+static NSString *const MSDataStorageErrorDomain = @"MSDataStorageErrorDomain";
```
clpolet:
```suggestion
static NSString *const kMSDataStorageErrorDomain = @"MSDataStorageErrorDomain";
```

AppCenterDataStorage/AppCenterDataStorageTests/MSCosmosDbIngestionTests.m
```diff
@@ -0,0 +1,55 @@
...
+  [ingestionMock sendAsync:testData
+                      eTag:testETag
+                 authToken:testAuthToken
+                    callId:testCallId
+         completionHandler:^(NSString *callId, NSHTTPURLResponse *response, NSData *data, NSError *error) {
+           XCTAssertEqualObjects(testCallId, callId);
+           XCTAssertNil(response);
+           XCTAssertNil(data);
+           XCTAssertNotNil(error);
+           XCTAssertEqualObjects(error.domain, MSDataStorageErrorDomain);
```
clpolet:
```suggestion
           XCTAssertEqualObjects(error.domain, kMSDataStorageErrorDomain);
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
achocron: Why do we use a new error domain? There is already one defined for AppCenter. I think that we should use it for all modules.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -186,6 +186,18 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
                                           completionHandler:completionHandler];
 }
 
++ (void)setOfflineMode:(BOOL)offlineMode {
+  @synchronized([MSDataStore sharedInstance]) {
+    [MSDataStore sharedInstance].offlineMode = offlineMode;
```
achocron: This does not need to be synchronized since the property is `atomic`

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -186,6 +186,18 @@ + (void)deleteDocumentWithPartition:(NSString *)partition
...
+  }
+}
+
++ (BOOL)isOfflineMode {
+  @synchronized([MSDataStore sharedInstance]) {
+    return [MSDataStore sharedInstance].offlineMode;
```
achocron: This does not need to be synchronized since the property is `atomic`

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
jaelim-ms: Do you want to have a single domain that covers all different errors? What's the benefit of combining all A/C errors to one domain?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
jaelim-ms: Yeah.. That's true..

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
jaelim-ms: Yep. It was changed from `nonatomic` to `atomic` and forgot to clean up those.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
achocron: It's just more appropriate. Cleaner and easier for customers to use. Best to just see if it's an "app center" error instead of having to check for every single module's error domain type. Plus, every time we create a new error domain, we use the wrong format, but the one in the core module is formatted correctly.

PR #1459, opened 2019-04-02
Feature/storage list documents v2

yochaude:

* [x] Did you add unit tests?

## Description

Adding list API to the DataStorage module.

## Related PRs or issues

## Misc

### DIFF
--- a/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
+++ b/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSPage.h"
+#import "MSReadOptions.h"
+
+@interface MSPaginatedDocuments()
+
+// Read-only.
+@property(nonatomic, copy, readonly) NSString *partition;
+@property(nonatomic, readonly) Class documentType;
+@property(nonatomic, readonly) MSReadOptions *readOptions;
+
+// Read-write (to implement pagination).
+@property(nonatomic, readwrite) MSPage *currentPage;
+@property(nonatomic, copy, readwrite, nullable) NSString *continuationToken;
+
+@end
--- a/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
+++ b/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
@@ -4,7 +4,7 @@
 #import "MSPage.h"
 #import "MSReadOptions.h"
 
-@interface MSPaginatedDocuments()
+@interface MSPaginatedDocuments ()
 
 // Read-only.
 @property(nonatomic, copy, readonly) NSString *partition;
--- a/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
+++ b/AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
@@ -12,7 +12,7 @@
 @property(nonatomic, readonly) MSReadOptions *readOptions;
 
 // Read-write (to implement pagination).
-@property(nonatomic, readwrite) MSPage *currentPage;
-@property(nonatomic, copy, readwrite, nullable) NSString *continuationToken;
+@property(nonatomic) MSPage *currentPage;
+@property(nonatomic, copy, nullable) NSString *continuationToken;
 
 @end
--- a/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
+++ b/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
@@ -22,35 +22,15 @@
 #import "MSTokensResponse.h"
 #import "MSWriteOptions.h"
 
-/**
- * Service storage key name.
- */
+// Constants.
 static NSString *const kMSServiceName = @"DataStorage";
-
-/**
- * The group ID for storage.
- */
 static NSString *const kMSGroupId = @"DataStorage";
-
-/**
- * CosmosDb document timestamp key.
- */
+static NSString *const kMSDocumentIdKey = @"id";
 static NSString *const kMSDocumentTimestampKey = @"_ts";
-
-/**
- * CosmosDb document eTag key.
- */
 static NSString *const kMSDocumentEtagKey = @"_etag";
-
-/**
- * CosmosDb document key.
- */
 static NSString *const kMSDocumentKey = @"document";
-
-/**
- * CosmosDb upsert header key.
- */
 static NSString *const kMSDocumentUpsertHeaderKey = @"x-ms-documentdb-is-upsert";
+static NSString *const kMSDocumentContinuationTokenHeaderKey = @"x-ms-continuation";
 
 /**
  * Singleton.
@@ -91,7 +71,7 @@ static dispatch_once_t onceToken;
 + (void)readWithPartition:(NSString *)partition
                documentId:(NSString *)documentId
              documentType:(Class)documentType
-              readOptions:(MSReadOptions *)readOptions
+              readOptions:(MSReadOptions *_Nullable)readOptions
         completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
   [[MSDataStore sharedInstance] readWithPartition:partition
                                        documentId:documentId
@@ -103,21 +83,35 @@ static dispatch_once_t onceToken;
 + (void)listWithPartition:(NSString *)partition
              documentType:(Class)documentType
         completionHandler:(MSPaginatedDocumentsCompletionHandler)completionHandler {
-  // @todo
-  (void)partition;
-  (void)documentType;
-  (void)completionHandler;
+  [[MSDataStore sharedInstance] listWithPartition:partition
+                                     documentType:documentType
+                                      readOptions:nil
+                                continuationToken:nil
+                                completionHandler:completionHandler];
 }
 
 + (void)listWithPartition:(NSString *)partition
              documentType:(Class)documentType
-              readOptions:(MSReadOptions *)readOptions
+              readOptions:(MSReadOptions *_Nullable)readOptions
         completionHandler:(MSPaginatedDocumentsCompletionHandler)completionHandler {
-  // @todo
-  (void)partition;
-  (void)documentType;
-  (void)readOptions;
-  (void)completionHandler;
+  [[MSDataStore sharedInstance] listWithPartition:partition
+                                     documentType:documentType
+                                      readOptions:readOptions
+                                continuationToken:nil
+                                completionHandler:completionHandler];
+}
+
+// Internal signature with continuation token.
++ (void)listWithPartition:(NSString *)partition
+             documentType:(Class)documentType
+              readOptions:(MSReadOptions *_Nullable)readOptions
+        continuationToken:(NSString *_Nullable)continuationToken
+        completionHandler:(MSPaginatedDocumentsCompletionHandler)completionHandler {
+  [[MSDataStore sharedInstance] listWithPartition:partition
+                                     documentType:documentType
+                                      readOptions:readOptions
+                                continuationToken:continuationToken
+                                completionHandler:completionHandler];
 }
 
 + (void)createWithPartition:(NSString *)partition
@@ -134,7 +128,7 @@ static dispatch_once_t onceToken;
 + (void)createWithPartition:(NSString *)partition
                  documentId:(NSString *)documentId
                    document:(id<MSSerializableDocument>)document
-               writeOptions:(MSWriteOptions *)writeOptions
+               writeOptions:(MSWriteOptions *_Nullable)writeOptions
           completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
   [[MSDataStore sharedInstance] createWithPartition:partition
                                          documentId:documentId
@@ -157,7 +151,7 @@ static dispatch_once_t onceToken;
 + (void)replaceWithPartition:(NSString *)partition
                   documentId:(NSString *)documentId
                     document:(id<MSSerializableDocument>)document
-                writeOptions:(MSWriteOptions *)writeOptions
+                writeOptions:(MSWriteOptions *_Nullable)writeOptions
            completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
   [self replaceWithPartition:partition
                   documentId:documentId
@@ -177,7 +171,7 @@ static dispatch_once_t onceToken;
 
 + (void)deleteDocumentWithPartition:(NSString *)partition
                          documentId:(NSString *)documentId
-                       writeOptions:(MSWriteOptions *)writeOptions
+                       writeOptions:(MSWriteOptions *_Nullable)writeOptions
                   completionHandler:(MSDataSourceErrorCompletionHandler)completionHandler {
   [[MSDataStore sharedInstance] deleteDocumentWithPartition:partition
                                                  documentId:documentId
@@ -198,7 +192,7 @@ static dispatch_once_t onceToken;
 - (void)replaceWithPartition:(NSString *)partition
                   documentId:(NSString *)documentId
                     document:(id<MSSerializableDocument>)document
-                writeOptions:(MSWriteOptions *)writeOptions
+                writeOptions:(MSWriteOptions *_Nullable)writeOptions
            completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
 
   // In the current version we do not support E-tag optimistic concurrency logic and replace will call create.
@@ -213,7 +207,7 @@ static dispatch_once_t onceToken;
 - (void)readWithPartition:(NSString *)partition
                documentId:(NSString *)documentId
              documentType:(Class)documentType
-              readOptions:(MSReadOptions *)__unused readOptions
+              readOptions:(MSReadOptions *_Nullable)__unused readOptions
         completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {
   [self performOperationForPartition:partition
                           documentId:documentId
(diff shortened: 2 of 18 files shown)


AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
```diff
@@ -0,0 +1,18 @@
...
+
+@interface MSPaginatedDocuments()
+
+// Read-only.
+@property(nonatomic, copy, readonly) NSString *partition;
+@property(nonatomic, readonly) Class documentType;
+@property(nonatomic, readonly) MSReadOptions *readOptions;
+
+// Read-write (to implement pagination).
+@property(nonatomic, readwrite) MSPage *currentPage;
```
jaelim-ms: `readwrite` is by default. You can remove that.
```suggestion
@property(nonatomic) MSPage *currentPage;
```

AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
```diff
@@ -0,0 +1,18 @@
...
+@property(nonatomic, copy, readwrite, nullable) NSString *continuationToken;
```
jaelim-ms: `readwrite` is by default. You can remove that.
```suggestion
@property(nonatomic, copy, nullable) NSString *continuationToken;
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -22,35 +22,15 @@
 #import "MSTokensResponse.h"
 #import "MSWriteOptions.h"
 
-/**
```
jaelim-ms: Can we keep all of docs for constants?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -103,21 +83,35 @@ + (void)readWithPartition:(NSString *)partition
...
-  (void)readOptions;
-  (void)completionHandler;
+  [[MSDataStore sharedInstance] listWithPartition:partition
+                                     documentType:documentType
+                                      readOptions:readOptions
+                                continuationToken:nil
+                                completionHandler:completionHandler];
+}
+
+// Internal signature with continuation token.
```
jaelim-ms: You don't need this comment here. The method declaration is already in internal header so it should be enough.
```suggestion
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +345,108 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+- (void)listWithPartition:(NSString *)partition
+             documentType:(Class)documentType
+              readOptions:(MSReadOptions *_Nullable)readOptions
+        continuationToken:(nullable NSString *)continuationToken
+        completionHandler:(MSPaginatedDocumentsCompletionHandler)completionHandler {
+  NSMutableDictionary *additionalHeaders = [NSMutableDictionary new];
+  if (continuationToken) {
+    [additionalHeaders setObject:(NSString *)continuationToken forKey:kMSDocumentContinuationTokenHeaderKey];
+  }
+  // Call cosmos DB.
```
jaelim-ms:
```suggestion

  // Call cosmos DB.
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +345,108 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+  [self performOperationForPartition:partition
+                          documentId:nil
+                          httpMethod:kMSHttpMethodGet
+                                body:nil
+                   additionalHeaders:additionalHeaders
+                   completionHandler:^(NSData *_Nullable data, NSHTTPURLResponse *_Nullable response, NSError *_Nullable cosmosDbError) {
+                     // If not OK.
+                     if (!data || [MSDataSourceError errorCodeFromError:cosmosDbError] != MSACDocumentSucceededErrorCode) {
+                       MSLogError([MSDataStore logTag], @"Not able to retrieve documents: %@", [cosmosDbError description]);
```
jaelim-ms: We use `localizedDescription` for logs but I think this is correct. We shouldn't use localized description for logs.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
yochaude: Can I switch to // form though?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,109 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                       MSPaginatedDocuments *documents = [[MSPaginatedDocuments alloc] initWithError:dataSourceDeserializeError];
+                       completionHandler(documents);
+                       return;
+                     }
+
+                     // Parse the documents.
+                     NSMutableArray<MSDocumentWrapper *> *items = [NSMutableArray new];
+                     for (id document in jsonDocuments) {
+                       // Deserialize current document.
+                       // TODO: handle deserialization here.
```
jwallra: Hadn't we done this and the wrapper creation in the previous PR?  Not sure why they're TODO now

jwallra: Actually I'm not sure why this TODO is here at all, it seems like you're already doing deserialization here

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,109 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                       // TODO: handle deserialization error for CosmosDB internal properties here.
+                       NSTimeInterval interval = [(NSString *)document[kMSDocumentTimestampKey] doubleValue];
+                       NSDate *date = [NSDate dateWithTimeIntervalSince1970:interval];
+                       NSString *documentId = document[kMSDocumentIdKey];
+                       NSString *eTag = document[kMSDocumentEtagKey];
+                       NSString *jsonValue;
+                       {
+                         NSError *error;
+                         NSData *jsonData = [NSJSONSerialization dataWithJSONObject:document options:0 error:&error];
+                         if (!error)
```
jwallra: We don't usually use `if`s without brackets, please add them

AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -762,4 +795,152 @@ - (void)testOfflineModeBehavior {
...
+  id httpClient = OCMClassMock([MSHttpClient class]);
+  OCMStub([httpClient new]).andReturn(httpClient);
+  self.sut.httpClient = httpClient;
+  id msTokenEchange = OCMClassMock([MSTokenExchange class]);
+  OCMStub([msTokenEchange retrieveCachedToken:[OCMArg any]])
+      .andReturn([[MSTokenResult alloc] initWithDictionary:[self prepareMutableDictionary]]);
+  __weak XCTestExpectation *expectation = [self expectationWithDescription:@"List first page"];
+  NSMutableDictionary *continuationHeaders = [NSMutableDictionary new];
+  continuationHeaders[@"x-ms-continuation"] = @"continuation token";
+  // First page
```
jwallra:
```suggestion

  // First page
```

AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -762,4 +795,152 @@ - (void)testOfflineModeBehavior {
...
+  NSDictionary *firstPageHeaders = [MSCosmosDb defaultHeaderWithPartition:@"partition" dbToken:kMSTokenTest additionalHeaders:nil];
+  OCMStub([httpClient sendAsync:OCMOCK_ANY method:@"GET" headers:firstPageHeaders data:nil completionHandler:OCMOCK_ANY])
+      .andDo(^(NSInvocation *invocation) {
+        [invocation retainArguments];
+        MSHttpRequestCompletionHandler completionHandler;
+        [invocation getArgument:&completionHandler atIndex:6];
+        NSData *payload = [self getJsonFixture:@"oneDocumentPage"];
+        completionHandler(payload, [MSHttpTestUtil createMockResponseForStatusCode:200 headers:continuationHeaders], nil);
+      });
+  // Second page
```
jwallra:
```suggestion

  // Second page
```

AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -762,4 +795,152 @@ - (void)testOfflineModeBehavior {
...
+  NSDictionary *secondPageHeaders = [MSCosmosDb defaultHeaderWithPartition:@"partition"
+                                                                   dbToken:kMSTokenTest
+                                                         additionalHeaders:continuationHeaders];
+  OCMStub([httpClient sendAsync:OCMOCK_ANY method:@"GET" headers:secondPageHeaders data:nil completionHandler:OCMOCK_ANY])
+      .andDo(^(NSInvocation *invocation) {
+        [invocation retainArguments];
+        MSHttpRequestCompletionHandler completionHandler;
+        [invocation getArgument:&completionHandler atIndex:6];
+        NSData *payload = [self getJsonFixture:@"zeroDocumentsPage"];
```
jwallra: Would this ever happen?  Having a non-null continuation token and then getting a page with no documents on it?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
yochaude: We rely on user's implementation. If the deserialization method is not present, this throws some exceptions. The user's implementation could fail too so I think we should always wrap any kind of user serialization/deserialization in a try/catch block. It's not only list API it's all the read/writes, I will add a task on the board, this PR is getting big.

AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
yochaude: In theory yes. In practice it is less likely because we don't do cross-partition queries (scenario where this happens usually). For the sake of the test it's fine I think. The semantic is not wrong.

A valid example of this happening: you list a page of data and get a continuation token, at the same time someone delete the documents that were supposed to be in the next page, you use your continuation token to query the next page, it could be empty.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
yochaude: About logs, I want us to normalize logging in the "fit and finish" feature to make sure we do the same thing between Apple and Android platforms. @clpolet Also mentioned we probably should not log documents payload (would sensitive information be there), I think that's fair and would be part of this normalization.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
yochaude: Actually I was referring to the TODO few lines below. You are correct that one should not be there (artifacts of my conflicts resolution skills :)).

Fixed (my previous comment still holds but for the next TODO).

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
jaelim-ms: After looking into `NSError` and `NSObject` implementation, I should recall my previous comment. I got confused of *localized*.
`description` is in `NSObject` and it doesn't give you any of error details. You have to use `localizedDescription` which is a property of `NSError`. Sorry for wrong info.
```suggestion
                       MSLogError([MSDataStore logTag], @"Not able to retrieve documents: %@", [cosmosDbError localizedDescription]);
```

AppCenterDataStorage/AppCenterDataStorage/MSPaginatedDocuments.m
```diff
@@ -2,25 +2,74 @@
...
+    _continuationToken = nil;
+  }
+  return self;
 }
 
-- (MSPage<id<MSSerializableDocument>> *)currentPage {
-  // @todo
-  return nil;
+- (instancetype)initWithError:(MSDataSourceError *)error {
+  if ((self = [super init])) {
```
clpolet: Here you would `MSPage *pageWithError = [[MSPage alloc] initWithError:error];` before the `if` and call:
```suggestion
  if ((self = [self initWithPage:pageWithError])) {
```

AppCenterDataStorage/AppCenterDataStorage/MSPaginatedDocuments.m
```diff
@@ -2,25 +2,74 @@
...
+    MSPage *pageWithError = [[MSPage alloc] initWithError:error];
+    return [self initWithPage:pageWithError];
```
clpolet: With the above comment you don't need the return at this line ;)

AppCenterDataStorage/AppCenterDataStorageTests/Fixtures/oneDocumentPage.json
```diff
@@ -0,0 +1,19 @@
+{
```
clpolet: Nice you introduced fixtures for these tests 👍 .

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,110 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                           [(id<MSSerializableDocument>)[documentType alloc] initFromDictionary:(NSDictionary *)document];
+
+                       // Create a document wrapper object.
+                       // TODO: handle deserialization error for CosmosDB internal properties here.
+                       NSTimeInterval interval = [(NSString *)document[kMSDocumentTimestampKey] doubleValue];
+                       NSDate *date = [NSDate dateWithTimeIntervalSince1970:interval];
+                       NSString *documentId = document[kMSDocumentIdKey];
+                       NSString *eTag = document[kMSDocumentEtagKey];
+                       NSString *jsonValue;
+                       {
```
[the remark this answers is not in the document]
yochaude: I find it confusing to leak so many variables in the outer scope.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,110 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                         NSError *error;
+                         NSData *jsonData = [NSJSONSerialization dataWithJSONObject:document options:0 error:&error];
```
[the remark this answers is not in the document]
yochaude: Our document wrapper exposes the CosmosDB JSON payload as a string. It's what I am retrieving here. Could I do it differently?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,110 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                       MSDataSourceError *dataSourceDeserializeError = [[MSDataSourceError alloc] initWithError:deserializeError];
+                       MSPaginatedDocuments *documents = [[MSPaginatedDocuments alloc] initWithError:dataSourceDeserializeError];
+                       completionHandler(documents);
+                       return;
+                     }
+
+                     // Parse the documents.
+                     NSMutableArray<MSDocumentWrapper *> *items = [NSMutableArray new];
+                     for (id document in jsonDocuments) {
+                       // Deserialize current document.
```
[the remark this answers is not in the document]
jwallra: FYI: The formatter doesn't seem to like new lines directly after `{` - it will often remove them.  We use the formatter as the final check so this may have to stay like this

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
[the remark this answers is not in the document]
jwallra: It's a style that we haven't really used before.  Personally I think it's weird too, I would prefer it just have a comment above the code to separate it and get rid of the extraneous scope

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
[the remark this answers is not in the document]
yochaude: Confirmed, the formatter just removed the new line again...

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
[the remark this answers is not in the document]
clpolet: I guess it depends on what is this value used for by customer?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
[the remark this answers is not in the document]
clpolet: On a for loop? It only does this on blocks that's so weird!

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -349,17 +374,110 @@ - (void)createOrReplaceWithPartition:(NSString *)partition
...
+                       id<MSSerializableDocument> deserializedDocument =
+                           [(id<MSSerializableDocument>)[documentType alloc] initFromDictionary:(NSDictionary *)document];
+
+                       // Create a document wrapper object.
+                       // TODO: handle deserialization error for CosmosDB internal properties here.
+                       NSTimeInterval interval = [(NSString *)document[kMSDocumentTimestampKey] doubleValue];
+                       NSDate *date = [NSDate dateWithTimeIntervalSince1970:interval];
```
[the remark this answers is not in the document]
yochaude: Yeah I'll live that as future improvements.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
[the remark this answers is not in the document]
clpolet: The outer scope is not that big, shouldn't matter much.

jwallra:
🎉

PR #1460, opened 2019-04-02
Added functionality and tests for for failed status on token exchange

t-nafle:

* [X] Did you add unit tests?

## Description

Added functionality and tests for for failed status on token exchange

## Misc

[AB#59172](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=59172)


AppCenterDataStorage/AppCenterDataStorage/Internal/Client/MSTokenExchange.m
```diff
@@ -140,18 +139,21 @@ + (MSTokenResult *)retrieveCachedToken:(NSString *)partitionName {
     NSString *tokenString = [MSKeychainUtil stringForKey:[MSTokenExchange tokenKeyNameForPartition:partitionName]];
     if (tokenString) {
       MSTokenResult *tokenResult = [[MSTokenResult alloc] initWithString:tokenString];
+      if (tokenResult && ![tokenResult.status isEqualToString:kMSTokenResultSucceed]) {
```
guperrot: You can call a selector on nil and it will return NO here.
```suggestion
      if (![tokenResult.status isEqualToString:kMSTokenResultSucceed]) {
```

jaelim-ms: `tokenResult` is just instantiated so `tokenResult` won't be `nil`. Redundant check.
```suggestion
      if (![kMSTokenResultSucceed isEqualToString:tokenResult.status]) {
```
If you use the constant to compare the string, it won't have the issue that @guperrot mentioned above.

AppCenterDataStorage/AppCenterDataStorage/Internal/Client/MSTokenExchange.h
```diff
@@ -10,6 +10,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 typedef void (^MSGetTokenAsyncCompletionHandler)(MSTokensResponse *tokenResponses, NSError *_Nullable error);
 
+static NSString *const kMSTokenResultSucceed = @"Succeed";
```
jaelim-ms: This should go to private header.

AppCenterDataStorage/AppCenterDataStorageTests/MSTokenExchangeTests.m
```diff
@@ -218,6 +217,48 @@ - (void)testValidCachedTokenExists {
...
+  OCMStub([self.keychainUtilMock stringForKey:mockTokenKeyName]).andReturn(tokenString);
+  id<MSHttpClientProtocol> httpMock = OCMProtocolMock(@protocol(MSHttpClientProtocol));
+
+  // Mock returning failed token.
+  NSObject *failedToken = [self getFailedTokenData];
+  NSMutableDictionary *tokenList = [@{kMSTokens : @[ failedToken ]} mutableCopy];
+  NSData *jsonTokenData = [NSJSONSerialization dataWithJSONObject:tokenList options:NSJSONWritingPrettyPrinted error:nil];
+  OCMStub([httpMock sendAsync:OCMOCK_ANY method:OCMOCK_ANY headers:OCMOCK_ANY data:OCMOCK_ANY completionHandler:OCMOCK_ANY])
+      .andDo(^(NSInvocation *invocation) {
+        __block MSHttpRequestCompletionHandler completionBlock;
```
jaelim-ms: You don't need `__block`.

AppCenterDataStorage/AppCenterDataStorageTests/MSTokenExchangeTests.m
```diff
@@ -218,6 +217,48 @@ - (void)testValidCachedTokenExists {
...
+        [invocation getArgument:&completionBlock atIndex:6];
+        NSHTTPURLResponse *response = [NSHTTPURLResponse new];
+        id mockResponse = OCMPartialMock(response);
+        OCMStub([mockResponse statusCode]).andReturn(MSHTTPCodesNo200OK);
```
jaelim-ms: Use `MSHttpTestUtil` to generate 200OK response instead of partial mocking response.

AppCenterDataStorage/AppCenterDataStorageTests/MSTokenExchangeTests.m
```diff
@@ -218,6 +217,48 @@ - (void)testValidCachedTokenExists {
   [utilityMock stopMocking];
 }
 
+- (void)testRetrieveCachedTokenFails {
```
jaelim-ms: Can you update the test method name or add a comment what you are testing here?
The test name says it is testing a failure case when it retrieves cached token but there is a HTTP client mock that also returns failed token. The test contains two failure tokens 1. in cache 2. from HTTP response. What are you testing here?

AppCenterDataStorage/AppCenterDataStorage/Internal/Client/MSTokenExchange.m
```diff
@@ -140,18 +139,21 @@ + (MSTokenResult *)retrieveCachedToken:(NSString *)partitionName {
     NSString *tokenString = [MSKeychainUtil stringForKey:[MSTokenExchange tokenKeyNameForPartition:partitionName]];
     if (tokenString) {
       MSTokenResult *tokenResult = [[MSTokenResult alloc] initWithString:tokenString];
+      if (![kMSTokenResultSucceed isEqualToString:tokenResult.status]) {
```
yochaude: Can we also validate the actual status code from the HTTP call that I assume is done earlier.

I caught this bug when the token endpoint was returning us a 400 code and we were still trying to parse the non-existent payload.

yochaude: In `performDbTokenAsyncOperationWithHttpClient` (MSTokenExchange) we should also validate the status code before proceeding in parsing the response.

yochaude: And in this method actually, we only read a token from the cache. If you look in the method I linked above, only valid tokens (status = succeeded) are cached so this test is not needed.

AppCenterDataStorage/AppCenterDataStorageTests/MSTokenExchangeTests.m
```diff
@@ -218,6 +219,49 @@ - (void)testValidCachedTokenExists {
...
+  NSObject *failedToken = [self getFailedTokenData];
+  NSMutableDictionary *tokenList = [@{kMSTokens : @[ failedToken ]} mutableCopy];
+  NSData *jsonTokenData = [NSJSONSerialization dataWithJSONObject:tokenList options:NSJSONWritingPrettyPrinted error:nil];
+  OCMStub([httpMock sendAsync:OCMOCK_ANY method:OCMOCK_ANY headers:OCMOCK_ANY data:OCMOCK_ANY completionHandler:OCMOCK_ANY])
+      .andDo(^(NSInvocation *invocation) {
+        MSHttpRequestCompletionHandler completionBlock;
+        [invocation getArgument:&completionBlock atIndex:6];
+        NSHTTPURLResponse *response = [NSHTTPURLResponse new];
+        id mockResponse = OCMPartialMock(response);
+        //[MSHttpTestUtil stubHttp404Response];
```
jaelim-ms: Remove this one.

PR #1462, opened 2019-04-03
Clear token on error and cancel actions

MatkovIvan:
* [x] Did you add unit tests?

## Description

Wait token to expire (can modify expiration in code to avoid waiting 1 hour)
Track event and cause sign in refresh to fail or cancel (can simulate failure by modifying code as it will likely succeed most of the time).
=> Channel blocked.

Expected behavior: unblock channel sending and send logs anonymously.

### DIFF
--- a/AppCenterIdentity/AppCenterIdentity/MSIdentity.m
+++ b/AppCenterIdentity/AppCenterIdentity/MSIdentity.m
@@ -372,6 +372,7 @@ static dispatch_once_t onceToken;
                                 completionBlock:^(MSALResult *result, NSError *e) {
                                   typeof(self) strongSelf = weakSelf;
                                   if (e) {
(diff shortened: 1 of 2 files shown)


AppCenterIdentity/AppCenterIdentity/MSIdentity.m
```diff
@@ -372,6 +372,7 @@ - (void)acquireTokenInteractively {
                                 completionBlock:^(MSALResult *result, NSError *e) {
                                   typeof(self) strongSelf = weakSelf;
                                   if (e) {
+                                    [[MSAuthTokenContext sharedInstance] setAuthToken:nil withAccountId:nil expiresOn:nil];
```
guperrot: We need that in silent sign in as this is the one used for refresh, here the method is called interactive so that might not be the flow used during refresh.
This is good to have as well but should not be enough for the reported bug.

PR #1466, opened 2019-04-04
Change account Id verification for unit test: testSetAuthTokenDoesTriggerNewUserOnNewAccount

jinquan2019:
Change the account Id verification for second setting authtoken with new user account


jaelim-ms:
@dhei That's because the method got called twice with two different values. Previous implementation just didn't verify `expectedAccountId2`. Unfortunately there is no `times` supported but we can increment counter in stub when it called and verify the counter at the end.
However, it isn't a big deal since this will verify both.

PR #1480, opened 2019-04-06
Reset appcenter shared instance

guperrot:

* [ ] Did you add unit tests?

## Description

This should be done since MSAppCenter is configured in those tests.


AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 #import "MSAppCenter.h"
```
Jamminroot: I think this line can be removed since we use `MSAppCenterPrivate.h`

clpolet: They are actually not importing each others.

PR #1496, opened 2019-04-11
Fix build errors with Xcode 10.2

jinquan2019:

* [ ] Did you add unit tests?


AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
```diff
@@ -7,12 +7,12 @@
...
-@property(nonatomic, copy, readonly) NSString *partition;
-@property(nonatomic, readonly) Class documentType;
-@property(nonatomic, readonly) MSReadOptions *readOptions;
+@property(nonatomic, copy, readonly, nullable) NSString *partition;
+@property(nonatomic, readonly, nullable) Class documentType;
+@property(nonatomic, readonly, nullable) MSReadOptions *readOptions;
 
 // Read-write (to implement pagination).
-@property(nonatomic) MSPage *currentPage;
+@property(nonatomic, nonnull) MSPage *currentPage;
```
jwallra: I would prefer to use `NS_ASSUME_NONNULL_BEGIN`/`END` as a convention instead of adding nonnull individually.  @clpolet ?

clpolet: Yes definitely, please use the `NS_ASSUME_NONNULL_BEGIN/END` pattern.

Jamminroot:
Since there's `NS_ASSUME_NONNULL_BEGIN` added, please make sure that there are nil-checks in files which import those headers and refer to properties.

jinquan2019:
> Since there's `NS_ASSUME_NONNULL_BEGIN` added, please make sure that there are nil-checks in files which import those headers and refer to properties.

Yes, I have done the nil-checks in files and there is no problem

jaelim-ms:
The change looks good but there are some unit test failures in macOS (w/ Xcode 10.2) which might be caused by this change. Marking don't merge for the PR until we figure out that causes the issue.

AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
```diff
@@ -4,15 +4,19 @@
...
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface MSPaginatedDocuments ()
 
 // Read-only.
-@property(nonatomic, copy, readonly) NSString *partition;
-@property(nonatomic, readonly) Class documentType;
-@property(nonatomic, readonly) MSReadOptions *readOptions;
+@property(nonatomic, copy, readonly, nullable) NSString *partition;
```
clpolet: This should be non nullable. The change will need to be propagated.
```suggestion
@property(nonatomic, copy, readonly) NSString *partition;
```

AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
```diff
@@ -4,15 +4,19 @@
...
+@property(nonatomic, readonly, nullable) Class documentType;
```
clpolet: This one also should be non nullable. The change will need to be propagated.
```suggestion
@property(nonatomic, readonly) Class documentType;
```

clpolet:
Also @jinquan2019 please merge develop into this branch.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTestUtil.m
```diff
@@ -47,6 +47,10 @@ + (BOOL)copyFixtureCrashReportWithFileName:(NSString *)filename {
     return NO;
 
   NSError *error = nil;
+  NSString *desPath = [plcrCrashesDir stringByAppendingPathComponent:@"live_report.plcrash"];
```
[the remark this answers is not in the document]
jinquan2019: At Xcode 10.2, it couldn't copy the `live_report.plcrash` file into crash cache path: com.plausiblelabs.crashreporter.data when the file with same name exists in the cache path.
It will occur the issue: 
image
But at Xcode 9.x, the copy action can work well. 

So I add these codes as the workaround to fix this build error.
Do you have any other ideas for it

AppCenterDataStorage/AppCenterDataStorage/Internal/Model/MSPaginatedDocumentsInternal.h
jinquan2019: I see MSPaginatedDocuments class has a contrstuer which will pass nil value to these two properties, so I set them as nullable.
image
For your comment, I should pass non null value from them, right?

PR #1499, opened 2019-04-11
Fix issue where readonly partition lacked unique columns constraint

achocron:

* [x] Did you add unit tests?

## Description
The readonly storage table was not being created with the unique columns constraint that all tables require. As a result, when a document in the readonly table was replaced (usually to update TTL), it was actually duplicated instead of replaced.

### DIFF
--- a/AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
+++ b/AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
@@ -10,15 +10,33 @@
 @implementation MSDBStorage
 
 - (instancetype)initWithSchema:(MSDBSchema *)schema version:(NSUInteger)version filename:(NSString *)filename {
-  if ((self = [self initWithVersion:version filename:filename])) {
+  if ((self = [super init])) {
+    BOOL newDatabase = ![MSUtility fileExistsForPathComponent:filename];
+    _dbFileURL = [MSUtility createFileAtPathComponent:filename withData:nil atomically:NO forceOverwrite:NO];
+    _maxSizeInBytes = kMSDefaultDatabaseSizeInBytes;
+
+    // If it is custom SQLite library we need to turn on URI filename capability.
+    sqlite3_config(SQLITE_CONFIG_URI, 1);
     int result;
     sqlite3 *db = [MSDBStorage openDatabaseAtFileURL:self.dbFileURL withResult:&result];
     if (db) {
+      _pageSize = [MSDBStorage getPageSizeInOpenedDatabase:db];
+      NSUInteger databaseVersion = [MSDBStorage versionInOpenedDatabase:db];
 
-      // Create tables based on schema.
+      // Create table
       [MSDBStorage createTablesWithSchema:schema inOpenedDatabase:db];
+      if (newDatabase) {
+        MSLogInfo([MSAppCenter logTag], @"Created \"%@\" database with %lu version.", filename, (unsigned long)version);
+        [self customizeDatabase:db];
+      } else if (databaseVersion < version) {
+        MSLogInfo([MSAppCenter logTag], @"Migrate \"%@\" database from %lu to %lu version.", filename, (unsigned long)databaseVersion,
+                  (unsigned long)version);
+        [self migrateDatabase:db fromVersion:databaseVersion];
+      }
+      [MSDBStorage enableAutoVacuumInOpenedDatabase:db];
+      [MSDBStorage setVersion:version inOpenedDatabase:db];
       sqlite3_close(db);
-    }
+    };
   }
   return self;
 }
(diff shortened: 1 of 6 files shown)


yochaude:
What was missing for the upsert not to work? I cannot really tell from this PR.

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -1061,13 +1061,20 @@ - (void)testErrorDeletingOldestLog {
...
+  [self.storageTestUtil deleteDatabase];
+
   // When
   self.sut = [MSLogDBStorage new];
 
   // Then
   NSString *currentTable =
       [self.sut executeSelectionQuery:[NSString stringWithFormat:@"SELECT sql FROM sqlite_master WHERE name='%@'", kMSLogTableName]][0][0];
   assertThat(currentTable, is(kMSLatestSchema));
+//  NSArray *result =
```
jwallra: Remove commented code... Actually why were we saving the result before?  It doesn't look like it's used

YiMysty: This is the test code, I will remove it!

YiMysty: removed! :D

AppCenter/AppCenterTests/MSLogDBStorageTests.m
```diff
@@ -1061,6 +1061,9 @@ - (void)testErrorDeletingOldestLog {
 
 - (void)testCreateFromLatestSchema {
 
+  // If
+  [self.storageTestUtil deleteDatabase];
```
dhei: `[self.storageTestUtil deleteDatabase]` is already called in [`(void)tearDown`](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1499/files#diff-ae42d5f18ad4a9fc0a4426cbbb33eb98R57), why duplicate here?

AppCenterDataStorage/AppCenterDataStorageTests/MSDBDocumentStoreTests.m
```diff
@@ -202,6 +200,29 @@ - (void)testReadUserDocumentFromLocalDatabaseNotFound {
   XCTAssertEqualObjects(documentWrapper.documentId, documentId);
 }
 
+- (void)testUpsertReplacesCorrectlyInAppStorage {
```
dhei: Test case suggestion:

On Android we have a test case (https://github.com/Microsoft/AppCenter-SDK-Android/blob/b563fc68e6919d2a20dd12fd367ca4c080b1f7f9/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java#L498) that verifies `replace` called twice with insertion the first time and update the second time. 

I haven't seen any other test methods in this class verifying `replace` called only once, so maybe we do the same test as Android? 😄

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -11,21 +11,48 @@ @implementation MSDBStorage
...
       _pageSize = [MSDBStorage getPageSizeInOpenedDatabase:db];
+      NSUInteger databaseVersion = [MSDBStorage versionInOpenedDatabase:db];
+
+      // Create table
+      [MSDBStorage createTablesWithSchema:schema inOpenedDatabase:db];
+      if (newDatabase) {
+        MSLogInfo([MSAppCenter logTag], @"Created \"%@\" database with %lu version.", filename, (unsigned long)version);
+        [self customizeDatabase:db];
+      } else if (databaseVersion < version) {
+        MSLogInfo([MSAppCenter logTag], @"Migrate \"%@\" database from %lu to %lu version.", filename, (unsigned long)databaseVersion,
```
jwallra:
```suggestion
        MSLogInfo([MSAppCenter logTag], @"Migrating \"%@\" database from version %lu to %lu.", filename, (unsigned long)databaseVersion,
```

AppCenterDataStorage/AppCenterDataStorageTests/MSDBDocumentStoreTests.m
```diff
@@ -202,6 +200,37 @@ - (void)testReadUserDocumentFromLocalDatabaseNotFound {
...
+
+  // If
+  MSDocumentWrapper *expectedDocumentWrapper = [MSDocumentUtils documentWrapperFromData:[self jsonFixture:@"validTestDocument"]
+                                                                           documentType:[MSDictionaryDocument class]];
+  MSWriteOptions *writeOptions = [[MSWriteOptions alloc] initWithDeviceTimeToLive:MSDataStoreTimeToLiveInfinite];
+
+  // When
+  // Upsert twice to ensure that replacement is correct.
```
jwallra: This is no longer upserting twice in this part, comment should be corrected

achocron: Just meant in general we do it twice, not necessarily in that block i guess

PR #1500, opened 2019-04-11
Feature/storage read options

achocron:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

This updates the TTL based on the read options when reading a document from the cache.

PR #1503, opened 2019-04-12
Change default device time-to-live from 1 hour to 1 day

dhei:

* [ ] Did you add unit tests?

## Description

The default device time-to-live is 1 hour which is not too short. We're going to use 1 day as the default value.

### DIFF
--- a/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
+++ b/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
@@ -35,7 +35,7 @@ static NSString *const MSDataStoreUserDocumentsPartition = @"user";
 static NSString *const MSDataStoreAppDocumentsPartition = @"readonly";
 
 /**
- * Cache does not expire.
+ * No expiration on cache.
  */
 static int const MSDataStoreTimeToLiveInfinite = -1;
 
@@ -45,7 +45,7 @@ static int const MSDataStoreTimeToLiveInfinite = -1;
 static int const MSDataStoreTimeToLiveNoCache = 0;
 
 /**
- * Default cahcing value of one day.
+ * Default expiration on cache.
  */
 static int const MSDataStoreTimeToLiveDefault = 60 * 60 * 24;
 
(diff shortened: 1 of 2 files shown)


AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
```diff
@@ -35,11 +35,19 @@ static NSString *const MSDataStoreUserDocumentsPartition = @"user";
 static NSString *const MSDataStoreAppDocumentsPartition = @"readonly";
 
 /**
- * Time to live constants
+ * Cache does not expire.
  */
 static int const MSDataStoreTimeToLiveInfinite = -1;
+
+/**
+ * Do not cache.
```
jaelim-ms: Why do we have TTL value for No Cache? It could have *never expire* but I don't think it makes sense to have *no cache*. Any use cases?

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
```diff
@@ -35,11 +35,19 @@ static NSString *const MSDataStoreUserDocumentsPartition = @"user";
 static NSString *const MSDataStoreAppDocumentsPartition = @"readonly";
 
 /**
- * Time to live constants
+ * Cache does not expire.
```
jaelim-ms: It isn't a strong suggestion because I couldn't think of better doc for the constant. Maybe
```suggestion
 * No expiration on cache.
```

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
dhei: @jaelim-ms PR #1500 now implements `read` API to check ReadOptions's TTL for NoCache and remove cache if needed. (see https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1500/files#diff-b33d4fad61841860c5d0921184064395R169)

Note that this PR is based on #1500.

AppCenterDataStorage/AppCenterDataStorage/MSDataStore.h
```diff
@@ -35,11 +35,19 @@ static NSString *const MSDataStoreUserDocumentsPartition = @"user";
...
+/**
+ * Do not cache.
+ */
 static int const MSDataStoreTimeToLiveNoCache = 0;
-static int const MSDataStoreTimeToLiveDefault = 60 * 60;
+
+/**
+ * Default cahcing value of one day.
+ */
+static int const MSDataStoreTimeToLiveDefault = 60 * 60 * 24;
```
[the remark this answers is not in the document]
dhei: Let's fix that in a different PR to avoid merge conflicts.

PR #1513, opened 2019-04-17
Pass an error when Cosmos DB returns error response

jaelim-ms:
* [X] Did you add unit tests?

## Description

HTTP request to Cosmos DB returns an error code without an error object which causes unexpected process in Data module. Add code to instantiate an error object with details in this case.

### DIFF
--- a/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
+++ b/AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
@@ -11,6 +11,7 @@
 #import "MSCosmosDb.h"
 #import "MSDBDocumentStore.h"
 #import "MSDataSourceError.h"
+#import "MSDataStorageConstants.h"
 #import "MSDataStoreErrors.h"
 #import "MSDataStoreInternal.h"
 #import "MSDataStorePrivate.h"
@@ -18,6 +19,7 @@
 #import "MSDocumentUtils.h"
 #import "MSDocumentWrapperInternal.h"
 #import "MSHttpClient.h"
+#import "MSHttpUtil.h"
 #import "MSPaginatedDocuments.h"
 #import "MSReadOptions.h"
 #import "MSServiceAbstractProtected.h"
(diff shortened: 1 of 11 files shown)


AppCenterDataStorage/AppCenterDataStorage/MSDataStore.m
```diff
@@ -469,7 +471,22 @@ - (void)performCosmosDbOperationWithPartition:(NSString *)partition
                                                                               httpMethod:httpMethod
                                                                                     body:body
                                                                        additionalHeaders:additionalHeaders
-                                                                       completionHandler:completionHandler];
+                                                                       completionHandler:^(NSData *_Nullable data,
+                                                                                           NSHTTPURLResponse *_Nullable response,
+                                                                                           NSError *_Nullable cosmosDbError) {
+                                                                         if (!cosmosDbError &&
+                                                                             ![MSHttpUtil isSuccessStatusCode:response.statusCode]) {
```
guperrot: I see only 1 new test but for 4 code branches here.

jaelim-ms: All other branches are covered by other tests already.
This is handling edge case and the unit test tests the edge case.

guperrot: ok

AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -520,6 +519,55 @@ - (void)testPerformCosmosDbAsyncOperationWithHttpClient {
...
+                                completionHandler:^(NSData *_Nullable responseBody, NSHTTPURLResponse *_Nullable response,
+                                                    NSError *_Nullable error) {
+                                  XCTAssertNil(responseBody);
+                                  actualResponse = response;
+                                  actualError = error;
+                                  [expectation fulfill];
+                                }];
+
+  // Then
+  [self waitForExpectationsWithTimeout:100
```
yochaude: Maybe use less than 100 here so the tests does not hang too long when it breaks 😬 ?

jaelim-ms: Oops.. That was for my test. 😓

PR #1523, opened 2019-04-19
Upgrade sdk and test apps with swift 5.0 and ios12.0

jinquan2019:
Upgrade sdk and test apps with swift 5.0, iOS12.0, macOS 10.14


Config/Global.xcconfig
```diff
@@ -2,13 +2,13 @@
...
+MS_SIM_ARCHS = x86_64
 SDKROOT = iphoneos
 
 // :Mark: Deployment
-IPHONEOS_DEPLOYMENT_TARGET = 9.0
-TVOS_DEPLOYMENT_TARGET = 10.0
-MACOSX_DEPLOYMENT_TARGET = 10.9
+IPHONEOS_DEPLOYMENT_TARGET = 12.0
+TVOS_DEPLOYMENT_TARGET = 12.0
+MACOSX_DEPLOYMENT_TARGET = 10.14
```
MatkovIvan: I guess that's a wrong change

clpolet:
Cancelling this, upgrading Swift will come back later. Also upgrading to Swift 4 should be enough but we'll see that later.

PR #1525, opened 2019-04-19
Fix infinite recursion in Storage.replace call

annakocheshkova:
* [ ] Did you add unit tests?

## Description

One picture is worth a thousand words...
image

### DIFF
--- a/AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
+++ b/AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
@@ -300,7 +300,6 @@ static NSString *const kMSDocumentIdTest = @"documentId";
                                }];
 }
 
-
 - (void)testReplaceWhenDataModuleDisabled {
 
   // If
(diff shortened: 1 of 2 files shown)


AppCenterDataStorage/AppCenterDataStorageTests/MSDataStoreTests.m
```diff
@@ -243,6 +243,63 @@ - (void)testCreateWhenDataModuleDisabled {
...
+  [MSDataStore createWithPartition:kMSPartitionTest
+                        documentId:kMSDocumentIdTest
+                          document:mockSerializableDocument
+                 completionHandler:^(MSDocumentWrapper *data) {
+                   completionHandlerCalled = YES;
+                   actualDocumentWrapper = data;
+                   [expectation fulfill];
+                 }];
+  id<MSSerializableDocument> replaceMockSerializableDocument = [MSFakeSerializableDocument new];
+  [MSDataStore replaceWithPartition:kMSPartitionTest
```
dhei: I don't see a test case covering the replace API which takes `writeOptions:` as parameter. Please consider adding a test for it.

annakocheshkova: It is called implecitly with nil options. But I can add a test case for writeOptions, too :)

yochaude: Let's not block the PR on that, the read/write stuff is being tested at the data proxy level already.

Thanks for catching that Anna!

jaelim-ms: We decided to unblock this.

PR #1555, opened 2019-04-29
Fix/use msalb2c authority instead of msal authority

jinquan2019:

* [ ] Did you add unit tests?

## Description

Once the Authority is a B2C Authority then use `MSALB2CAuthority` class instead of the generic MSALAuthority


AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -298,16 +299,26 @@ - (void)configAuthenticationClient {
 
   // Init MSAL client application.
   NSError *error;
-  MSALAuthority *auth = [MSALAuthority authorityWithURL:(NSURL * __nonnull) self.authConfig.authorities[0].authorityUrl error:nil];
+  NSURL *authUrl = nil;
+  for (MSALAuthority *auth in self.authConfig.authorities) {
```
jaelim-ms: The `authorities` in `authConfig` is `MSAuthority` type which is defined in App Center.
You can't cast it to `MSALAuthority`, this is dead code and `authUrl` will always be `nil`.
The task is just using `MSALB2CAuthority` instead of `MSALAuthority` so you just need to replace the type from `MSALAuthority` to `MSALB2CAuthority`. @clpolet FYI.

AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -298,13 +299,13 @@ - (void)configAuthenticationClient {
...
   NSError *error;
-  MSALAuthority *auth = [MSALAuthority authorityWithURL:(NSURL * __nonnull) self.authConfig.authorities[0].authorityUrl error:nil];
+  MSALB2CAuthority *auth = [[MSALB2CAuthority alloc] initWithURL:(NSURL * __nonnull)self.authConfig.authorities[0].authorityUrl error:nil];
   MSALPublicClientApplicationConfig *config =
       [[MSALPublicClientApplicationConfig alloc] initWithClientId:(NSString * __nonnull) self.authConfig.clientId
                                                       redirectUri:self.authConfig.redirectUri
                                                         authority:auth];
   if (!auth) {
-    MSLogError([MSAuth logTag], @"Auth config doesn't contain a valid authority.");
+    MSLogError([MSAuth logTag], @"Auth config doesn't contain a valid B2C authority.");
```
clpolet:
```suggestion
    MSLogError([MSAuth logTag], @"Auth config doesn't contain a valid default B2C authority.");
```

PR #1560, opened 2019-04-29
Changed the Error constructor signature

MehrdadMzfr:

* [ ] Did you add unit tests?

## Description

Changed the Error constructor signature

### DIFF
--- a/AppCenterData/AppCenterData/Internal/Client/MSCosmosDb.m
+++ b/AppCenterData/AppCenterData/Internal/Client/MSCosmosDb.m
@@ -202,7 +202,7 @@ static NSString *const kMSHeaderMsDate = @"x-ms-date";
   }
 
   // Return the error.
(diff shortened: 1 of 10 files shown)


MehrdadMzfr:
@jwallra this addresses leftover comments from previous PR. the Refactoring is already merged and this PR is only the method name change, which is not a risk.

jwallra:
@MehrdadMzfr Could you take a look at the board item I linked?  I still don't think we should bother with the renaming but if we're going to do it we should improve it to incorporate the changes suggested in the work item.  Here's the link because it didn't link correctly before: https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/61019

jwallra:
@MehrdadMzfr That's my bad, I was thinking of MSDocumentWrapper. 😞  I'll unblock

PR #1561, opened 2019-04-30
Fix different processing of json values offline and online

guperrot:

* [ ] Did you add unit tests?

## Description

Fixes for #60954 and #60953 where json was inconsistent between offline and online scenarios.
The json value must always be customer content in sqlite and in memory objects.

### DIFF
--- a/AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
+++ b/AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
@@ -124,7 +124,7 @@ static NSString *const kMSDocumentKey = @"document";
       ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSDocumentKey keyType:[NSDictionary class]]) {
 
     // Prepare and return error.
-    NSString *errorMessage = @"Can't deserialize document (missing system properties or partition key)";
+    NSString *errorMessage = @"Can't deserialize document (missing system properties, partition key, or document)";
     MSDataError *dataError = [[MSDataError alloc] initWithErrorCode:MSACDataErrorJSONSerializationFailed
                                                          innerError:nil
                                                             message:errorMessage];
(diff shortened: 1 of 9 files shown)


guperrot:
Delete online broken. It's also indirectly related to remaining unit test failures.

```
2019-04-29 18:49:04.714641-0700 SasquatchPuppet[7420:1727737] [AppCenterData] ERROR: +[MSDocumentUtils documentWrapperFromDocumentData:documentType:eTag:lastUpdatedDate:partition:documentId:pendingOperation:fromDeviceCache:]/99 Error deserializing data: The data couldn’t be read because it isn’t in the correct format.
```

guperrot:
> Error deserializing data

That's on develop, nevermind: AB#61035

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
```diff
@@ -120,7 +120,8 @@ + (MSDocumentWrapper *)documentWrapperFromDictionary:(NSObject *)object
   if (![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSDocumentIdKey keyType:[NSString class]] ||
       ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSDocumentTimestampKey keyType:[NSNumber class]] ||
       ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSDocumentEtagKey keyType:[NSString class]] ||
-      ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSPartitionKey keyType:[NSString class]]) {
+      ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSPartitionKey keyType:[NSString class]] ||
+      ![MSDocumentUtils isReferenceDictionaryWithKey:object key:kMSDocumentKey keyType:[NSDictionary class]]) {
```
guperrot: That's not correct as delete has null customer payload.

guperrot: And this is what makes unit test fails.

guperrot:
Screen Shot 2019-04-29 at 7 52 14 PM

The test is actually right. On develop, even on error we get etag, partition and date being set.
This PR introduces a regression where we don't set these fields anymore because we fail early, the return error statements were actually not a good change @jwallra @achocron .

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
guperrot: Fixed in https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1561/commits/712173afaa8b24c169608caacbc00d922eae9a5a

guperrot:
@jwallra @achocron the remaining tests are failing for a good reason.
We return error too early now in utils, and thus some metadata is missing.
We need to keep populating those metadata on JSON error (so keep the asserts as is and fix utils code).

AppCenterData/AppCenterDataTests/MSDBDocumentStoreTests.m
```diff
@@ -73,7 +73,7 @@ - (void)testReadAppDocumentFromLocalDatabase {
   // If
   NSString *documentId = @"12829";
   NSString *eTag = @"398";
-  NSString *jsonString = @"{ \"document\": {\"key\": \"value\"}}";
```
YiMysty: Is a specific reason to change the test string here? :)

achocron: Yes, now we do not store the root json string in storage, just the json of the document property

AppCenterData/AppCenterDataTests/MSDocumentUtilsTests.m
```diff
@@ -152,7 +111,7 @@ - (void)testDocumentWrapperFromDictionaryWithDocument {
   XCTAssertTrue([[document eTag] isEqualToString:@"etag"]);
   XCTAssertNotNil([document lastUpdatedDate]);
   XCTAssertTrue([[document partition] isEqualToString:@"readonly"]);
-  XCTAssertNotNil([document jsonValue]);
+  XCTAssertNil([document jsonValue]);
```
jaelim-ms: Why should it be `nil`? Was the test wrong before?

achocron: Yeah now there is no metadata in the json value, just the document itself. It was inconsistent before.

guperrot: This is nil on document parsing error now. Before you could have deserialized value non nil and failed jsonValue being nil.

PR #1566, opened 2019-04-30
Deletion p0 issue

MehrdadMzfr:
there are 2 bug fixes as part of this PR:
1-Deletion operation was not performed on the local DB
2-when document is read from local DB, the pending operation value is NSNull (not nil)

* [ ] Did you add unit tests?

## Description

A few sentences describing the overall goals of the pull request.

## Related PRs or issues

## Misc

### DIFF
--- a/AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
+++ b/AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
@@ -69,20 +69,23 @@
     if ([self shouldAttemptRemoteOperationForDocument:cachedDocument]) {
       MSLogInfo([MSData logTag], @"Performing remote operation");
       remoteDocumentBlock(^(MSDocumentWrapper *_Nonnull remoteDocument) {
-
         // If a valid remote document was retrieved, update local store
         if (remoteDocument.error == nil) {
-          
-          // For delete operation, delete the document from local cache.
-          // Otherewise, for other online scenarios, the intended pending operation in the local store is nil.
-          NSString *localStorageOperation = [kMSPendingOperationDelete isEqualToString:(NSString *)operation] ? kMSPendingOperationDelete : nil;
-          
-          // Perform local storage operation.
-          [self updateLocalStore:token
-              currentCachedDocument:cachedDocument
-                  newCachedDocument:remoteDocument
-                   deviceTimeToLive:deviceTimeToLive
-                          operation:localStorageOperation];
+
+          // If operation is delete, directly delete the document from the local cache.
+          if ([kMSPendingOperationDelete isEqualToString:(NSString *)operation]) {
+            MSLogInfo([MSData logTag], @"Delete the document from local storage (partition: %@, id: %@)", token.partition,
+                      cachedDocument.documentId);
+            [self.documentStore deleteWithToken:token documentId:cachedDocument.documentId];
+          }
+          // For other online scenarios, the intended pending operation in the local store is nil.
+          else {
+            [self updateLocalStore:token
+                currentCachedDocument:cachedDocument
+                    newCachedDocument:remoteDocument
+                     deviceTimeToLive:deviceTimeToLive
+                            operation:nil];
+          }
         }
         completionHandler(remoteDocument);
       });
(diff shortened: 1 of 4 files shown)


AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -69,14 +69,20 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+
         // If a valid remote document was retrieved, update local store
         if (remoteDocument.error == nil) {
+          
+          // For delete operation, delete the document from local cache.
+          // Otherewise, for other online scenarios, the intended pending operation in the local store is nil.
+          NSString *localStorageOperation = [kMSPendingOperationDelete isEqualToString:(NSString *)operation] ? kMSPendingOperationDelete : nil;
+          
+          // Perform local storage operation.
           [self updateLocalStore:token
```
yochaude: This will not delete the operation from the cache. It will only create or update and entry and set the pending operation to DELETE or nil for other operations.

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -239,17 +245,21 @@ - (void)updateLocalStore:(MSTokenResult *)token
     [self.documentStore deleteWithToken:token documentId:currentCachedDocument.documentId];
   }
 
-  /*
-   * If the cached document has a create or replace pending operation, and no eTags, and if the current operation is a
-   * deletion, delete the document from the store.
-   */
+   //If the cached document has a create or replace pending operation, and no eTags, delete the document from the store.
```
yochaude: So if I do two offline updates, would that delete the document from local storage? We need the current operation to be a deletion in order to delete the local document from storage.

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -239,17 +245,21 @@ - (void)updateLocalStore:(MSTokenResult *)token
...
             [kMSPendingOperationReplace isEqualToString:currentCachedDocument.pendingOperation]) &&
-           !currentCachedDocument.eTag && operation && [kMSPendingOperationDelete isEqualToString:(NSString *)operation]) {
+           !currentCachedDocument.eTag) {
     MSLogInfo([MSData logTag], @"Removing never-synced document from local storage (partition: %@, id: %@)", token.partition,
               currentCachedDocument.documentId);
     [self.documentStore deleteWithToken:token documentId:currentCachedDocument.documentId];
   }
+  
+  // If operation is delete, delete the document from the local storage.
+  else if ([kMSPendingOperationDelete isEqualToString:(NSString *)operation]) {
```
yochaude: This might prevent offline delete to work?

When offline, we want a deletion to place a record in the local document store.

MehrdadMzfr: valid point, fixing

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
jwallra: Right, we need to keep the last condition and add the new one like you did below.

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -69,14 +69,20 @@ - (void)performOperation:(NSString *_Nullable)operation
     if ([self shouldAttemptRemoteOperationForDocument:cachedDocument]) {
       MSLogInfo([MSData logTag], @"Performing remote operation");
       remoteDocumentBlock(^(MSDocumentWrapper *_Nonnull remoteDocument) {
+
         // If a valid remote document was retrieved, update local store
         if (remoteDocument.error == nil) {
+          
+          // For delete operation, delete the document from local cache.
+          // Otherewise, for other online scenarios, the intended pending operation in the local store is nil.
```
achocron: "Otherewise" -> "Otherwise", also not a blocker but we usually use '/*...*/' for multiline comments on iOS.

PR #1570, opened 2019-05-01
For an offline read, we should maintain the existing pending operatio…

yochaude:

* [x] Did you add unit tests?

## Description

On an offline read, we should maintain the existing pending operation that exists in the local storage.

## Related PRs or issues

## Misc

### DIFF
--- a/AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
+++ b/AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
@@ -392,6 +392,59 @@
                                }];
 }
 
+- (void)testLocalReadWhenCachedDocumentMaintainPendingOperation {
+
+  // If
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Completed with local read."];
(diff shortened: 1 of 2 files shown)


AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
```diff
@@ -392,6 +392,59 @@ - (void)testReadOperationFailsWhenPendingDelete {
...
+                                                                                                jsonValue:@""
+                                                                                                partition:@"partition"
+                                                                                               documentId:@"documentId"
+                                                                                                     eTag:@""
+                                                                                          lastUpdatedDate:nil
+                                                                                         pendingOperation:kMSPendingOperationReplace
+                                                                                                    error:nil
+                                                                                          fromDeviceCache:YES];
+  __block MSDocumentWrapper *wrapper;
+  OCMStub([self.documentStoreMock readWithToken:OCMOCK_ANY documentId:OCMOCK_ANY documentType:OCMOCK_ANY]).andReturn(cachedDocumentWrapper);
```
jaelim-ms: nit: It would be better to minimize stub scope to specific parameter values. Not a blocker.

PR #1586, opened 2019-05-03
Use `extern` instead of `static const` to avoid code duplication

dhei:

* [ ] Did you add unit tests?

## Description

Use `extern` whenever possible to avoid code duplication.

Fixes [AB#60566](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/60566)


jaelim-ms:
Waiting for release + we need to talk about this change.

jaelim-ms:
Please remove all `extern`s and move static constant definitions to header. Then you can import the header and use those in test classes.

PR #1593, opened 2019-05-03
Bug 61019: [iOS] Errors should include partition and nullable eTag

michaelsmitha:

* [x] Did you add unit tests?

## Description

Document partition is usually known in document errors, and eTag is sometimes known. We should include these in the error when available.

Well formed documents should never have an error, so we've removed that parameter.

## Related PRs or issues

[AB#61019](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61019)


AppCenterData/AppCenterData/Internal/MSDocumentWrapperInternal.h
```diff
@@ -27,18 +26,55 @@
...
 /**
  * Initialize a `MSDocumentWrapper` instance.
  *
  * @param error Document error.
  * @param documentId Document Id.
+ * @param partition Partition key.
+ *
+ * @return A new `MSDocumentWrapper` instance.
+ */
+- (instancetype)initWithError:(MSDataError *)error documentId:(NSString *)documentId partition:(NSString *)partition;
```
achocron: Do we need this initializer when the one below exists? Can just pass `nil` for eTag if it doesn't exist or isn't known.

AppCenterData/AppCenterData/Internal/MSDocumentWrapperInternal.h
```diff
@@ -27,18 +26,55 @@
...
+ * @param documentId Document id.
+ * @param eTag Document eTag.
+ * @param lastUpdatedDate Last updated date of the document.
+ * @param pendingOperation The name of the pending operation, or nil.
+ * @param fromDeviceCache Flag indicating if the document wrapper was retrieved remotely or not.
+ * @param error Document error.
+ *
+ * @return A new `MSDocumentWrapper` instance.
+ */
+- (instancetype)initWithDeserializedValue:(id<MSSerializableDocument>)deserializedValue
```
achocron: Is this initializer needed? When there's an error, one of the initializers below should suffice.

dhei: Good point, will do.

AppCenterData/AppCenterData/Internal/MSDocumentWrapperInternal.h
michaelsmitha: @achocron is this just a convention that we prefer is to pass `nil` instead of overloading methods?

dhei: @achocron, we add this method overload to avoid changing everywhere that invokes this initializer. This initializer is called in 10+ places in SDK (excluding tests) but the below initializer is called in only 1 place. 

Any other reason against adding this method overload?

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
```diff
@@ -163,15 +163,18 @@ + (MSDocumentWrapper *)documentWrapperFromDictionary:(NSDictionary *)dictionary
...
-                                                    jsonValue:jsonValue
-                                                    partition:partition
-                                                   documentId:documentId
-                                                         eTag:eTag
-                                              lastUpdatedDate:lastUpdatedDate
-                                             pendingOperation:pendingOperation
-                                                        error:dataError
-                                              fromDeviceCache:fromDeviceCache];
+  if (dataError) {
+    return [[MSDocumentWrapper alloc] initWithError:dataError documentId:documentId partition:partition eTag:eTag];
```
dhei: Let's not drop `lastUpdatedDate`, `pendingOperation` and `fromDeviceCache` in the `MSDocumentWrapper` object returned here. Let's extends the initializer of `initWithError:` to take those 3 parameters.

AppCenterData/AppCenterData/Internal/MSDocumentWrapperInternal.h
```diff
@@ -27,18 +26,19 @@
...
  *
  * @param error Document error.
  * @param documentId Document Id.
+ * @param partition Partition key.
+ * @param eTag Document eTag.
  *
  * @return A new `MSDocumentWrapper` instance.
  */
-- (instancetype)initWithError:(MSDataError *)error documentId:(NSString *)documentId;
+- (instancetype)initWithError:(MSDataError *)error documentId:(NSString *)documentId partition:(NSString *)partition eTag:(NSString *)eTag;
```
jaelim-ms: Is the order of the parameters reviewed/finalized with discussion? Seems it should be like
```suggestion
- (instancetype)initWithError:(MSDataError *)error partition:(NSString *)partition documentId:(NSString *)documentId eTag:(NSString *)eTag;
```

michaelsmitha: @jaelim-ms good idea, that'll be more consistent. Looks like I'll be adding more parameters anyway, and I'll make those consistent with the order of `initWithDeserializedValue`

clpolet: FYI The order might change in the future but +1 for consistency. It might change because `initWithDeserializedValue` was introduce at the time when the `read`, `create`, `replace` APIs from MSData.h first parameter was a partition.

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
```diff
@@ -65,7 +65,7 @@ + (MSDocumentWrapper *)documentWrapperFromData:(nullable NSData *)data
     }
     MSLogError([MSData logTag], @"Error deserializing data: %@", [error description]);
     MSDataError *dataError = [[MSDataError alloc] initWithErrorCode:MSACDataErrorJSONSerializationFailed innerError:error message:nil];
-    return [[MSDocumentWrapper alloc] initWithError:dataError documentId:nil];
+    return [[MSDocumentWrapper alloc] initWithError:dataError documentId:nil partition:nil eTag:nil];
```
clpolet: We might actually know the documentId and partition here. You would need to pass them to this method but look at where `documentWrapperFromData: documentType:fromDeviceCache:` is called, they are accessible : https://github.com/microsoft/AppCenter-SDK-Apple/blob/99417bf1de17f5df50890eabe1988ff37279007a/AppCenterData/AppCenterData/MSData.m#L520

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
```diff
@@ -129,7 +129,7 @@ + (MSDocumentWrapper *)documentWrapperFromDictionary:(NSObject *)object
                                                          innerError:nil
                                                             message:errorMessage];
     MSLogError([MSData logTag], @"Error deserializing data: %@.", [dataError localizedDescription]);
-    return [[MSDocumentWrapper alloc] initWithError:dataError documentId:nil];
+    return [[MSDocumentWrapper alloc] initWithError:dataError documentId:nil partition:nil eTag:nil];
```
clpolet: Same here the caller of `documentWrapperFromDictionary:documentType:fromDeviceCache` knows about the partition, if it's important we can pass it through.

AppCenterData/AppCenterDataTests/MSDocumentUtilsTests.m
```diff
@@ -198,7 +198,7 @@ - (void)testDocumentWrapperFromDataDeserializationError {
   XCTAssertNil([document jsonValue]);
 }
 
-- (void)testDocumentWrapperFromDocumentDataDeserializationßError {
```
clpolet: Good catch, that's an interesting one :).

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
michaelsmitha: @dhei I've added another overload to take those parameters. That could change based on @achocron 's feedback.

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
michaelsmitha: @clpolet Making this change is resulting in a cascade of other required changes. Okay to break these requests in to a different item?

clpolet: Can be done in a separate PR :).

michaelsmitha:
@jaelim-ms @dhei thought we were okay with our current test coverage on this PR

AppCenterData/AppCenterData/Internal/Util/MSDocumentUtils.m
```diff
@@ -163,15 +163,24 @@ + (MSDocumentWrapper *)documentWrapperFromDictionary:(NSDictionary *)dictionary
...
-                                                    jsonValue:jsonValue
-                                                    partition:partition
-                                                   documentId:documentId
-                                                         eTag:eTag
-                                              lastUpdatedDate:lastUpdatedDate
-                                             pendingOperation:pendingOperation
-                                                        error:dataError
-                                              fromDeviceCache:fromDeviceCache];
+  if (dataError) {
+    return [[MSDocumentWrapper alloc] initWithError:dataError
```
dhei: We need a unit test for this code path.

michaelsmitha: @dhei pair after lunch? Don't have my head around how the tests work yet.

AppCenterData/AppCenterData/MSDocumentWrapper.m
```diff
@@ -27,17 +26,41 @@ - (instancetype)initWithDeserializedValue:(id<MSSerializableDocument>)deserializ
...
+}
+
+- (instancetype)initWithError:(MSDataError *)error partition:(NSString *)partition documentId:(NSString *)documentId eTag:(NSString *)eTag {
+  return [self initWithError:error
+                   partition:partition
+                  documentId:documentId
+                        eTag:eTag
+             lastUpdatedDate:nil
+            pendingOperation:nil
+             fromDeviceCache:nil];
```
dhei:
```suggestion
             fromDeviceCache:NO];
```

AppCenterData/AppCenterDataTests/MSDocumentUtilsTests.m
```diff
@@ -158,6 +159,36 @@ - (void)testDocumentWrapperFromDictionaryWithUnserializable {
   XCTAssertEqual([document documentId], @"document-id");
 }
 
+- (void)testDocumentWrapperFromDictionaryWithDataError {
+  NSMutableDictionary *documentDictionary = [NSMutableDictionary new];
```
jaelim-ms:
```suggestion

  // If
  NSMutableDictionary *documentDictionary = [NSMutableDictionary new];
```

AppCenterData/AppCenterDataTests/MSDocumentUtilsTests.m
```diff
@@ -158,6 +159,36 @@ - (void)testDocumentWrapperFromDictionaryWithUnserializable {
...
+  documentDictionary[@"shouldFail"] = [NSSet set];
+  NSMutableDictionary *dictionary = [NSMutableDictionary new];
+  dictionary[@"document"] = documentDictionary;
```
jaelim-ms: Above two lines could be like
```
  NSDictionary *dictionary = @{ @"document": documentDictionary };
```

AppCenterData/AppCenterDataTests/MSDocumentUtilsTests.m
```diff
@@ -158,6 +159,36 @@ - (void)testDocumentWrapperFromDictionaryWithUnserializable {
...
+                                                              pendingOperation:@"pendingOperation"
+                                                               fromDeviceCache:NO];
+
+  // Then
+  XCTAssertNotNil(document);
+  XCTAssertNotNil([document error]);
+  XCTAssertTrue([[document error] isKindOfClass:[MSDataError class]]);
+  XCTAssertNotNil([document error].innerError);
+  XCTAssertEqual([document error].innerError.code, MSACDataErrorJSONSerializationFailed);
+  XCTAssertEqualObjects(document.documentId, @"docId");
```
jaelim-ms: Use variables for the values and use it in `// When`

PR #1600, opened 2019-05-07
Update script to use V2 pipeline

jaelim-ms:
Improve merge build pipeline. 45 mins -> 25 mins.

### DIFF
--- a/vsts/publish-podspec.sh
+++ b/vsts/publish-podspec.sh
@@ -66,6 +66,17 @@ if [ "$mode" == "internal" ] || [ "$mode" == "test" ]; then
 
   if [ "$mode" == "test" ]; then
 
+    # Read publish version for current build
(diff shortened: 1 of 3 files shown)


jaelim-ms:
The merge requires build pipeline configuration change to use V2.

jaelim-ms:
Valid point @guperrot but we need to merge this one and switch the definition to new one to verify it.

vsts/publish-podspec.sh
```diff
@@ -66,6 +66,17 @@ if [ "$mode" == "internal" ] || [ "$mode" == "test" ]; then
 
   if [ "$mode" == "test" ]; then
 
+    # Read publish version for current build
+    version=$(cat $CURRENT_BUILD_VERSION_FILENAME)
+
+    # Exit if response doesn't contain an array
```
clpolet: Does it really contain an array, not just the version?

jaelim-ms: Copy-paste. 😄

PR #1603, opened 2019-05-08
Fix formatter script

annakocheshkova:

* [ ] Did you add unit tests?

## Description

Fix formatter script due to organization rename.

### DIFF
--- a/clang-format-changed-files.sh
+++ b/clang-format-changed-files.sh
@@ -5,7 +5,7 @@
 
 # This script runs clang_format on all .h, .m, and .mm files that have changes from develop branch.
 # Before running this, ensure that your local develop branch is up to date.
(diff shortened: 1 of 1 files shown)


clang-format-changed-files.sh
```diff
@@ -5,7 +5,7 @@
 
 # This script runs clang_format on all .h, .m, and .mm files that have changes from develop branch.
 # Before running this, ensure that your local develop branch is up to date.
-upstream=$(git remote -v | grep "Microsoft/" | awk '{print $1}' | uniq)
+upstream=$(git remote -v | grep "microsoft/" | awk '{print $1}' | uniq)
```
jaelim-ms: If we don't care character cases, please ignore cases by adding `-i` to `grep`.

PR #1609, opened 2019-05-13
Remove arm64e from iOS Release build for CrashLib

dhei:

* [ ] Did you add unit tests?

## Description

Created a new **CrashLib Release iOS.xcconfig** which inherits from **{root}/Config/iOS.xcconfig** but overrides iOS architectures to excludes `arm64e`.

Tested via:
- archiving with bitcode locally in Xcode 10.1 with iOS/macOS/tvOS.
- AC build with bitcode.

### DIFF
--- a/CrashLib/Config/CrashLib Debug.xcconfig
+++ b/CrashLib/Config/CrashLib Debug.xcconfig
@@ -1,6 +1,4 @@
 #include "./CrashLib.xcconfig"
 
-ARCHS[config=Debug][sdk=iphoneos*] = armv7 armv7s arm64 arm64e
-
 ENABLE_TESTABILITY = YES
 COPY_PHASE_STRIP = NO
--- a/CrashLib/Config/CrashLib Debug.xcconfig
+++ b/CrashLib/Config/CrashLib Debug.xcconfig
@@ -1,4 +1,6 @@
 #include "./CrashLib.xcconfig"
 
+ARCHS[config=Debug][sdk=iphoneos*] = armv7 armv7s arm64 arm64e
+
 ENABLE_TESTABILITY = YES
 COPY_PHASE_STRIP = NO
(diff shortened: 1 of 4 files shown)


CrashLib/Config/CrashLib Debug.xcconfig
```diff
@@ -1,4 +1,6 @@
 #include "./CrashLib.xcconfig"
 
+ARCHS[config=Debug][sdk=iphoneos*] = armv7 armv7s arm64 arm64e
```
jaelim-ms: The file is already for Debug, `[config=Debug]` doesn't seem it is necessary.

CrashLib/Config/CrashLib Release iOS.xcconfig
```diff
@@ -0,0 +1,3 @@
+#include "../../Config/iOS.xcconfig"
+
+ARCHS[sdk=iphoneos*] = $(PRE_XCODE_10_1_ARCHS)
```
jaelim-ms: Doesn't it require `VALID_ARCHS[sdk=iphoneos*] = $(ARCHS)`?

dhei: Yes it already does that, note it's including `{root}/Config/iOS.xcconfig` which defines `VALID_ARCHS[sdk=iphoneos*] = $(ARCHS)` (https://github.com/microsoft/appcenter-sdk-apple/blob/fix/arm64e-crash-lib/Config/iOS.xcconfig#L10).

jaelim-ms: My concern is that VALID_ARCHS is overwritten before you change ARCHS here.

clpolet: As soon as it's declared somewhere down the road it works, doesn't need to be redeclared. See the result processed by Xcode:
Screen Shot 2019-05-14 at 11 16 38

PR #1618, opened 2019-05-14
Feature/initialization vector

achocron:

* [x] Did you add unit tests?

## Description

Make encryption compliant with requirements. Use initialization vectors, add metadata to encrypted text, and expire keys on a yearly basis.


AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -171,4 +217,22 @@ + (NSData *)generateKeyWithTag:(NSString *)keyTag {
   return resultKey;
 }
 
++ (NSData *)generateInitializationVector {
```
Jamminroot: Can this can be parametrized and re-used in `generateAndSaveKeyWithTag` since the only difference is amount of bytes?
```obj-c
+ (NSData *)generateAndSaveKeyWithTag:(NSString* )keyTag {
 NSData *resultKey = [self generateInitializationVector:kMSEncryptionKeySize];

 // Save key to the Keychain.
 NSString *stringKey = [resultKey base64EncodedStringWithOptions:0];
 [MSKeychainUtil storeString:stringKey forKey:keyTag];
 return resultKey;
}

+ (NSData *)generateInitializationVector {
 return [self generateInitializationVector:kCCBlockSizeAES128];
}

+ (NSData *)generateInitializationVector:(NSUInteger)keySize {
```

achocron: We could, but this would slightly obfuscate things and there is not much similarity between the key and the initialization vectors at a conceptual level. It is more of a coincidence that they are generated the same way. Resolving this but feel free to reopen if you want to discuss further.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+
++ (NSString *)getCurrentKeyTag {
+  @synchronized(classLock) {
+    NSString *keyMetadata = [[MSUserDefaults shared] objectForKey:kMSEncryptionKeyMetadataKey];
+    if (!keyMetadata) {
+      [self rotateToNewKeyTag:kMSEncryptionKeyTagAlternate];
+      return kMSEncryptionKeyTagAlternate;
     }
+
+    // Format is {keyTag}/{expiration as iso}
```
achocron:
```suggestion
    // Format is {keyTag}/{expiration as iso}.
```

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
-  return result;
 }
 
-+ (void)deleteKeyWithTag:(NSString *)keyTag {
-  [MSKeychainUtil deleteStringForKey:keyTag];
++ (void)rotateToNewKeyTag:(NSString *)newKeyTag {
+  NSDate *expiration = [[NSDate date] dateByAddingTimeInterval:kMSEncryptionKeyLifetimeInSeconds];
+  NSString *expirationIso = [MSUtility dateToISO8601:expiration];
+
+  // Format is {keyTag}/{expiration as iso}
```
achocron:
```suggestion
  // Format is {keyTag}/{expiration as iso}.
```

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -17,80 +19,411 @@ @implementation MSEncrypterTests
...
+  id nsdateMock = OCMClassMock([NSDate class]);
+  OCMStub(ClassMethod([nsdateMock timeIntervalSinceReferenceDate])).andReturn(timeSinceReferenceDate);
+  OCMStub(ClassMethod([nsdateMock date])).andReturn(referenceDate);
+  NSDate *expectedExpirationDate = [[NSDate date] dateByAddingTimeInterval:kMSEncryptionKeyLifetimeInSeconds];
+  NSString *expectedExpirationDateIso = [MSUtility dateToISO8601:expectedExpirationDate];
+
+  // Save key to the Keychain.
+  NSString *currentKey = [self generateTestEncryptionKey];
+  [MSMockKeychainUtil storeString:currentKey forKey:kMSEncryptionKeyTagOriginal];
+  MSEncrypter *encrypter = [[MSEncrypter alloc] init];
```
jwallra: Can be simplified (here and elsewhere in the file):

```suggestion
  MSEncrypter *encrypter = [MSEncrypter new];
```

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -17,80 +19,411 @@ @implementation MSEncrypterTests
...
+  // Then
+  XCTAssertEqualObjects(decryptedString, clearText);
+}
+
+- (void)testEncryptionCreatesKeyWhenNoKeyIsSaved {
+
+  // If
+  NSString *clearText = @"clear text";
+  NSString *expectedMetadata = [NSString stringWithFormat:@"%@/AES/CBC/PKCS7/32", kMSEncryptionKeyTagAlternate];
+  MSMockUserDefaults *mockUserDefaults = [[MSMockUserDefaults alloc] init];
```
jwallra:
```suggestion
  MSMockUserDefaults *mockUserDefaults = [MSMockUserDefaults new];
```

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -171,4 +217,22 @@ + (NSData *)generateKeyWithTag:(NSString *)keyTag {
   return resultKey;
 }
 
++ (NSData *)generateInitializationVector {
+  uint8_t *ivBytes = malloc(kCCBlockSizeAES128 * sizeof(uint8_t));
+  memset((void *)ivBytes, 0x0, kCCBlockSizeAES128);
```
jaelim-ms: Just curiosity, any reason of initialize the memory space?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+      CCCrypt(operation, kMSEncryptionAlgorithm, kCCOptionPKCS7Padding, [key bytes], kMSEncryptionKeySize, [initializationVector bytes],
+              [input bytes], input.length, outputBuffer, outputBufferSize, &numBytesNeeded);
+  if (status != kCCSuccess) {
+
+    // Do not print the status; it is a security requirement that specific crypto errors are not printed.
+    MSLogError([MSAppCenter logTag], @"Error performing encryption or decryption.");
+  } else {
+    result = [NSData dataWithBytes:outputBuffer length:numBytesNeeded];
+    if (!result) {
+      MSLogWarning([MSAppCenter logTag], @"Could not create NSData object from encrypted or decrypted bytes.");
```
jaelim-ms: Why is this a warning not an error?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
 
-+ (void)deleteKeyWithTag:(NSString *)keyTag {
-  [MSKeychainUtil deleteStringForKey:keyTag];
++ (void)rotateToNewKeyTag:(NSString *)newKeyTag {
+  NSDate *expiration = [[NSDate date] dateByAddingTimeInterval:kMSEncryptionKeyLifetimeInSeconds];
+  NSString *expirationIso = [MSUtility dateToISO8601:expiration];
+
+  // Format is {keyTag}/{expiration as iso}.
+  NSString *keyMetadata = [@[ newKeyTag, expirationIso ] componentsJoinedByString:kMSEncryptionMetadataInternalSeparator];
+  [[MSUserDefaults shared] setObject:keyMetadata forKey:kMSEncryptionKeyMetadataKey];
```
jaelim-ms:
```suggestion
  [MS_USER_DEFAULTS setObject:keyMetadata forKey:kMSEncryptionKeyMetadataKey];
```

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+- (NSData *)getKeyWithKeyTag:(NSString *)keyTag {
+  NSData *key;
+  @synchronized(self) {
+
+    // Read inside of a synchronized block because NSDictionary is not thread safe.
+    key = [self.keys objectForKey:keyTag];
+  }
+
+  // If key is not cached, try loading it from Keychain.
+  if (!key) {
```
jaelim-ms: Should this be in below synchronized block? Isn't it possible that the key is generated right before the line 154 `if (!stringKey) {`?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -4,100 +4,48 @@
...
 }
 
 - (NSData *_Nullable)encryptData:(NSData *)data {
-  [self createCryptorContexts:kMSEncryptionKeyTag];
-  if (!self.key) {
-    MSLogError([MSAppCenter logTag], @"Could not perform encryption. Encryption key is missing.");
-    return nil;
+  NSString *keyTag = [[self class] getCurrentKeyTag];
+  NSData *key = [self getKeyWithKeyTag:keyTag];
+  NSData *initializationVector = [[self class] generateInitializationVector];
```
jaelim-ms:
```suggestion
  NSData *initializationVector = [MSEncrypter generateInitializationVector];
```

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
-  uint8_t *cipherBuffer = malloc(cipherBufferSize);
-  size_t numBytesDecrypted = 0;
-  CCCryptorStatus status = CCCrypt(kCCDecrypt, kMSEncryptionAlgorithm, kCCOptionPKCS7Padding, [self.key bytes], kMSCipherKeySize, nil,
-                                   [data bytes], data.length, cipherBuffer, cipherBufferSize, &numBytesDecrypted);
-  if (status != kCCSuccess) {
-    MSLogError([MSAppCenter logTag], @"Error performing decryption with CCCryptorStatus: %d.", status);
+  NSData *key;
+  NSData *initializationVector;
+  NSData *cipherText;
+  if (!metadata) {
```
jaelim-ms: Why don't you swap `if` block and `else` block by reversing the condition? Don't need NOT operation.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+
+    // If there is no metadata, this is old data, so use the old key and an empty initialization vector.
+    key = [self getKeyWithKeyTag:kMSEncryptionKeyTagOriginal];
+    initializationVector = nil;
```
jaelim-ms: Redundant. Remove.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,150 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+    cipherText = data;
   } else {
-    result = [NSData dataWithBytes:cipherBuffer length:numBytesDecrypted];
-    if (!result) {
-      MSLogWarning([MSAppCenter logTag], @"Could not create NSData object from decrypted bytes.");
+
+    // Extract key from metadata.
+    NSString *keyTag = [metadata componentsSeparatedByString:kMSEncryptionMetadataInternalSeparator][0];
+    NSRange ivRange = NSMakeRange(metadataLocation + 1, kCCBlockSizeAES128);
```
jaelim-ms: This line isn't a key range. Move this below line 95.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
```diff
@@ -90,3 +91,22 @@ static const NSUInteger kMSPersistenceFlagsMask = 0xFF;
...
+static int const kMSEncryptionAlgorithm = kCCAlgorithmAES;
+static NSString *const kMSEncryptionAlgorithmName = @"AES";
+static NSString *const kMSEncryptionCipherMode = @"CBC";
+
+// One astronomical year.
+static NSTimeInterval const kMSEncryptionKeyLifetimeInSeconds = 365.25 * 24.0 * 60.0 * 60.0;
+static int const kMSEncryptionKeySize = kCCKeySizeAES256;
+static NSString *const kMSEncryptionKeyMetadataKey = @"MSEncryptionKeyMetadata";
+static NSString *const kMSEncryptionKeyTagAlternate = @"kMSEncryptionKeyTagAlternate";
+static NSString *const kMSEncryptionKeyTagOriginal = @"kMSEncryptionKeyTag";
```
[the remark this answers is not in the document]
achocron: Because of backwards compatibility this is the one that existed before the update. So it is helpful to call that out in the name, I believe.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
achocron: Nope. This is copied from the code to generate key, which was doing this. We didn't figure out why, though. If nobody has a reason for it, we can just remove it from both places.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
achocron: Essentially the same message was a warning before. Could change to error.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
achocron: Yep, it is possible. That's why we re-check on line 158, inside the block. Doing it this way ensures that the synchronized code path is rarely hit.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
```diff
@@ -90,3 +91,22 @@ static const NSUInteger kMSPersistenceFlagsMask = 0xFF;
...
  */
 static NSString *const kMSCommonSchemaPrefixSeparator = @":";
+
+// Encryption constants.
+static int const kMSEncryptionAlgorithm = kCCAlgorithmAES;
+static NSString *const kMSEncryptionAlgorithmName = @"AES";
+static NSString *const kMSEncryptionCipherMode = @"CBC";
+
+// One astronomical year.
+static NSTimeInterval const kMSEncryptionKeyLifetimeInSeconds = 365.25 * 24.0 * 60.0 * 60.0;
```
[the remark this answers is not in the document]
achocron: It's just more accurate, years vary in length since we add a day every 4 years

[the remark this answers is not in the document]
achocron: Why would you prefer 365 over 365.25?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
jaelim-ms: Can't you simplify locks?

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -17,80 +19,411 @@ @implementation MSEncrypterTests
...
 
   // If
-  MSEncrypter *encrypter = [[MSEncrypter alloc] initWitKeyTag:self.keyTag];
-  NSString *stringToEncrypt = @"Test string";
+  NSString *clearText = @"clear text";
+  NSString *keyTag = kMSEncryptionKeyTagAlternate;
+  NSString *expectedMetadata = [NSString stringWithFormat:@"%@/AES/CBC/PKCS7/32", keyTag];
+
+  // Save metadata to user defaults.
+  MSMockUserDefaults *mockUserDefaults = [MSMockUserDefaults new];
```
jaelim-ms: Use property for this mock. You can see an example in other test classes that have `self.settingsMock`.

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -17,80 +19,411 @@ @implementation MSEncrypterTests
 - (void)setUp {
   [super setUp];
   self.keychainUtilMock = [MSMockKeychainUtil new];
-  self.keyTag = @"kMSTestEncryptionKeyTag";
 }
 
 - (void)tearDown {
   [self.keychainUtilMock stopMocking];
-  [MSEncrypter deleteKeyWithTag:self.keyTag];
+  [MSMockKeychainUtil clear];
```
jaelim-ms: Why do you need this one? `[self.keychainUtilMock stopMocking];` should be enough.

AppCenter/AppCenterTests/MSEncrypterTests.m
```diff
@@ -17,80 +19,411 @@ @implementation MSEncrypterTests
...
+  // Save metadata to user defaults.
+  MSMockUserDefaults *mockUserDefaults = [MSMockUserDefaults new];
+  NSDate *expiration = [NSDate dateWithTimeIntervalSinceNow:10000000];
+  NSString *expirationIso = [MSUtility dateToISO8601:expiration];
+  NSString *keyMetadataString = [NSString stringWithFormat:@"%@/%@", keyTag, expirationIso];
+  [mockUserDefaults setObject:keyMetadataString forKey:kMSEncryptionKeyMetadataKey];
+
+  // Save key to the Keychain.
+  NSString *currentKey = [self generateTestEncryptionKey];
+  [MSMockKeychainUtil storeString:currentKey forKey:keyTag];
```
jaelim-ms: You can use `MSKeychainUtil` instead.

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: Are you saying that [MSEncrypter deleteKeyWithTag:self.keyTag]; was not needed before then?

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: We sometimes use properties for mocks and we sometimes don't. Why is there a preference for one way or the other?

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,148 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
 }
 
-+ (void)deleteKeyWithTag:(NSString *)keyTag {
-  [MSKeychainUtil deleteStringForKey:keyTag];
++ (void)rotateToNewKeyTag:(NSString *)newKeyTag {
+  NSDate *expiration = [[NSDate date] dateByAddingTimeInterval:kMSEncryptionKeyLifetimeInSeconds];
+  NSString *expirationIso = [MSUtility dateToISO8601:expiration];
+
+  // Format is {keyTag}/{expiration as iso}.
+  NSString *keyMetadata = [@[ newKeyTag, expirationIso ] componentsJoinedByString:kMSEncryptionMetadataInternalSeparator];
```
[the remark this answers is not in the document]
achocron: I don't think it's much if any more costly than instantiating an nsstring and parsing it (which is what happens when you use string formatting).

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -4,100 +4,48 @@
...
 - (NSData *_Nullable)encryptData:(NSData *)data {
-  [self createCryptorContexts:kMSEncryptionKeyTag];
-  if (!self.key) {
-    MSLogError([MSAppCenter logTag], @"Could not perform encryption. Encryption key is missing.");
-    return nil;
+  NSString *keyTag = [[self class] getCurrentKeyTag];
+  NSData *key = [self getKeyWithKeyTag:keyTag];
+  NSData *initializationVector = [[self class] generateInitializationVector];
+  NSData *result = [[self class] performCryptoOperation:kCCEncrypt input:data initializationVector:initializationVector key:key];
+  if (result) {
```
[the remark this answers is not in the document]
achocron: That is logged in `performCryptoOperation`

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: When should MSMockKeychainUtil be used? I don't think we should actually use a keychain on the test device

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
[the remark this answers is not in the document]
achocron: (the NSString in this case being `@"%@%@%@"`)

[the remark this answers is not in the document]
achocron: Seems like componentsJoinedByString is made for this.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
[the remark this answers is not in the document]
jaelim-ms: Then are you going to take care of it every 100 years and every 400 years? It isn't a big deal. Let's make it simple.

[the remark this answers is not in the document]
jwallra: Yes, it's a convention that's strange.  It's weird for a key to expire on 5/15 some of the time when it was created on 5/14

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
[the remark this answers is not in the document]
jaelim-ms: Use `NSMutableString`. `appendString` will do this.

AppCenter/AppCenterTests/MSEncrypterTests.m
jaelim-ms: If it is used by a few tests, we don't need it to be a property. In your case, it is used by all tests except one.

AppCenter/AppCenterTests/MSEncrypterTests.m
jaelim-ms: Sorry for confusion. I'm talking about `[MSMockKeychainUtil clear];`.

AppCenter/AppCenterTests/MSEncrypterTests.m
jaelim-ms: Unlike `MSMockUserDefaults` `MSMockKeychainUtil` is class mocking `MSKeychainUtil` so that you can just use `MSKeychainUtil` directly. We are not using `MSMockKeychainUtil` directly.

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: But before, you can see that it was doing `[self.keychainUtilMock stopMocking];` *and* `[MSEncrypter deleteKeyWithTag:self.keyTag];`, so i'm wondering why that was the case?

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: But i think this way works too, no?

jaelim-ms: Eventually it would be same. The stubs are actually calling methods in the mock class.

AppCenter/AppCenterTests/MSEncrypterTests.m
jaelim-ms: Ah yes. The test class doesn't have `NSUserDefaults` mock so it it will actually using a real one. Let's keep it.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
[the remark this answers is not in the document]
achocron: Fixed

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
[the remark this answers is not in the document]
achocron: Discussed offline.

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: Got it, good to know!

AppCenter/AppCenterTests/MSEncrypterTests.m
achocron: Got it. Fixed.

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
```diff
@@ -90,3 +91,22 @@ static const NSUInteger kMSPersistenceFlagsMask = 0xFF;
  * Common schema prefix separator used in various field values.
  */
 static NSString *const kMSCommonSchemaPrefixSeparator = @":";
+
+// Encryption constants.
+static int const kMSEncryptionAlgorithm = kCCAlgorithmAES;
```
clpolet: Are these constants intended to be used by the encrypter only? If it's not useful to other classes then they can be moved to the encrypter private header.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +65,148 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+  NSString *expirationIso = [MSUtility dateToISO8601:expiration];
+
+  // Format is {keyTag}/{expiration as iso}.
+  NSString *keyMetadata = [@[ newKeyTag, expirationIso ] componentsJoinedByString:kMSEncryptionMetadataInternalSeparator];
+  [MS_USER_DEFAULTS setObject:keyMetadata forKey:kMSEncryptionKeyMetadataKey];
+}
+
+- (NSData *)getKeyWithKeyTag:(NSString *)keyTag {
+  NSData *key;
+  @synchronized(self) {
```
clpolet: Would it be possible to synchronized the whole block on self at this point. It means it will lock on classLock but that shouldn't introduce a deadlock since classLock doesn't lock on self (it wouldn't make sense). That would simplify the code.

jwallra: We want to keep as much code as possible outside of synchronized blocks for performance reasons though, no? I'd prefer to keep the (only slightly) less readable version if it means a small improvement in performance

achocron: We could but that would really reduce performance. The way its written means that class lock will almost never be used. I don't think we should change that unless we have to. We can discuss offline if you aren't convinced

AppCenter/AppCenter/Internals/Util/MSConstants+Internal.h
achocron: Fixed.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
jaelim-ms: I also commented synchronization in the method (sadly no answers for the comment).
We need to think about simplifying this. Sounds like it is a trade-off of simplifying/readability vs performance. Can we talk about this one?

achocron: Removed the lock here altogether. Resolving.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +59,146 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
-  NSString *stringKey = [MSKeychainUtil stringForKey:keyTag];
-  if (stringKey) {
-    keyData = [[NSData alloc] initWithBase64EncodedString:stringKey options:0];
+- (NSData *)getKeyWithKeyTag:(NSString *)keyTag {
+  NSData *keyData;
+  BOOL isOriginalKeyTag = [keyTag isEqualToString:kMSEncryptionKeyTagOriginal];
+  keyData = isOriginalKeyTag ? self.originalKeyData : self.alternateKeyData;
+
+  // If key is not cached, try loading it from Keychain.
+  if (!keyData) {
```
jaelim-ms: Revert the condition to avoid nested `if`.

AppCenter/AppCenter/Internals/Util/MSEncrypter.m
```diff
@@ -117,52 +59,146 @@ - (NSString *_Nullable)decryptString:(NSString *)string {
...
+    if (stringKey) {
+      keyData = [[NSData alloc] initWithBase64EncodedString:stringKey options:0];
+    }
+    else {
+
+      // If key is not saved in Keychain, create one and save it.
+      @synchronized(classLock) {
+
+        // Recheck if the key has been written from another thread.
+        stringKey = [MSKeychainUtil stringForKey:keyTag];
```
jaelim-ms: You explained this but access keychain is a bit expensive.

jaelim-ms: Can the method just be like this?
```

- (NSData *)getKeyWithKeyTag:(NSString *)keyTag {
  NSData *keyData;
  BOOL isOriginalKeyTag = [keyTag isEqualToString:kMSEncryptionKeyTagOriginal];
  keyData = isOriginalKeyTag ? self.originalKeyData : self.alternateKeyData;
  if (keyData) {
    return keyData;
  }

  // If key is not cached, try loading it from Keychain.
  @synchronized(classLock) {

    NSString *stringKey = [MSKeychainUtil stringForKey:keyTag];
    if (stringKey) {
      keyData = [[NSData alloc] initWithBase64EncodedString:stringKey options:0];
    } else {

      // Key is not saved in Keychain, create one and save it.
      keyData = [MSEncrypter generateAndSaveKeyWithTag:keyTag];
    }
  }
  if (isOriginalKeyTag) {
    self.originalKeyData = keyData;
  } else {
    self.alternateKeyData = keyData;
  }
  return keyData;
}
```

PR #1631, opened 2019-05-21
Batching on custom interval

russelarms:

* [ ] Did you add unit tests?

## Description

If a customer wants to send logs only once every day, the SDK must not ignore that time interval once 50 logs are pending.

The send immediately 50 logs rule must apply only when the default of 3 seconds is used.

We must still split logs sent to ingestion by batches of 50 (would be better byte size based but that's out of scope) to avoid hitting payload size limit.

So if SDK has 200 logs, it must still wait transmission interval (1 day for example) and then send all the 200 logs at once (send 3 batches of 50, wait 1 to complete, send the 4th one immediately after 1 of the batches completed).

## Related PRs or issues

https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/61888/

## Misc

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -372,8 +372,9 @@ static NSString *const kMSStartTimestampPrefix = @"MSChannelStartTimer";
 - (void)checkPendingLogs {
 
   // Skip checking pending logs if the channel is paused.
-  if (self.paused)
+  if (self.paused) {
     return;
+  }
 
   // If the interval is default and we reached batchSizeLimit flush logs now.
   if (self.configuration.flushInterval == kMSFlushIntervalDefault && self.itemsCount >= self.configuration.batchSizeLimit) {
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -371,19 +371,30 @@ - (void)flushQueue {
 
 - (void)checkPendingLogs {
 
-  // Flush now if current batch is full or delay to later.
-  if (self.itemsCount >= self.configuration.batchSizeLimit) {
+  // Skip checking pending logs if the channel is paused.
+  if (self.paused)
```
MatkovIvan: Add `{}`

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -428,17 +438,30 @@ - (void)startTimer {
...
+    // If the interval is over.
+    else if ([now compare:[startTime dateByAddingTimeInterval:flushInterval]] == NSOrderedDescending) {
+      return 0;
+    }
+
+    // We still have to wait for the rest of the interval.
+    else {
+      flushInterval -= (NSUInteger)[now timeIntervalSinceDate:startTime];
     }
     return MAX(flushInterval, kMSFlushIntervalDefault);
```
MatkovIvan:
```suggestion
     return flushInterval;
```

PR #1632, opened 2019-05-22
Fixed the redundant token exchange errors

MehrdadMzfr:

* [ ] Did you add unit tests?

## Description

With the current SDK behavior, the Reachability framework sends an event on every app launch which triggers pending Data operation resume. Therefore, there is a call to token exchange to get the token an since user is not authenticated, it'll log an error with 404 error which can mislead developers that something is wrong with the app setup. Downgraded the logging to warning to avoid confusion.

### DIFF
--- a/AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
+++ b/AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
@@ -79,7 +79,7 @@ static NSString *const kMSGetTokenPath = @"/data/tokens";
 
             // Token store returned non-200 response code.
             if (response.statusCode != MSHTTPCodesNo200OK) {
-              MSLogError([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
+              MSLogWarning([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
               completionHandler([[MSTokensResponse alloc] initWithTokens:nil],
                                 [[NSError alloc] initWithDomain:kMSACDataErrorDomain
                                                            code:MSACDataErrorHTTPError
--- a/AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
+++ b/AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
@@ -79,7 +79,7 @@ static NSString *const kMSGetTokenPath = @"/data/tokens";
 
             // Token store returned non-200 response code.
             if (response.statusCode != MSHTTPCodesNo200OK) {
-              MSLogWarning([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
+              MSLogError([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
               completionHandler([[MSTokensResponse alloc] initWithTokens:nil],
                                 [[NSError alloc] initWithDomain:kMSACDataErrorDomain
                                                            code:MSACDataErrorHTTPError
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -753,12 +753,18 @@ static dispatch_once_t onceToken;
   // If user logs in.
   if (userInfomation && userInfomation) {
     [self.dataOperationProxy.documentStore createUserStorageWithAccountId:userInfomation.accountId];
+
+    // Persist Auth context availability flag.
+    [MS_USER_DEFAULTS setObject:@YES forKey:kMSAuthContextAvailabilityKey];
   } else {
     // If user logs out.
     [MSTokenExchange removeAllCachedTokens];
 
     // Delete all the data (user and read-only).
     [self.dataOperationProxy.documentStore resetDatabase];
+
+    // Clean Auth context availability flag.
+    [MS_USER_DEFAULTS removeObjectForKey:kMSAuthContextAvailabilityKey];
   }
 }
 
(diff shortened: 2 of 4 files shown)


AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
```diff
@@ -79,7 +79,7 @@ + (void)performDbTokenAsyncOperationWithHttpClient:(id<MSHttpClientProtocol>)htt
 
             // Token store returned non-200 response code.
             if (response.statusCode != MSHTTPCodesNo200OK) {
-              MSLogError([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
+              MSLogWarning([MSData logTag], @"The token store returned %ld", (long)response.statusCode);
```
jaelim-ms: This could hide some other issues. If 404 could be expected, can you log warning message only for 404? I guess it could be not a warning level but something lower than warning.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -786,8 +786,8 @@ - (void)processPendingOperations {
                                       reachability:self.reachability
                                  completionHandler:^(MSTokensResponse *_Nonnull tokenResponses, NSError *_Nonnull error) {
                                    if (error) {
-                                     MSLogError([MSData logTag], @"Cannot read from local storage because there is no "
-                                                                 @"account ID cached and failed to retrieve token.");
+                                     MSLogWarning([MSData logTag], @"Cannot read from local storage because there is no "
```
jaelim-ms: This seems a duplicate log for a single operation. We could eliminate one of them.

AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
jwallra: @jaelim-ms I agree.  Do we have other logic that splits out error codes by number and gives different log levels based on the number, though?  I think at worst this case is an error, so if we want it to cover the worst number codes, it should remain an error.  If we want to make it less confusing, which is @MehrdadMzfr's goal, we should make more granular responses

AppCenterData/AppCenterData/MSData.m
dhei: Agree with @jaelim-ms. If `(void)performDbTokenAsyncOperationWithHttpClient:` already logs properly for each error case (https://github.com/microsoft/appcenter-sdk-apple/pull/1632/files#diff-a797f8f186af5727476c3ecc3ad4f2fbR29), then we can remove this line completely.

AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
dhei: I don't have all the context of this issue so I'd hope you can help me understand these two questions:

1. Would there be any case in this codepath where 404 response from token exchange is caused by an actual error which we want to log as MSLogError? In other words, any false negatives?

2. From a design perspective, ideally we want to avoid making a call to token exchange when we know it's going to be a 404. Is there any way to filter out this scenario when we make this call in `MSTokenExchange`? I don't think there's a way to do that. Just want to confirm this is the case.

MehrdadMzfr: reverted this change.

AppCenterData/AppCenterData/MSData.m
MehrdadMzfr: I'd prefer to keep this as Warning since it provides more context around the call that ends up to an unwanted state. Token Exchange only logs generic errors.

AppCenterData/AppCenterData/Internal/Client/MSTokenExchange.m
MehrdadMzfr: @dhei 
1- your point is valid, reverted the change and kept the original error

2- added the change to avoid redundant call to token exchange service.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -776,6 +782,14 @@ - (void)networkStateChanged:(NSNotificationCenter *)__unused notification {
 }
 
 - (void)processPendingOperations {
+
+  // Only process pending operations when auth context is available.
+  NSNumber *authContextFlag = [MS_USER_DEFAULTS objectForKey:kMSAuthContextAvailabilityKey];
+  if (!authContextFlag.boolValue) {
```
jaelim-ms: What happens if a user is authenticated but the token is expired after a while? This will still lead 404 error?

AppCenterData/AppCenterData/MSData.m
```diff
@@ -753,12 +753,18 @@ - (void)authTokenContext:(MSAuthTokenContext *)__unused authTokenContext didUpda
...
+    [MS_USER_DEFAULTS setObject:@YES forKey:kMSAuthContextAvailabilityKey];
   } else {
     // If user logs out.
     [MSTokenExchange removeAllCachedTokens];
 
     // Delete all the data (user and read-only).
     [self.dataOperationProxy.documentStore resetDatabase];
+
+    // Clean Auth context availability flag.
+    [MS_USER_DEFAULTS removeObjectForKey:kMSAuthContextAvailabilityKey];
```
jaelim-ms: Since you are removing the value for the key here, it should be added to a method when Data is disabled.

AppCenterData/AppCenterData/MSData.m
MehrdadMzfr: the token will be renewed explicitly. this is not any different from the current behavior.

jaelim-ms: It is probably an edge case but if it takes time to refresh token but Data started and trying to talk to token exchange service. It still has a chance to get 404 in that case?

MehrdadMzfr: when Data wants to manipulate doc, it needs a token, if the cached token is expired, then it tries to get a new token.
if it fails to renew a token, then it’ll log error which is a valid case.
the bug was that Data tried to get token when app is not signed in, therefore, it logged 404

MehrdadMzfr:
@dhei valid point, was not aware of that API. changed accordingly.

PR #1633, opened 2019-05-23
Fix tokens enumeration

MatkovIvan:
* [x] Did you add unit tests?

## Description

Resuming Analytics doesn't send pending Analytics logs from all tokens. It actually send logs corresponding to the oldest entry in the token history only. Resuming (or Pause/resuming) again triggers the sending for logs corresponding to other entries in the token history.

Repro:
1) Sign-out
2) Pause Analytics
3) Track an event
4) Sign-in 
5) Track an event
6) Resume (the anonymous event is sent)
7) Pause/Resume (the authenticated event is sent)

Expected behavior: Both events are sent on step 6)

## Related PRs or issues

[AB#61939](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61939)

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
@@ -300,28 +300,28 @@
                         if (logArray.count > 0) {
                           MSLogContainer *container = [[MSLogContainer alloc] initWithBatchId:batchId andLogs:logArray];
                           [self sendLogContainer:container withAuthTokenFromArray:tokenArray atIndex:tokenIndex];
-                        } else {
+                        }
 
-                          // No logs available with given params.
-                          if (tokenIndex == 0 && tokenArray[tokenIndex].endTime != nil &&
-                              [self.storage countLogsBeforeDate:tokenArray[tokenIndex].endTime] == 0) {
+                        // No logs available with given params.
+                        else if (tokenIndex == 0 && tokenArray[tokenIndex].endTime != nil &&
+                            [self.storage countLogsBeforeDate:tokenArray[tokenIndex].endTime] == 0) {
 
-                            // Delete token from history if we don't have logs fitting it in DB.
-                            [[MSAuthTokenContext sharedInstance] removeAuthToken:tokenInfo.authToken];
-                          }
-
-                          // Check to determine if the next index is within bounds.
-                          if (tokenIndex + 1 < tokenArray.count) {
-
-                            // Iterate to next token in array.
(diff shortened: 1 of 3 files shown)


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -300,28 +300,28 @@ - (void)flushQueueForTokenArray:(NSArray<MSAuthTokenValidityInfo *> *)tokenArray
...
-                          }
+                          // Delete token from history if we don't have logs fitting it in DB.
+                          [[MSAuthTokenContext sharedInstance] removeAuthToken:tokenInfo.authToken];
                         }
                       }];
 
   // Flush again if there is another batch to send.
-  if (self.availableBatchFromStorage && !self.pendingBatchQueueFull) {
-    [self flushQueueForTokenArray:tokenArray withTokenIndex:tokenIndex];
+  if (!self.pendingBatchQueueFull) {
```
jwallra: This kind of seems like a band-aid for the problem.  Yes, we can just send more logs if they are present, but it seems like we should be able to determine which logs should be sent all at once... Maybe I'm misunderstanding, though?  It's two different ways to solve the same problem, but this seems like a reactive solution as opposed to a proactive one

Could you explain your logic for this particular solution @MatkovIvan ?

MatkovIvan: @jwallra the issue is clear and this is not workaround or dirty fix or something like that.
Let me simplify this stuff (unwrap load function) and provide pseudo-code:

```
logArray = read_from_db(batchSizeLimit + 1)
if (logArray.count > batchSizeLimit) {
  logArray.removeLastObject();
  availableBatchFromStorage = YES;
}

// BEFORE:
noMoreLogs = logArray.count == 0;

// NOW:
noMoreLogs = !availableBatchFromStorage;
```

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -1299,7 +1299,7 @@ - (void)testFlushQueueIteratesThroughArrayRecursively {
   [tokenValidityArray addObject:token1];
   [tokenValidityArray addObject:token2];
   [tokenValidityArray addObject:token3];
-  NSArray<id<MSLog>> *logsForToken1 = [self getValidMockLogArrayForDate:date1 andCount:0];
+  NSArray<id<MSLog>> *logsForToken1 = [self getValidMockLogArrayForDate:date1 andCount:3];
```
clpolet: I am confused in this test about how it verifies that the correct logs are sent with the correct token.

PR #1646, opened 2019-06-01
Add remoteOperation delegate to data module

smithab:

* [x] Did you add unit tests?

## Description

We need to enable a delegate which will get triggered when the pending operations get processed after the device goes back online

## Related PRs or issues

## Misc

### DIFF
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -925,11 +925,12 @@ static dispatch_once_t onceToken;
   }
 
   // If the Remote operation is set
+  id<MSRemoteOperationDelegate> strongDelegate;
   @synchronized(self) {
-    id<MSRemoteOperationDelegate> strongDelegate = self.remoteOperationDelegate;
-    if (strongDelegate) {
-      [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
-    }
+    strongDelegate = self.remoteOperationDelegate;
+  }
+  if (strongDelegate) {
+    [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
   }
 }
 
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -94,9 +94,6 @@ static dispatch_once_t onceToken;
   [[MSData sharedInstance] setTokenExchangeUrl:(NSURL *)[NSURL URLWithString:tokenExchangeUrl]];
 }
 
-/**
- * Set a delegate which will be invoked on network status change to notify pending operations execution status.
- */
 + (void)setRemoteOperationDelegate:(nullable id<MSRemoteOperationDelegate>)delegate {
   @synchronized(self) {
     [[MSData sharedInstance] setRemoteOperationDelegate:delegate];
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -925,9 +925,11 @@ static dispatch_once_t onceToken;
   }
 
   // If the Remote operation is set
-  id<MSRemoteOperationDelegate> strongDelegate = self.remoteOperationDelegate;
-  if (strongDelegate) {
-    [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
+  @synchronized(self) {
+    id<MSRemoteOperationDelegate> strongDelegate = self.remoteOperationDelegate;
+    if (strongDelegate) {
+      [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2152,4 +2159,87 @@ - (NSHTTPURLResponse *)generateResponseWithStatusCode:(NSInteger)statusCode {
...
+  //                                                      object:self.reachabilityMock];
+  
+  [self.sut processPendingOperations];
+  
+  // Then
+  id<MSRemoteOperationDelegate> strongDelegate = [MSData sharedInstance].remoteOperationDelegate;
+  XCTAssertNotNil(strongDelegate);
+  XCTAssertEqual(strongDelegate, delegateMock);
+  
+  OCMVerify([delegateMock data:self.sut didCompletePendingOperation:kMSPendingOperationCreate forDocument:localStorageDocument withError:nil]);
```
smithab: I am having a timing issue with this test. The verification step checks even before the pending operations get complete. It goes ahead and fails saying the delegate was never invoked. Any suggestions on how I can fix this

jaelim-ms: 1. You need to set auth token at the beginning of the test. delegate isn't being called.
2. As you mentioned there is a timing issue. You need to use `XCTestExpectation` and wait until delegate is called before the verification.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -905,16 +915,24 @@ - (void)synchronizeLocalCacheWithCosmosDbWithToken:(MSTokenResult *)token
...
+    document = nil;
   }
-
+  
   // Delete the document form the local cache.
   if (shouldDeleteLocalCache) {
     [self.dataOperationProxy.documentStore deleteWithToken:token documentId:documentId];
   }
+  
+  //If the Remote operation is set
```
jaelim-ms: Space after `//`

AppCenterData/AppCenterData/MSRemoteOperationDelegate.h
```diff
@@ -0,0 +1,31 @@
...
+@class MSDocumentWrapper;
+@class MSDataError;
+
+@protocol MSRemoteOperationDelegate <NSObject>
+
+@optional
+
+/**
+ * A callback that is called when the device network status gets changed from offine to online mode.
+ * The Data operations are then performed against the cosmosDB
```
clpolet:
```suggestion
 * The Data operations are then performed against the cosmosDB.
```

AppCenterData/AppCenterData/MSRemoteOperationDelegate.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param data The instance of MSData.
+ * @param operation Operation ran.
+ * @param document Document that was synched. Nil if error encountered
```
clpolet: Nit: should use the formal form plus code formatting:
```suggestion
 * @param document Document that was synchronized. `nil` if error encountered
```

AppCenterData/AppCenterData/MSRemoteOperationDelegate.h
```diff
@@ -0,0 +1,31 @@
...
+
+@protocol MSRemoteOperationDelegate <NSObject>
+
+@optional
+
+/**
+ * A callback that is called when the device network status gets changed from offine to online mode.
+ * The Data operations are then performed against the cosmosDB
+ *
+ * @param data The instance of MSData.
```
clpolet: Nit:
```suggestion
 * @param data The instance of `MSData`.
```

AppCenterData/AppCenterData/MSRemoteOperationDelegate.h
```diff
@@ -0,0 +1,31 @@
...
+ * @param operation Operation ran.
+ * @param document Document that was synched. Nil if error encountered
+ * @param error Error details or nil when the sync was successful
```
clpolet:
```suggestion
 * @param error Error details or `nil` when the synchronization was successful.
```

AppCenterData/AppCenterData/MSData.m
```diff
@@ -915,6 +923,14 @@ - (void)synchronizeLocalCacheWithCosmosDbWithToken:(MSTokenResult *)token
   if (shouldDeleteLocalCache) {
     [self.dataOperationProxy.documentStore deleteWithToken:token documentId:documentId];
   }
+
+  // If the Remote operation is set
+  @synchronized(self) {
+    id<MSRemoteOperationDelegate> strongDelegate = self.remoteOperationDelegate;
+    if (strongDelegate) {
+      [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
```
clpolet: As a good practice the execution of the delegate is happening outside of the `synchronized`. This is because what the delegate is doing is either unknown or uneasy to track. In this case it's unknown since it's developer code.
This means lines 930 to 932 should be moved after the `synchronized` block.

clpolet: That will also help in keeping the `synchronized` block to it's minimal.

jaelim-ms: Yeah.. This is caused by my comment over offline communication channel.
My intention was to ask to change this as below.
```
  // If the Remote operation is set
  id<MSRemoteOperationDelegate> strongDelegate;
  @synchronized(self) {
    strongDelegate = self.remoteOperationDelegate;
  }
  if (strongDelegate) {
    [strongDelegate data:self didCompletePendingOperation:pendingOperation forDocument:document withError:documentWrapper.error];
  }
```

PR #1651, opened 2019-06-06
Validate URL scheme for Auth

jaelim-ms:
* [X] Did you add unit tests?

## Description

Add custom URL scheme validation on Auth start.

## Related PRs or issues

[AB#62677](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/62677)

### DIFF
--- a/AppCenterAuth/AppCenterAuth/MSAuth.m
+++ b/AppCenterAuth/AppCenterAuth/MSAuth.m
@@ -72,9 +72,13 @@ static dispatch_once_t onceToken;
                     appSecret:(nullable NSString *)appSecret
       transmissionTargetToken:(nullable NSString *)token
               fromApplication:(BOOL)fromApplication {
-  [[MSAuthTokenContext sharedInstance] preventResetAuthTokenAfterStart];
-  [super startWithChannelGroup:channelGroup appSecret:appSecret transmissionTargetToken:token fromApplication:fromApplication];
-  MSLogVerbose([MSAuth logTag], @"Started Auth service.");
+  if ([self checkURLSchemeRegistered:[NSString stringWithFormat:kMSMSALCustomSchemeFormat, appSecret]]) {
+    [[MSAuthTokenContext sharedInstance] preventResetAuthTokenAfterStart];
+    [super startWithChannelGroup:channelGroup appSecret:appSecret transmissionTargetToken:token fromApplication:fromApplication];
+    MSLogVerbose([MSAuth logTag], @"Started Auth service.");
+  } else {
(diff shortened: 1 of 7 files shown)


AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -258,6 +262,22 @@ - (void)signOut {
 
 #pragma mark - Private methods
 
+- (BOOL)checkURLSchemeRegistered:(NSString *)urlScheme {
+  NSArray *schemes;
+  NSString *typeRole;
+  NSArray *types = [MS_APP_MAIN_BUNDLE objectForInfoDictionaryKey:kMSCFBundleURLTypes];
```
achocron: I guess this will crash if it is not an `NSArray`

AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -258,6 +262,22 @@ - (void)signOut {
...
+  for (NSDictionary *urlType in types) {
+    schemes = urlType[kMSCFBundleURLSchemes];
```
achocron: I think this will crash if `urlType` is not an `NSDictionary`, and same if `scheme` and `typeRole` are not `NSString`

AppCenterAuth/AppCenterAuth/MSAuth.m
achocron: Also can it be `nil`?

AppCenterAuth/AppCenterAuth/MSAuth.m
achocron: Or if `scheme` is not an `NSArray`. Basically, should we type check these?

jaelim-ms: `CFBundleURLTypes` is an array of dictionaries. [Apple Doc](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleurltypes)
`CFBundleURLSchemes` is an array of strings. [Apple Doc](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleurltypes/cfbundleurlschemes)

AppCenterAuth/AppCenterAuth/MSAuth.m
jaelim-ms: It is an `NSArray`. Look at below answer.
I also added a unit test for the case if it is `nil`.

jaelim-ms:
@achocron I don't think we need to consider that invalid case in the code.

PR #1663, opened 2019-06-14
List offline

smithab:

* [x] Did you add unit tests?

## Description

- When we are offline, we need list to get all documents from the local store. 
- If there is a document with pending operation DELETE, it should be ignored. 
- If we are online and still have pending operations, we need to list from local store
- If we encounter an expired document in the local store, delete the entry from the table
- If we are online and no pending operations, talk to remote. 

## Related PRs or issues

## Misc


AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,64 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+      MSPaginatedDocuments *cachedDocumentsList = [self.documentStore listWithToken:token
+                                                                          partition:partition
+                                                                       documentType:documentType
+                                                                        baseOptions:baseOptions];
+
+      completionHandler(cachedDocumentsList);
+      return;
+    }
+
+    // Execute remote operation online and does not have any pending operations.
```
jwallra:
```suggestion
    // Execute remote operation while online and there are no pending operations.
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -199,45 +248,15 @@ - (BOOL)shouldAttemptRemoteOperationForDocument:(MSDocumentWrapper *)document {
...
+ * @param partition The partition under which to check for pending operations.
  *
- * @param token The CosmosDB token.
- * @param currentCachedDocument The current cached document.
- * @param newCachedDocument The new document that should be cached.
- * @param deviceTimeToLive The device time to live for the new cached document.
- * @param operation The operation being intended (nil - read, CREATE, UPDATE, DELETE).
+ * @return YES if a remote operation should be attempted; NO otherwise.
  */
-- (void)updateLocalStore:(MSTokenResult *)token
```
jwallra: Why was this removed?

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -199,45 +248,15 @@ - (BOOL)shouldAttemptRemoteOperationForDocument:(MSDocumentWrapper *)document {
...
-
-  // Update document storage.
-  else {
-    MSLogInfo([MSData logTag], @"Updating/inserting document into local storage (partition: %@, id: %@, operation: %@)", token.partition,
-              currentCachedDocument.documentId, operation);
-    [self.documentStore upsertWithToken:token documentWrapper:newCachedDocument operation:operation deviceTimeToLive:deviceTimeToLive];
-  }
+- (BOOL)shouldAttemptRemoteOperationForPartition:(NSString *)partition {
+  return [self.reachability currentReachabilityStatus] != NotReachable &&
+         [self.documentStore hasPendingOperationsForPartition:partition] == false;
```
jwallra:
```suggestion
         ![self.documentStore hasPendingOperationsForPartition:partition];
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT * FROM \"%@\" WHERE \"%@\" = \"%@\"", tableName, kMSPartitionColumnName, token.partition];
+  NSArray *listResult = [self.dbStorage executeSelectionQuery:selectionQuery];
+
+  // Return an error if no documents were found.
+  if (listResult.count == 0) {
+    NSString *errorMessage = [NSString stringWithFormat:@"Unable to find any document in local store for partition '%@'", token.partition];
+    MSLogWarning([MSData logTag], @"%@", errorMessage);
+
+    // return an empty list
```
jwallra:
```suggestion
    // Return an empty list.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+                                                                    documentType:documentType
+                                                                deviceTimeToLive:baseOptions.deviceTimeToLive
+                                                               continuationToken:nil];
+    return documents;
+  }
+
+  // Parse the documents.
+  for (id documentRow in listResult) {
+
+    // If an expired document is found exclude it from the list and delete in from the local store.
```
jwallra:
```suggestion
    // If an expired document is found exclude it from the list and delete it from the local store.
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,64 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+                                                                          partition:partition
+                                                                       documentType:documentType
+                                                                        baseOptions:baseOptions];
+
+      completionHandler(cachedDocumentsList);
+      return;
+    }
+
+    // Execute remote operation online and does not have any pending operations.
+    if ([self shouldAttemptRemoteOperationForPartition:partition]) {
```
yoannchaudet: replace line by `else`

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+
+        // Ignore document, if the pending operation is found to be Delete
+        NSString *message = [NSString
+            stringWithFormat:
+                @"Local document pending deletion in local storage for partition '%@' and document ID '%@', excluding from the list",
+                token.partition, documentId];
+        MSLogError([MSData logTag], @"%@", message);
+      } else {
+
+        // Deserialize document.
```
jwallra: This block and the two following it (Deserialize, "if operation NSNull", and "if etag NSNull") can be factored out of this method and `readWithToken`

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +367,49 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
...
 }
 
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition {
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:partition];
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT * FROM \"%@\" WHERE \"%@\" IS NOT NULL", tableName, kMSPendingOperationColumnName];
+  NSArray *result = [self.dbStorage executeSelectionQuery:selectionQuery];
+
+  // If empty list
+  if (result.count == 0) {
```
jwallra: can be simplified:

```suggestion
return result.count != 0;
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -80,11 +82,11 @@ - (void)performOperation:(NSString *_Nullable)operation
           }
           // For other online scenarios, the intended pending operation in the local store is nil.
           else {
-            [self updateLocalStore:token
-                currentCachedDocument:cachedDocument
-                    newCachedDocument:remoteDocument
-                     deviceTimeToLive:deviceTimeToLive
-                            operation:nil];
+            [self.documentStore updateDocumentWithToken:token
```
jwallra: Why rename this and move it?

AppCenterData/AppCenterData/Internal/LocalStore/MSDocumentStore.h
```diff
@@ -87,6 +88,45 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString *)documentId documentType:(Class)documentType;
 
+/**
+ * List all the documents from the store and return it if it did not expired.
```
jwallra:
```suggestion
 * List all the documents from the store and return the list if not expired.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+  // Get effective device time to live.
+  NSInteger deviceTimeToLive = baseOptions ? baseOptions.deviceTimeToLive : kMSDataTimeToLiveDefault;
+
+  // Execute the query.
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:token.partition];
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT * FROM \"%@\" WHERE \"%@\" = \"%@\"", tableName, kMSPartitionColumnName, token.partition];
+  NSArray *listResult = [self.dbStorage executeSelectionQuery:selectionQuery];
+
+  // Return an error if no documents were found.
```
yoannchaudet: Fix comment. We log a warning but return an empty list.

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+  if (listResult.count == 0) {
```
yoannchaudet: Since in the iteration on cached documents, first condition checks if the document has expired prior to adding it to the local list ; it would be better to move this condition after the for loop and test the local list.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -407,69 +417,25 @@ - (void)listDocumentsWithType:(Class)documentType
...
-                                  MSPage *page = [[MSPage alloc] initWithItems:items];
-                                  MSPaginatedDocuments *documents = [[MSPaginatedDocuments alloc]
-                                           initWithPage:page
-                                              partition:partition
-                                           documentType:documentType
-                                      continuationToken:[response allHeaderFields][kMSDocumentContinuationTokenHeaderKey]];
-                                  completionHandler(documents);
-                                }];
+      [self.dataOperationProxy performListOnDocumentType:documentType
+          partition:partition
```
jwallra: Format document

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+
+  // Parse the documents.
+  for (id documentRow in listResult) {
+
+    // If an expired document is found exclude it from the list and delete in from the local store.
+    long expirationTime = [(NSNumber *)(documentRow[self.expirationTimeColumnIndex]) longValue];
+    NSString *documentId = documentRow[self.documentIdColumnIndex];
+    if (expirationTime != kMSDataTimeToLiveInfinite) {
+      NSDate *expirationDate = [NSDate dateWithTimeIntervalSince1970:expirationTime];
+      NSDate *currentDate = [NSDate date];
```
yoannchaudet: nit: move the current date out of the loop to initialize it once

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+      if (expirationDate && [expirationDate laterDate:currentDate] == currentDate) {
+
+        NSString *errorMessage =
```
yoannchaudet: nit: rename to `warningMessage`

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+        // Delete the local document when found to be expired
+        [self deleteWithToken:token documentId:documentId];
+      } else if ([kMSPendingOperationDelete isEqualToString:documentRow[self.pendingOperationColumnIndex]]) {
+
+        // Ignore document, if the pending operation is found to be Delete
+        NSString *message = [NSString
+            stringWithFormat:
+                @"Local document pending deletion in local storage for partition '%@' and document ID '%@', excluding from the list",
+                token.partition, documentId];
+        MSLogError([MSData logTag], @"%@", message);
```
yoannchaudet: probably should be warning too

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +191,108 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+                                                                                documentType:documentType
+                                                                                        eTag:etag
+                                                                             lastUpdatedDate:lastUpdatedDate
+                                                                                   partition:token.partition
+                                                                                  documentId:documentId
+                                                                            pendingOperation:pendingOperation
+                                                                             fromDeviceCache:YES];
+        [localListItems addObject:cachedDocument];
+
+        // Push back cached document expiration time then return it.
```
yoannchaudet: fix comment, you are not returning a document here

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +367,49 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
   return pendingDocuments;
 }
 
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition {
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:partition];
+  NSString *selectionQuery =
```
yoannchaudet: How about doing a select count(*) instead of trying to get all documents and instantiating them? It would be much cheaper (perf wise).

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+  [MSData listDocumentsWithType:[MSDictionaryDocument class]
+                      partition:kMSPartitionTest
+              completionHandler:^(MSPaginatedDocuments *_Nonnull documents) {
+                // Then
+                XCTAssertEqualObjects(expectedDocuments, documents);
+                [expectation fulfill];
+              }];
+
+  // Then
+  [self waitForExpectationsWithTimeout:3
```
jwallra: Most other tests use a timeout of 1.  We'd prefer to use lower values, have you tried this test with a value of 1?  Is it flaky?  How about on CI?

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+                // Then
+                XCTAssertNotNil(documents);
+                XCTAssertNil([[documents currentPage] items]);
+                XCTAssertNotNil(documents.currentPage.error);
+                XCTAssertEqual(documents.currentPage.error.domain, kMSACDataErrorDomain);
+                [expectation fulfill];
+              }];
+
+  // Then
+  [self waitForExpectationsWithTimeout:8
```
jwallra: Why is timeout 8?

AppCenterData/AppCenterDataTests/MSDataTests.m
jwallra: Also consider this a blanket for other tests; we should use a value of 1 where possible

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+  OCMStub([reachabilityMock currentReachabilityStatus]).andReturn(NotReachable);
+  self.sut.dataOperationProxy.reachability = reachabilityMock;
+
+  // Mock expired document in local storage.
+  MSDataError *dataError = [[MSDataError alloc] initWithErrorCode:MSACDataErrorLocalDocumentExpired innerError:nil message:nil];
+  id<MSDocumentStore> localStorageMock = OCMProtocolMock(@protocol(MSDocumentStore));
+  self.sut.dataOperationProxy.documentStore = localStorageMock;
+  MSDocumentWrapper *expiredDocument = [[MSDocumentWrapper alloc] initWithError:dataError partition:nil documentId:@"4"];
+  OCMStub([localStorageMock readWithToken:tokenResult documentId:OCMOCK_ANY documentType:OCMOCK_ANY]).andReturn(expiredDocument);
+
```
jwallra:
```suggestion
```

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+                                                                                    fromDeviceCache:YES]);
+  NSMutableArray<MSDocumentWrapper *> *localListItems = [NSMutableArray new];
+  [localListItems addObject:localDocumentWrapper];
+  MSPage *page = [[MSPage alloc] initWithItems:localListItems];
+  MSPaginatedDocuments *localDocumentList = [[MSPaginatedDocuments alloc] initWithPage:page
+                                                                             partition:kMSPartitionTest
+                                                                          documentType:[MSDictionaryDocument class]
+                                                                      deviceTimeToLive:kMSDataTimeToLiveDefault
+                                                                     continuationToken:nil];
+
```
jwallra:
```suggestion
```

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+  OCMStub(localDocumentWrapper.eTag).andReturn(@"some other etag");
+
```
jwallra:
```suggestion
```

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+  OCMStub([localStorageMock listWithToken:tokenResult partition:OCMOCK_ANY documentType:OCMOCK_ANY baseOptions:OCMOCK_ANY])
+      .andReturn(localDocumentList);
+
+  // Mock CosmosDB requests.
+
```
jwallra:
```suggestion
```

AppCenterData/AppCenterData/MSPaginatedDocuments.m
```diff
@@ -16,22 +16,29 @@ @implementation MSPaginatedDocuments
...
   }
   return self;
 }
 
 - (instancetype)initWithError:(MSDataError *)error partition:(NSString *)partition documentType:(Class)documentType {
-  return [self initWithPage:[[MSPage alloc] initWithError:error] partition:partition documentType:documentType continuationToken:nil];
+  return [self initWithPage:[[MSPage alloc] initWithError:error]
+                  partition:partition
+               documentType:documentType
+           deviceTimeToLive:0
```
yoannchaudet: Can you check what value we return on Android?

I would almost introduce a dedicated constant for "unknown" device time to live value instead. Or at least use the `kMSDataTimeToLiveNoCache` constant.

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,257 @@ - (void)testListWhenOffline {
...
+  // Then
+  [self waitForExpectationsWithTimeout:3
+                               handler:^(NSError *_Nullable error) {
+                                 if (error) {
+                                   XCTFail(@"Expectation Failed with error: %@", error);
+                                 }
+                               }];
+}
+
+- (void)testListFromRemoteIfNotExpiredAndOnlineWithNoPendingOperation {
```
yoannchaudet: Can we add a test to check if: we do a remote operation when we have only expired documents in the cache and we are online. I think currently this would return an empty list but it should make a remote call. Kind of a special case.

AppCenterData/AppCenterData/MSPaginatedDocuments.m
smithab: Will use kMSDataTimeToLiveNoCache. Android doesnt seem to have an initWithError kind of implementation

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
smithab: Its moved to the Storage class level cos list api would need it extensively

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,64 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+                                  newCachedDocument:documentWrapper
+                                   deviceTimeToLive:deviceTimeToLive
+                                          operation:operation];
         completionHandler(documentWrapper);
       }
     }
   });
 }
 
+- (void)performListOnDocumentType:(Class)documentType
```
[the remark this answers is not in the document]
smithab: Looks like this is the convention . renaming it to `performListWithDocumentType`

AppCenterData/AppCenterData/MSPaginatedDocuments.m
clpolet: That's because passing the error to the customer is different in Android. The error/exception pattern are not the same between platforms.

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
jwallra: @smithab Unresolving... Is there a reason these can't or shouldn't be factored out?  I see there's a bit of a difference with the document row, but at the very least, the "operation == NSNull" and "ETag == NSNull" blocks should be able to be factored out

AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
```diff
@@ -40,6 +46,9 @@ - (void)tearDown {
 
   [self.documentStoreMock stopMocking];
   [self.reachability stopMocking];
+
+  // Delete existing database.
+  [MSUtility deleteItemForPathComponent:kMSDBDocumentFileName];
```
jwallra: We shouldn't have to do this in both `setup` and `tearDown`, pick one place

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
smithab: Fixed it.

AppCenterData/AppCenterDataTests/MSDataTests.m
smithab: Good catch. It wad a bug. !! Fixed it

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,58 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
...
+    [self upsertWithToken:token documentWrapper:newCachedDocument operation:operation deviceTimeToLive:deviceTimeToLive];
+  }
+}
+
+- (NSString *)checkForNSNull:(NSString *)column {
+
+  // If column is NSNull, change it to nil.
+  NSString *nullableColumn = column;
+  NSString *fColumn = [nullableColumn isEqual:[NSNull null]] ? nil : nullableColumn;
+
```
jwallra:
```suggestion
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -207,12 +207,19 @@ - (void)performListOnDocumentType:(Class)documentType
                                                                        documentType:documentType
                                                                         baseOptions:baseOptions];
 
+      if ([self.reachability currentReachabilityStatus] == ReachableViaWiFi && [[cachedDocumentsList currentPage] items].count == 0) {
```
yoannchaudet: It's better to check that the status is not unrechable. This test does not include cellular reachability I guess.

guperrot: sounds like a bug to me

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # App Center SDK for iOS and macOS Change Log
 
+## Version 2.2.0
```
guperrot:
```suggestion
## Version 2.2.0 (Under development)
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.h
```diff
@@ -61,6 +61,24 @@ typedef void (^MSCachedTokenCompletionHandler)(MSTokensResponse *_Nullable token
...
+ * using a combination of the local store and/or CosmosDB remote calls.
+ *
+ * @param documentType The documents type.
+ * @param partition The partition of the documents.
+ * @param baseOptions The base options from which to get the device time to live (if specified).
+ * @param cachedTokenBlock A block returning the cached token.
+ * @param remoteDocumentBlock A block returning the remote document.
+ * @param completionHandler The completion handler called ultimately.
+ */
+- (void)performListDocumentsWithType:(Class)documentType
```
clpolet: Now I get the context :). We can probably get rid of `perform` in `performListDocumentsWithType` since the list operation is explicit and already used in its verb form.

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,71 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+    // Extract first token.
+    MSTokenResult *token = tokensResponse.tokens.firstObject;
+
+    // Retrieve from cache when offline and when there are pending operations
+    if (![self shouldAttemptRemoteOperationForPartition:partition]) {
+      MSPaginatedDocuments *cachedDocumentsList = [self.documentStore listWithToken:token
+                                                                          partition:partition
+                                                                       documentType:documentType
+                                                                        baseOptions:baseOptions];
+
```
clpolet: Can remove this empty line or add a comment.

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,71 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+      if ([self.reachability currentReachabilityStatus] != NotReachable && [[cachedDocumentsList currentPage] items].count == 0) {
+        MSLogInfo([MSData logTag], @"Performing remote operation, since the local list if empty");
```
clpolet:
```suggestion
        MSLogInfo([MSData logTag], @"Performing remote operation, since the local list is empty");
```

AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
```diff
@@ -29,7 +31,8 @@ - (void)setUp {
   [super setUp];
 
   // Init properties.
-  _documentStoreMock = OCMClassMock([MSDBDocumentStore class]);
+  MSDBStorage *dbStorage = [[MSDBStorage alloc] initWithVersion:0 filename:kMSDBDocumentFileName];
```
clpolet: Why do we need this? I don't see any change on tests.

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,319 @@ - (void)testListWhenOffline {
...
-  XCTAssertEqual(actualPaginatedDocuments.currentPage.error.domain, kMSACDataErrorDomain);
-  XCTAssertEqual(actualPaginatedDocuments.currentPage.error.code, MSACDataErrorHTTPError);
+  XCTAssertNil(actualPaginatedDocuments.currentPage.error);
+  XCTAssertEqual([[actualPaginatedDocuments currentPage] items].count, 0);
+}
+
+- (void)testListReturnsFromLocalStorageWhenOffline {
+
+  // If
+  __weak XCTestExpectation *expectation = [self expectationWithDescription:@""];
```
clpolet: Any reason for weak expectation in these tests?

clpolet: Better to add a description to your expectation, it'll show up in tests logs on the CI.

AppCenterData/AppCenterDataTests/MSDataOperationProxyTests.m
smithab: There is a function which does update on the local store and does some validation on expiration, pending operation and then actual upsert on the document. It used to reside in the `MSDataOperationProxy` class but moved it to the `MSDBDocumentStore` layer cos list needs it extensively. So the tests which depend on this validation/upsert function needs a partial mock of the documentStore.

smithab: The function is called `updateDocumentWithToken` in the class `MSDBDocumentStore`

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,74 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+    // Retrieve from cache when offline and when there are pending operations
+    if (![self shouldAttemptRemoteOperationForPartition:partition]) {
+      MSPaginatedDocuments *cachedDocumentsList = [self.documentStore listWithToken:token
+                                                                          partition:partition
+                                                                       documentType:documentType
+                                                                        baseOptions:baseOptions];
+      if ([self.reachability currentReachabilityStatus] != NotReachable && [[cachedDocumentsList currentPage] items].count == 0) {
+        MSLogInfo([MSData logTag], @"Performing remote operation, since the local list is empty");
+        remoteDocumentBlock(^(MSPaginatedDocuments *_Nonnull remoteDocuments) {
+          // Update local store with the remote list of documents
```
annakocheshkova:
```suggestion

          // Update local store with the remote list of documents
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,74 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+      }
+      completionHandler(cachedDocumentsList);
+      return;
+    }
+
+    // Execute remote operation online and does not have any pending operations.
+    else if ([self shouldAttemptRemoteOperationForPartition:partition]) {
+      MSLogInfo([MSData logTag], @"Performing remote operation");
+      remoteDocumentBlock(^(MSPaginatedDocuments *_Nonnull remoteDocuments) {
+        // Update local store with the remote list of documents
```
annakocheshkova:
```suggestion

        // Update local store with the remote list of documents
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,17 +164,74 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+
+    // Retrieve from cache when offline and when there are pending operations
+    if (![self shouldAttemptRemoteOperationForPartition:partition]) {
+      MSPaginatedDocuments *cachedDocumentsList = [self.documentStore listWithToken:token
+                                                                          partition:partition
+                                                                       documentType:documentType
+                                                                        baseOptions:baseOptions];
+      if ([self.reachability currentReachabilityStatus] != NotReachable && [[cachedDocumentsList currentPage] items].count == 0) {
+        MSLogInfo([MSData logTag], @"Performing remote operation, since the local list is empty");
+        remoteDocumentBlock(^(MSPaginatedDocuments *_Nonnull remoteDocuments) {
```
annakocheshkova: I see that this code is repeated. Can we reuse it?

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT * FROM \"%@\" WHERE \"%@\" = \"%@\"", tableName, kMSPartitionColumnName, token.partition];
+  NSArray *listResult = [self.dbStorage executeSelectionQuery:selectionQuery];
+  NSDate *currentDate = [NSDate date];
+
+  // Parse the documents.
+  for (id documentRow in listResult) {
+
+    // If an expired document is found exclude it from the list and delete it from the local store.
+    long expirationTime = [(NSNumber *)(documentRow[self.expirationTimeColumnIndex]) longValue];
```
annakocheshkova: `[NSNumber numberWithLong]` or `[NSNumber numberWithInt]` might suit here

annakocheshkova: If it's represented as string then you can use `[NSString integerValue]` or `[NSString longValue]` to avoid a cast

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+    NSString *documentId = documentRow[self.documentIdColumnIndex];
+    if (expirationTime != kMSDataTimeToLiveInfinite) {
+      NSDate *expirationDate = [NSDate dateWithTimeIntervalSince1970:expirationTime];
+      if (expirationDate && [expirationDate laterDate:currentDate] == currentDate) {
```
annakocheshkova: use `isEqualToDate`

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+            [NSString stringWithFormat:@"Local document for partition '%@' and document ID '%@' expired at %@, discarding it",
+                                       token.partition, documentId, expirationDate];
+        MSLogWarning([MSData logTag], @"%@", warningMessage);
+
+        // Delete the local document when found to be expired
+        [self deleteWithToken:token documentId:documentId];
+      } else if ([kMSPendingOperationDelete isEqualToString:documentRow[self.pendingOperationColumnIndex]]) {
+
+        // Ignore document, if the pending operation is found to be Delete
+        NSString *warningMessage = [NSString
```
annakocheshkova: Why don't you just use `MSLogError` to format instead of introducing a new variable?

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+            stringWithFormat:
+                @"Local document pending deletion in local storage for partition '%@' and document ID '%@', excluding from the list",
+                token.partition, documentId];
+        MSLogError([MSData logTag], @"%@", warningMessage);
+      } else {
+
+        // Deserialize document.
+        NSString *jsonString = documentRow[self.documentColumnIndex];
+        NSData *jsonData = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+        long lastUpdatedDateValue = [(NSNumber *)documentRow[self.operationTimeColumnIndex] longValue];
```
annakocheshkova: Same here, we can avoid the cast ^

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+                            operation:cachedDocument.pendingOperation];
+      }
+    }
+  }
+
+  // Return an empty list if no documents were found.
+  if (localListItems.count == 0) {
+    NSString *warningMessage =
+        [NSString stringWithFormat:@"Unable to find any document in local store for partition '%@'", token.partition];
+    MSLogWarning([MSData logTag], @"%@", warningMessage);
```
annakocheshkova: Same here, just use `MSLogWarning` to format the string, first parameter is a string for formatting

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,73 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
   return pendingDocuments;
 }
 
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition {
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:partition];
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT count(*) FROM \"%@\" WHERE \"%@\" IS NOT NULL", tableName, kMSPendingOperationColumnName];
+
+  NSArray<NSArray<NSNumber *> *> *result = [self.dbStorage executeSelectionQuery:selectionQuery];
```
annakocheshkova: Why do you use generic `NSArray<NSNumber *>` array as a return type here but `NSArray *` earlier, while the method itself returns `NSArray<NSArray *> *`? I think we need some consistency here.

AppCenterData/AppCenterData/Internal/LocalStore/MSDocumentStore.h
```diff
@@ -87,6 +88,56 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/**
+ * List all the documents from the store and return the list if not expired.
+ *
+ * @param token CosmosDB token.
+ * @param partition The CosmosDB partition key.
+ * @param documentType The document type to list.
+ * @param baseOptions Options for listing and storing the documents.
+ *
+ * @return A MSPaginatedDocuments object. List of documents found in the local store for a particular partition.
```
annakocheshkova:
```suggestion
 * @return List of documents found in the local store for a particular partition.
```
I guess there's no reason to specify the return type in the comment

AppCenterData/AppCenterData/Internal/LocalStore/MSDocumentStore.h
```diff
@@ -87,6 +88,56 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString *)documentId documentType:(Class)documentType;
 
+/**
+ * List all the documents from the store and return the list if not expired.
```
annakocheshkova: > and return the list if not expired

Please specify what is returned if expired

AppCenterData/AppCenterData/Internal/LocalStore/MSDocumentStore.h
```diff
@@ -87,6 +88,56 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition;
+
+/**
+ * Update the local store given a current/new cached document.
+ *
+ * @param token The CosmosDB token.
+ * @param currentCachedDocument The current cached document.
+ * @param newCachedDocument The new document that should be cached.
+ * @param deviceTimeToLive The device time to live for the new cached document.
+ * @param operation The operation being intended (nil - read, CREATE, UPDATE, DELETE).
```
annakocheshkova: Why READ operation is not a contant, too? It would be more explicit. Not blocking, just my opinion :)

AppCenterData/AppCenterData/MSData.m
```diff
@@ -607,6 +573,73 @@ - (void)upsertFromCosmosDbWithPartition:(NSString *)partition
...
+                              id jsonPayload = [NSJSONSerialization JSONObjectWithData:(NSData *)data options:0 error:&deserializeError];
+                              MSDataError *deserializeDataError = nil;
+                              if (!deserializeError && ![MSDocumentUtils isReferenceDictionaryWithKey:jsonPayload
+                                                                                                  key:kMSDocumentsKey
+                                                                                              keyType:[NSArray class]]) {
+                                deserializeDataError = [[MSDataError alloc] initWithErrorCode:MSACDataErrorJSONSerializationFailed
+                                                                                   innerError:deserializeError
+                                                                                      message:@"Can't deserialize documents"];
+                              }
+                              if (deserializeDataError) {
```
annakocheshkova: Move the check inside the previous _if_. `deserializeDataError` can only be assigned inside the previous _if_.

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -601,6 +603,10 @@ - (void)testListWhenOffline {
   self.sut.reachability = reachabilityMock;
   self.sut.httpClient = OCMProtocolMock(@protocol(MSHttpClientProtocol));
   OCMReject([self.sut.httpClient sendAsync:OCMOCK_ANY method:OCMOCK_ANY headers:OCMOCK_ANY data:OCMOCK_ANY completionHandler:OCMOCK_ANY]);
+  self.sut.dataOperationProxy.reachability = reachabilityMock;
+
+  // Mock cached token result.
+  MSTokenResult *tokenResult = [self mockTokenFetchingWithError:nil];
```
annakocheshkova: I can't find where it is used in this test

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,319 @@ - (void)testListWhenOffline {
...
+                [expectation fulfill];
+              }];
+
+  // Then
+  [self waitForExpectationsWithTimeout:1
+                               handler:^(NSError *_Nullable error) {
+                                 if (error) {
+                                   XCTFail(@"Expectation Failed with error: %@", error);
+                                 }
+                               }];
```
annakocheshkova: `[httpClient stopMocking]`

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -622,10 +628,319 @@ - (void)testListWhenOffline {
...
```
annakocheshkova: `[httpClient stopMocking]`

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
clpolet: It's within a block, the formatter won't let this new line be because it's the first line of the block unfortunately.

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
clpolet: SQLITE_INTEGER values results in NSNumber. LGTM.

AppCenterData/AppCenterDataTests/MSDataTests.m
clpolet: Good catch.

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
clpolet: SQLITE_INTEGER values result is NSNumber. LGTM.

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+      } else {
+
+        // Deserialize document.
+        NSString *jsonString = documentRow[self.documentColumnIndex];
+        NSData *jsonData = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+        long lastUpdatedDateValue = [(NSNumber *)documentRow[self.operationTimeColumnIndex] longValue];
+        NSDate *lastUpdatedDate = lastUpdatedDateValue > 0 ? [NSDate dateWithTimeIntervalSince1970:lastUpdatedDateValue] : nil;
+
+        // If operation is NSNull, change it to nil.
+        NSString *pendingOperation = [self checkForNSNull:documentRow[self.pendingOperationColumnIndex]];
```
MehrdadMzfr: >checkForNSNull [](start = 43, length = 14)

how about calling it `safeGetValue`?

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,19 +164,86 @@ - (void)performOperation:(NSString *_Nullable)operation
...
+                                                                          partition:partition
+                                                                       documentType:documentType];
+      completionHandler(documents);
+      return;
+    }
+
+    // Extract first token.
+    MSTokenResult *token = tokensResponse.tokens.firstObject;
+
+    // Retrieve from cache when offline and when there are pending operations
```
clpolet:
```suggestion
    // Retrieve from cache when offline and when there are pending operations.
```

AppCenterData/AppCenterData/Internal/Client/MSDataOperationProxy.m
```diff
@@ -162,19 +164,86 @@ - (void)performOperation:(NSString *_Nullable)operation
...
 #pragma mark Utilities
 
+- (void)performRemoteOperationWithToken:(MSTokenResult *)token
+                            baseOptions:(MSBaseOptions *_Nullable)baseOptions
+                    remoteDocumentBlock:(void (^)(MSPaginatedDocumentsCompletionHandler))remoteDocumentBlock
+                      completionHandler:(MSPaginatedDocumentsCompletionHandler)completionHandler
+
+{
+  remoteDocumentBlock(^(MSPaginatedDocuments *_Nonnull remoteDocuments) {
+    // Update local store with the remote list of documents
```
clpolet:
```suggestion
    // Update local store with the remote list of documents.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+    NSString *documentId = documentRow[self.documentIdColumnIndex];
+    if (expirationTime != kMSDataTimeToLiveInfinite) {
+      NSDate *expirationDate = [NSDate dateWithTimeIntervalSince1970:expirationTime];
+      if (expirationDate && [[expirationDate laterDate:currentDate] isEqualToDate:currentDate]) {
+        NSString *warningMessage =
+            [NSString stringWithFormat:@"Local document for partition '%@' and document ID '%@' expired at %@, discarding it",
+                                       token.partition, documentId, expirationDate];
+        MSLogWarning([MSData logTag], @"%@", warningMessage);
+
+        // Delete the local document when found to be expired
```
clpolet:
```suggestion
        // Delete the local document when found to be expired.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -189,6 +189,106 @@ - (MSDocumentWrapper *)readWithToken:(MSTokenResult *)token documentId:(NSString
...
+        [self deleteWithToken:token documentId:documentId];
+      } else if ([kMSPendingOperationDelete isEqualToString:documentRow[self.pendingOperationColumnIndex]]) {
+
+        // Ignore document, if the pending operation is found to be Delete
```
clpolet:
```suggestion
        // Ignore document, if the pending operation is found to be Delete.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,77 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
   return pendingDocuments;
 }
 
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition {
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:partition];
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT count(*) FROM \"%@\" WHERE \"%@\" IS NOT NULL", tableName, kMSPendingOperationColumnName];
+
```
clpolet:
```suggestion
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,77 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
...
+  NSArray<NSArray<NSNumber *> *> *result = [self.dbStorage executeSelectionQuery:selectionQuery];
+  NSUInteger count = (result.count > 0) ? result[0][0].unsignedIntegerValue : 0;
+
+  // If empty list
```
clpolet:
```suggestion
  // If empty list.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,77 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
...
+  // Get effective device time to live.
+  NSInteger deviceTimeToLive = baseOptions ? baseOptions.deviceTimeToLive : kMSDataTimeToLiveDefault;
+
+  // Parse the documents.
+  NSArray<MSDocumentWrapper *> *remoteListItems = [[documentList currentPage] items];
+  for (MSDocumentWrapper *document in remoteListItems) {
+    MSLogInfo([MSData logTag], @"Updating/inserting document into local storage (partition: %@, id: %@)", document.partition,
+              document.documentId);
+
+    // cache only valid documents
```
clpolet:
```suggestion
    // cache only valid documents.
```

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
clpolet: That `NSArray *` earlier if it's referring to `NSArray *listResult` is for a single row that contains multiple types so it makes sense to be generic. The method returns a generic parameter because it can't make assumption on how it'll be used, what query...But in this code since the the query is a count we now we are expecting a number so that sounds correct to me. :)

AppCenterData/AppCenterData/Internal/LocalStore/MSDBDocumentStore.m
```diff
@@ -263,4 +363,76 @@ + (NSString *)tableNameForPartition:(NSString *)partition {
   return pendingDocuments;
 }
 
+- (BOOL)hasPendingOperationsForPartition:(NSString *)partition {
+  NSString *tableName = [MSDBDocumentStore tableNameForPartition:partition];
+  NSString *selectionQuery =
+      [NSString stringWithFormat:@"SELECT count(*) FROM \"%@\" WHERE \"%@\" IS NOT NULL", tableName, kMSPendingOperationColumnName];
```
clpolet:
```suggestion
      [NSString stringWithFormat:@"SELECT COUNT(*) FROM \"%@\" WHERE \"%@\" IS NOT NULL", tableName, kMSPendingOperationColumnName];
```

PR #1666, opened 2019-06-17
Recovery logic for corrupted database

AnastasiaKubova:
* [x] Did you add unit tests?

## Description

Delete and recreate database that cannot be opened.

## Related PRs or issues

[AB#63196](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63196)


AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -38,32 +38,14 @@ - (instancetype)initWithSchema:(MSDBSchema *)schema version:(NSUInteger)version
...
-      [MSDBStorage setVersion:version inOpenedDatabase:db];
-      sqlite3_close(db);
-    };
+    int result = [self initializeDatabaseWithSchema:schema version:version filename:filename];
+    if (result == SQLITE_CORRUPT || result == SQLITE_NOTADB) {
+      [self dropDatabase];
+      result = [self initializeDatabaseWithSchema:schema version:version filename:filename];
+    }
+    if (result != SQLITE_OK) {
+      MSLogError([MSAppCenter logTag], @"Failed to initialize database.");
```
jwallra: I'd like to start including the SQLITE result code in the error log

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -82,8 +103,16 @@ - (int)executeQueryUsingBlock:(MSDBStorageQueryBlock)callback {
...
+  // Do not proceed with the query if the database is corrupted.
+  if (result == SQLITE_CORRUPT || result == SQLITE_NOTADB) {
+    sqlite3_close(db);
+    return result;
+  }
+
+  // Log a warning if max page count can't be set.
   if (result != SQLITE_OK) {
-    MSLogError([MSAppCenter logTag], @"Failed to open database with specified maximum size constraint.");
+    MSLogWarning([MSAppCenter logTag], @"Failed to open database with specified maximum size constraint.");
```
jwallra: Why was this downgraded to a warning?  Also I think we should include the result code in the log here

jwallra: Or if there are logs where the method is used I suppose we could put the result there.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -72,6 +54,45 @@ - (instancetype)initWithVersion:(NSUInteger)version filename:(NSString *)filenam
...
+  if (result != SQLITE_OK) {
+    sqlite3_close(db);
+    return result;
+  }
+
+  // Create table.
+  if (schema) {
+    result = [MSDBStorage createTablesWithSchema:schema inOpenedDatabase:db];
+    if (result != SQLITE_OK) {
+      sqlite3_close(db);
```
jwallra: We need to log an error here now

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -196,15 +229,19 @@ + (NSDictionary *)columnsIndexes:(MSDBSchema *)schema {
 }
 
 + (BOOL)tableExists:(NSString *)tableName inOpenedDatabase:(void *)db {
+  return [MSDBStorage tableExists:tableName result:nil inOpenedDatabase:db];
+}
+
++ (BOOL)tableExists:(NSString *)tableName result:(int *)result inOpenedDatabase:(void *)db {
```
jwallra: I think the result parameter should be last

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -269,11 +308,21 @@ + (int)executeNonSelectionQuery:(NSString *)query inOpenedDatabase:(void *)db {
...
 }
 
-+ (NSArray<NSArray *> *)executeSelectionQuery:(NSString *)query inOpenedDatabase:(void *)db {
++ (NSArray<NSArray *> *)executeSelectionQuery:(NSString *)query
+                             inOpenedDatabase:(void *)db {
+  return [self executeSelectionQuery:query result:nil inOpenedDatabase:db];
+}
+
++ (NSArray<NSArray *> *)executeSelectionQuery:(NSString *)query
+                                       result:(int *)result
```
jwallra: I think the result parameter should be last

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -126,15 +130,18 @@ - (void)testTableExists {
 
 - (void)testVersion {
   [self.sut executeQueryUsingBlock:^int(void *db) {
+
```
jwallra:
```suggestion
```

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -694,6 +702,30 @@ - (void)testDatabaseThatIsAutoVacuumedNotManuallyVacuumedWhenInitialized {
...
 
+- (void)testDatabaseThatFileWasCorrupted {
+
+  // If
+  const char *validFileHeader = "SQLite format 3";
+  NSURL *fileURL = [MSUtility fullURLForPathComponent:kMSTestDBFileName];
+  NSData *data = [NSData dataWithContentsOfURL:fileURL];
+  
+  // Check that database file is valid.
+  XCTAssertEqual(strcmp((const char *)data.bytes, validFileHeader), 0);
```
jwallra: Any reason why we have to use `strcmp`?  Why wouldn't this work:

```suggestion
  XCTAssertEqual((const char *)data.bytes, validFileHeader);
```

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -694,6 +702,30 @@ - (void)testDatabaseThatIsAutoVacuumedNotManuallyVacuumedWhenInitialized {
...
+  *((unsigned int *)mutabledata.mutableBytes) = 0xDEADBEEF;
+  [mutabledata writeToURL:fileURL atomically:YES];
+  
+  // When
+  self.sut = [[MSDBStorage alloc] initWithSchema:self.schema version:0 filename:kMSTestDBFileName];
+
+  // Then
+  // The database should be recreated.
+  data = [NSData dataWithContentsOfURL:fileURL];
+  XCTAssertEqual(strcmp((const char *)data.bytes, validFileHeader), 0);
```
jwallra:
```suggestion
  XCTAssertEqual((const char *)data.bytes, validFileHeader);
```

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: It is already printed in the previous line

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: The code is already printed. "warning" because in this case, the main query will be executed

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: Already printed, no need to duplicate

AppCenter/AppCenterTests/MSDBStorageTests.m
MatkovIvan: `XCTAssertEqual` doesn't compare C-strings

See https://developer.apple.com/documentation/xctest/xctassertequal

> An expression of **C scalar type**.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
jaelim-ms: We probably need to move error log to the method that actually taking a request from a caller to get rid of any confusions. I fully understand the @jwallra 's comment, we can probably move the error message to this method unless it would be a duplication of error messages.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
jaelim-ms: I think it should stay as an error level. If the operation fails, that means it won't use a limit that customers set.

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
```diff
@@ -72,6 +54,45 @@ - (instancetype)initWithVersion:(NSUInteger)version filename:(NSString *)filenam
   return (self = [self initWithSchema:nil version:version filename:filename]);
 }
 
+- (int)initializeDatabaseWithSchema:(MSDBSchema *)schema version:(NSUInteger)version filename:(NSString *)filename {
```
jaelim-ms: We should be careful on naming a method with `init` or `initialize`. It seems an initializer which isn't actually. @clpolet is a naming expert, any suggestions?

AppCenter/AppCenter/Internals/Storage/MSDBStoragePrivate.h
```diff
@@ -49,6 +49,13 @@ static const long kMSDefaultDatabaseSizeInBytes = 10 * 1024 * 1024;
  */
 - (int)executeQueryUsingBlock:(MSDBStorageQueryBlock)block;
 
+/**
+ * Create table with schema.
+ * @param schema Database schema.
```
jaelim-ms: Add an empty line between description and param.

AppCenter/AppCenterTests/MSDBStorageTests.m
```diff
@@ -694,6 +701,70 @@ - (void)testDatabaseThatIsAutoVacuumedNotManuallyVacuumedWhenInitialized {
...
+  // When
+  self.sut = [[MSDBStorage alloc] initWithSchema:self.schema version:1 filename:kMSTestDBFileName];
+
+  // Then
+  // The database should be recreated.
+  data = [NSData dataWithContentsOfURL:fileURL];
+  XCTAssertEqual(strcmp((const char *)data.bytes, validFileHeader), 0);
+}
+
+- (void)testInitWithSchemaResultIsNotOk {
```
jaelim-ms:
```suggestion
- (void)testInitWithSchemaWithErrorResult {
```

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
michaelsmitha: `configureDatabaseWithSchema`?

AppCenter/AppCenter/Internals/Storage/MSDBStorage.m
MatkovIvan: Let's do not remove the log from the common function. The additional log has been added.

PR #1672, opened 2019-06-18
Feature/improve merge build

jaelim-ms:
Update build script to get rid of apple_version.txt in Azure Blob


jaelim-ms:
Please do not merge this one. I need to clean up environment variable before merging this one for final test.

PR #1675, opened 2019-06-21
Add setting dialog for user Id

jinquan2019:

* [ ] Did you add unit tests?

## Description

Add the setting dialog to set user id
屏幕快照 2019-06-21 17 09 26


Sasquatch/Sasquatch/ViewControllers/MSMainViewController.swift
```diff
@@ -286,6 +307,14 @@ class MSMainViewController: UITableViewController, AppCenterProtocol {
     return acProdLogUrl
     #endif
   }
+    
+  func showUserId () -> String {
+    let userId = UserDefaults.standard.string(forKey: kMSUserIdKey) ?? "Unset";
+    if (userId.isEmpty) {
+        return "Empty";
```
MatkovIvan: Let's mark it somehow that it isn't real value

PR #1679, opened 2019-06-26
GetNextPage when offline

MehrdadMzfr:

* [x] Did you add unit tests?

## Description
There are 2 APIs related to this change.
1- `- (BOOL)hasNextPage;`
the implementation of this API remains the same and as long as continuation token exists, it return true. this change reverts this API.

2- `- (void)nextPageWithCompletionHandler:(void (^)(MSPage *page))completionHandler;`
for this API when device is offline, it returns a Page containing an error that indicates next page is not available cause device is offline. Also, we keep the continuation token around (won't clean it up when called network offline)

### DIFF
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -403,9 +403,10 @@ static dispatch_once_t onceToken;
       dataError = [self generateDisabledError:@"list" documentId:nil];
     } else if (![MSDocumentUtils isSerializableDocument:documentType]) {
       dataError = [self generateInvalidClassError];
-    } else if ([self.reachability currentReachabilityStatus] == NotReachable) {
+    } else if (continuationToken && [self.reachability currentReachabilityStatus] == NotReachable) {
 
-      // If offline, return an error.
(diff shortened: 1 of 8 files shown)


AppCenterData/AppCenterData/MSData.m
```diff
@@ -403,9 +403,18 @@ - (void)listDocumentsWithType:(Class)documentType
       dataError = [self generateDisabledError:@"list" documentId:nil];
     } else if (![MSDocumentUtils isSerializableDocument:documentType]) {
       dataError = [self generateInvalidClassError];
+    } else if ([self.reachability currentReachabilityStatus] == NotReachable) {
```
yoannchaudet: Doesn't it also prevent fetching the first page of data when we are offline (i.e. from the cache)?

yoannchaudet: I think this logic would be fine only if we are testing for `continuationToken` being not `nil` at the same time. And we are probably missing some unit tests if that did not break anything.

MehrdadMzfr: makes sense, updated the code, if continuation token is provided and device if offline, then it returns error. 
otherwise, it return the documents (including the first page)

PR #1706, opened 2019-07-23
Fix warnings in publc headers caused by missing lifetime attributes f…

MichaelBuckley:
…or object properties

## Description

This pull requests fixes warnings in public headers so clients including the AppCenter, AppCenterAnalytics, and AppCenterCrashes frameworks on Mac do not get warnings while compiling.

The warnings were caused by the use of `@property(nonatomic)`, without specifying a lifetime attribute, for object properties. Objective-C defaults to `assign` when there is no lifetime attribute, which is not appropriate for Objective-C objects.

## Related PRs or issues

There are no related PRs or issues that I could find.

## Misc

Because I wanted to keep this pull request small, I only updated the public headers. There are lots of internal headers and implementation files which also use `@property(nonatomic)` in this way. If you would like me to update those as well, I would be happy to.

This is my first pull request to this project, and although I attempted to run the local tests, I was unable to get them to build due to various linking errors in Xcode 10.3. I may be missing a step. Is there a script or other command I should be using to run the tests? If not, I can file issues on the linker errors I get when trying to run the tests in Xcode.

Thank you for your consideration.

### DIFF
--- a/AppCenter/AppCenter/MSLog.h
+++ b/AppCenter/AppCenter/MSLog.h
@@ -15,7 +15,7 @@
 /**
  * Log timestamp.
  */
-@property(nonatomic) NSDate *timestamp;
+@property(nonatomic, strong) NSDate *timestamp;
 
 /**
  * A session identifier is used to correlate logs together. A session is an abstract concept in the API and is not necessarily an analytics
@@ -36,13 +36,13 @@
 /**
  * Device properties associated to this log.
  */
-@property(nonatomic) MSDevice *device;
+@property(nonatomic, strong) MSDevice *device;
 
 /**
  * Transient object tag. For example, a log can be tagged with a transmission target. We do this currently to prevent properties being
  * applied retroactively to previous logs by comparing their tags.
  */
-@property(nonatomic) NSObject *tag;
+@property(nonatomic, strong) NSObject *tag;
 
 /**
  * Checks if the object's values are valid.
--- a/AppCenter/AppCenter/Model/MSLogWithProperties.h
+++ b/AppCenter/AppCenter/Model/MSLogWithProperties.h
@@ -10,6 +10,6 @@
 /**
  * Additional key/value pair parameters. [optional]
  */
-@property(nonatomic, copy) NSDictionary<NSString *, NSString *> *properties;
+@property(nonatomic, strong) NSDictionary<NSString *, NSString *> *properties;
 
 @end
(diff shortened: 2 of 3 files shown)


richardpineo:
Not an editor / member of the AppStudio team but I had the same issue linking in xcode 10.3. I had to download and use xcode 10.1 to get things to work. I believe the source doesn't support the newest IDE at this time.

jaelim-ms:
There are some of known issues when the SDK is built in Xcode 10.2+. Please use 10.1 for your contribution until the issue is resolved.

jaelim-ms:
@MichaelBuckley Thanks for your contribution. For your information, `strong` would be the default attribute for memory management since the SDK is under ARC.
You can find Apple doc [here](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW4)

MichaelBuckley:
@jaelim-ms Thanks for the explanation. I had missed that in the ARC docs. Most of our code is using ARC at this point, but we are including the AppCenter headers from a non-ARC file. This PR would fix the warnings for projects not yet using ARC, but it sounds like updating the internal headers and implementation is unnecessary.

I will track down a copy of Xcode 10.1 for future submissions. Thanks!

jaelim-ms:
Thanks for your contribution again @MichaelBuckley. We will merge the PR when it passes the CI build. Looks it had some hiccups on CI.

jaelim-ms:
@MichaelBuckley It is a real error in tests. Would you please run tests and check test errors there?
```
Failing tests:
	-[MSLogDBStorageTests testSaveLargeCriticalPriorityLogDoesNotPurgeOldLogs]
	-[MSLogDBStorageTests testSaveLargeNormalPriorityLogDoesNotPurgeOldLogs]
	-[MSLogDBStorageTests testUsesMillisecondsInSaveQuery]
** TEST FAILED **
```

MichaelBuckley:
@jaelim-ms Sorry about breaking the tests. It's second nature for me to specify classes with mutable counterparts (like NSDictionary) as `copy`, but it appears the tests were relying on modifying the property after assignment.

The tests should all be fixed in the commit I just pushed. Now that I know how to run them, I won't submit further PRs with failing tests. Thanks for your patience.

jaelim-ms:
Just curiosity, are you only using Analytics? There are many public headers in other modules such as Crashes, Push, Data, Auth, etc., I wonder why you only fixed headers in App Center core and Analytics. 😄

MichaelBuckley:
We're also using Crashes. I only fixed the warnings that appeared in our project when importing those three frameworks. For whatever reason, Crashes didn't throw any warnings.

If you'd like, I can take a look at the other frameworks when I have some time.

PR #1708, opened 2019-07-29
Handle memory warning

MatkovIvan:
* [x] Did you add unit tests?

## Related PRs or issues

[AB#65503](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65503)

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.h
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.h
@@ -94,7 +94,7 @@ typedef NS_ENUM(NSUInteger, MSUserConfirmation) {
  *
  * @return Returns YES is the app received memory warning in the last session.
  */
-+ (BOOL)hadMemoryWarningInLastSession;
++ (BOOL)hasReceivedMemoryWarningInLastSession;
 
 /**
  * Provides details about the crash that occurred in the last app session
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.h
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.h
@@ -90,6 +90,13 @@ typedef NS_ENUM(NSUInteger, MSUserConfirmation) {
 + (BOOL)hasCrashedInLastSession;
 
 /**
+ * Check if the app received memory warning in the last session.
+ *
+ * @return Returns YES is the app received memory warning in the last session.
+ */
++ (BOOL)hadMemoryWarningInLastSession;
+
+/**
  * Provides details about the crash that occurred in the last app session
  */
 + (nullable MSErrorReport *)lastSessionCrashReport;
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -62,7 +62,7 @@ static unsigned int kMaxAttachmentsPerCrashReport = 2;
 /**
  * Key for a low memory warning in the last session.
  */
-static NSString *const kMSAppDidReceiveLowMemory = @"MSAppDidReceiveLowMemory";
+static NSString *const kMSAppDidReceiveMemoryWarning = @"MSAppDidReceiveMemoryWarning";
 
 std::array<MSCrashesBufferedLog, ms_crashes_log_buffer_size> msCrashesLogBuffer;
 
@@ -139,6 +139,15 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
 @property BOOL didCrashInLastSession;
 
 /**
+ * Indicates if the app did receive a low memory warning in the last session.
+ * It may happen that low memory warning where send but couldn't be logged, since iOS killed the app before updating the flag in the filesystem did complete.
+ * This property may be true in case of low memory kills, but it doesn't have to be! Apps can also be killed without the app ever receiving a low memory warning.
+ * Also the app could have received a low memory warning, but the reason for being killed was actually different.
+ *
+ * @warning This property only has a correct value, once the sdk has been properly initialized!
+ */
+@property BOOL didReceiveMemoryWarningInLastSession;
+/**
  * Indicates if the delayedProcessingSemaphore will need to be released anymore. Useful for preventing overflows.
  */
 @property BOOL shouldReleaseProcessingSemaphore;
@@ -207,6 +216,10 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
 
 + (BOOL)hasCrashedInLastSession {
   return [[MSCrashes sharedInstance] didCrashInLastSession];
+}
+
++ (BOOL)hadMemoryWarningInLastSession {
+  return [[MSCrashes sharedInstance] didReceiveMemoryWarningInLastSession];
 }
 
 + (void)setUserConfirmationHandler:(_Nullable MSUserConfirmationHandler)userConfirmationHandler {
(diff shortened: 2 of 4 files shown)


AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -1129,7 +1183,7 @@ - (void)handleUserConfirmation:(MSUserConfirmation)userConfirmation {
      * Always send logs. Set the flag YES to bypass user confirmation next time.
      * Continue crash processing afterwards.
      */
-    [MS_USER_DEFAULTS setObject:[[NSNumber alloc] initWithBool:YES] forKey:kMSUserConfirmationKey];
+    [MS_USER_DEFAULTS setObject:@YES forKey:kMSUserConfirmationKey];
```
annakocheshkova: 😄 good catch

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -1203,6 +1205,49 @@ - (void)testCrashesSetCorrectUserIdToLogs {
...
 
+#if !TARGET_OS_OSX
+
+- (void)testObserverAddedNotOsxNotExtension {
+
+  // When
+  [self.sut applyEnabledState:YES];
+
+  // Then
+  OCMVerify([MS_NOTIFICATION_CENTER addObserver:self.sut
```
annakocheshkova: Maybe we should check that it actually responds to selctor?

Jamminroot: Well that was initial implementation 🙄

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -402,6 +450,12 @@ - (void)clearContextHistoryAndKeepCurrentSession {
   [[MSUserIdContext sharedInstance] clearUserIdHistory];
 }
 
+#pragma mark - Application life cycle
+
+- (void)didReceiveMemoryWarning:(NSNotification *)__unused notification {
+  [MS_USER_DEFAULTS setObject:@YES forKey:kMSAppDidReceiveLowMemory];
```
annakocheshkova: kMSAppDidReceiveLowMemory → kMSAppDidReceiveLowMemoryWarning

AppCenterCrashes/AppCenterCrashes/MSCrashes.h
```diff
@@ -89,6 +89,13 @@ typedef NS_ENUM(NSUInteger, MSUserConfirmation) {
  */
 + (BOOL)hasCrashedInLastSession;
 
+/**
+ * Check if the app received memory warning in the last session.
+ *
+ * @return Returns YES is the app received memory warning in the last session.
+ */
++ (BOOL)hadMemoryWarningInLastSession;
```
jaelim-ms: Is the name confirmed by the PM?

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -1203,6 +1205,54 @@ - (void)testCrashesSetCorrectUserIdToLogs {
   [settings stopMocking];
 }
 
+#if !TARGET_OS_OSX
+
+- (void)testObserverAddedNotOsxNotExtension {
```
jaelim-ms: You can add one more test for extension by mocking `MS_IS_APP_EXTENSION`.
Then the name of this test would be like `testObserverAdded` since the macro explains the test isn't being executed for macOS platform.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -1203,6 +1205,54 @@ - (void)testCrashesSetCorrectUserIdToLogs {
...
+  OCMVerify([defaultCenterMock addObserver:self.sut
+                                  selector:[OCMArg anySelector]
+                                      name:UIApplicationDidReceiveMemoryWarningNotification
+                                    object:nil]);
+  [defaultCenterMock stopMocking];
+}
+
+#endif
+
+- (void)testMemoryPressureSourceSetUpAndClearedInExtension {
```
jaelim-ms: Is this test only for extension?
```suggestion
- (void)testMemoryPressureSourceSetUpAndClearedInExtensionAndMacOS {
```

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -127,6 +138,16 @@ @interface MSCrashes ()
  */
 @property BOOL didCrashInLastSession;
 
+/**
+ * Indicates if the app did receive a low memory warning in the last session.
+ * It may happen that low memory warning where send but couldn't be logged, since iOS killed the app before updating the flag in the
+ * filesystem did complete. This property may be true in case of low memory kills, but it doesn't have to be! Apps can also be killed
+ * without the app ever receiving a low memory warning. Also the app could have received a low memory warning, but the reason for being
+ * killed was actually different.
```
michaelsmitha: please replace this with some grammatical changes/clarifications

```
Indicates that the app received a low memory warning in the last session.
It is possible that a low memory warning was sent but couldn't be logged if iOS killed the app before updating the flag in the filesystem.
Apps can also be killed without receiving a low memory warning, or receive the warning, but crash for another reason.
```

AppCenterCrashes/AppCenterCrashes/MSCrashes.h
achocron: +1 - since the API we are referring to is called `didReceiveMemoryWarning`, maybe it would make sense to use `didReceiveMemoryWarningInLastSession` or `receivedMemoryWarningInLastSession`

guperrot: We have `hasCrashedInLastSession` in all platforms.
The most consistent would be `hasReceivedMemoryWarningInLastSession` according to me.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: @michaelsmitha ok.
(the current version is from HockeyApp 😃)

https://github.com/bitstadium/HockeySDK-https://github.com/bitstadium/HockeySDK-iOS/blob/dabf7f9f66b27a4dca089ad6dbbece04bae7d7dd/Classes/BITCrashManager.h#L369-L387

michaelsmitha: @MatkovIvan haha gotcha, it was really hard to read :)

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -273,6 +297,12 @@ - (instancetype)init {
 - (void)applyEnabledState:(BOOL)isEnabled {
   [super applyEnabledState:isEnabled];
 
+#if !TARGET_OS_OSX
```
jaelim-ms: Why is it here? Should we only remove the observer when it is being disabled?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -367,6 +425,16 @@ - (void)startWithChannelGroup:(id<MSChannelGroupProtocol>)channelGroup
   [super startWithChannelGroup:channelGroup appSecret:appSecret transmissionTargetToken:token fromApplication:fromApplication];
   [self.channelGroup addDelegate:self];
   [self processLogBufferAfterCrash];
+  if (self.isEnabled) {
```
jaelim-ms: Can you extract this as a method?
`[self processMemoryWarningInLastSession]` should be more readable in the method. 😄

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: It's the same pattern as here
https://github.com/microsoft/appcenter-sdk-apple/blob/4bea6348f2f0b4c51898c86d50302e59888931f1/AppCenter/AppCenter/MSAppCenter.m#L553-L556

jaelim-ms: It isn't consistent. I can see this in `MSAuth`.
https://github.com/microsoft/appcenter-sdk-apple/blob/develop/AppCenterAuth/AppCenterAuth/MSAuth.m#L126

jaelim-ms: This isn't a big deal but it doesn't need to remove the observer when it is being enabled. I don't want to block this PR because of this one but we should revisit this one in `MSAppCenter`.

Issue #1705, opened 2019-07-22
`+[MSUtility(Application) sharedAppState]` must be called from the main thread


tonyarnold:
https://github.com/microsoft/appcenter-sdk-apple/blob/d3eb6c942c39be49a6363f28c86f799eb5c1f5c0/AppCenter/AppCenter/Internals/Util/MSUtility%2BApplication.m#L50

This line is tripping the main thread UI checker when called from `-[MSCrashes startDelayedCrashProcessing]`: 

```
=================================================================
Main Thread Checker: UI API called on a background thread: -[NSApplication isHidden]
PID: 61170, TID: 516248, Thread name: (none), Queue name: com.apple.root.default-qos, QoS: 0
Backtrace:
4   Reveal                              0x0000000100701cb4 +[MSUtility(Application) sharedAppState] + 53
5   Reveal                              0x000000010066e60c -[MSCrashes startCrashProcessing] + 61
6   Reveal                              0x000000010066e512 __40-[MSCrashes startDelayedCrashProcessing]_block_invoke + 33
7   libdispatch.dylib                   0x00007fff6c79c5de _dispatch_client_callout + 8
8   libdispatch.dylib                   0x00007fff6c79e790 _dispatch_continuation_pop + 414
9   libdispatch.dylib                   0x00007fff6c7ae430 _dispatch_source_invoke + 2084
10  libdispatch.dylib                   0x00007fff6c79e2d2 _dispatch_queue_override_invoke + 421
11  libdispatch.dylib                   0x00007fff6c7aa8e5 _dispatch_root_queue_drain + 326
12  libdispatch.dylib                   0x00007fff6c7ab025 _dispatch_worker_thread2 + 92
13  libsystem_pthread.dylib             0x00007fff6c9fb699 _pthread_wqthread.cold.1 + 67
14  libsystem_pthread.dylib             0x00007fff6c9f5856 _pthread_wqthread + 203
15  libsystem_pthread.dylib             0x00007fff6c9f5717 start_wqthread + 15
```

jwargo:
@tonyarnold thank you for sharing this with us, we’ll investigate.

rudyrichter:
this access probably needs to be wrapped with a dispatch_sync(dispatch_get_main_queue(), )

tonyarnold:
That was my thought, too, Rudy. I’ll pull together a PR later on if nobody beats me to it 👍🏻

AnastasiaKubova:
[...]

PR #1707, opened 2019-07-24
Ensure that -[NSApplication isHidden] is called from the main thread

tonyarnold:
Fixes #1705

PR #1712, opened 2019-07-30
Fix main thread checker's warning during crash processing on macOS

MatkovIvan:
* [ ] Did you add unit tests?

## Related PRs or issues

#1705
[AB#65127](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65127)

### DIFF
--- a/AppCenter/AppCenter/Internals/Util/MSUtility+Application.m
+++ b/AppCenter/AppCenter/Internals/Util/MSUtility+Application.m
@@ -47,6 +47,11 @@ NSString *MSUtilityApplicationCategory;
 
 #if TARGET_OS_OSX
 + (MSApplicationState)sharedAppState {
+
+  // UI API (isHidden) cannot be called from a background thread.
+  if (![NSThread isMainThread]) {
+    return MSApplicationStateUnknown;
+  }
   return [[MSUtility sharedApp] isHidden] ? MSApplicationStateBackground : MSApplicationStateActive;
 }
 #else
--- a/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
@@ -652,7 +652,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSCra
    * FIXME: There is no life cycle for app extensions yet so force start crash processing until then.
    * Note that macOS cannot access the application state from a background thread, so crash processing will start without this check.
    *
-   * Also force start crash processing when automatic processing is disabled. Though it sounds counterintuitive, his is important because
+   * Also force start crash processing when automatic processing is disabled. Though it sounds counterintuitive, this is important because
    * there are scenarios in some wrappers (i.e. ReactNative) where the application state is not ready by the time crash processing needs to
    * happen.
    */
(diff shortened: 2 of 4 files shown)


AppCenter/AppCenter/Internals/Util/MSUtility+Application.m
```diff
@@ -47,6 +47,11 @@ + (UIApplication *)sharedApp {
 
 #if TARGET_OS_OSX
 + (MSApplicationState)sharedAppState {
+
+  // UI API (isHidden) cannot be called from a background thread.
+  if (![NSThread isMainThread]) {
```
jaelim-ms: Can we get the state from UI thread? I know it isn't a good idea because it will be a synchronous call and block the thread.

achocron: Nit: invert if statement

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -644,11 +649,14 @@ - (void)startDelayedCrashProcessing {
 - (void)startCrashProcessing {
 
   /*
-   * FIXME: There is no life cycle for app extensions yet so force start crash processing until then. Also force start crash processing when
-   * automatic processing is disabled. Though it sounds counterintuitive, this is important because there are scenarios in some wrappers
-   * (i.e. RN) where the application state is not ready by the time crash processing needs to happen.
+   * FIXME: There is no life cycle for app extensions yet so force start crash processing until then.
+   * Note that macOS cannot access the application state from a background thread, so crash processing will start without this check.
+   *
+   * Also force start crash processing when automatic processing is disabled. Though it sounds counterintuitive, his is important because
```
achocron:
```suggestion
   * Also force start crash processing when automatic processing is disabled. Though it sounds counterintuitive, this is important because
```

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # App Center SDK for iOS, macOS and tvOS Change Log
 
-## Version 2.2.1
+## Version 2.2.1 (Under development)
+
+### App Center Crashes
+
+* **[Fix]** Fix main thread checker's warning during crash processing on macOS.
```
achocron: To be honest, I'm not exactly sure what is meant by `main thread checker` so I guess this will be confusing for customers, too.

AppCenter/AppCenter/Internals/Util/MSUtility+Application.m
MatkovIvan: No, main thread can be blocked by waiting this so queues doesn't work here.

See https://github.com/microsoft/appcenter-sdk-apple/pull/1707#issuecomment-516429117 with my variants

CHANGELOG.md
tonyarnold: Main Thread Checker is a runtime diagnostic provided by Xcode:

image

I'm pretty sure it's on by default in new projects.

achocron: Ah got it, I wasn't aware of this setting! Thanks :)

AppCenter/AppCenter/Internals/Util/MSUtility+Application.m
```diff
@@ -47,6 +47,11 @@ + (UIApplication *)sharedApp {
...
+    return MSApplicationStateUnknown;
```
clpolet: The problem is now the caller has to make sure to be running on the main thread to have a reliable result. Shouldn't we make it async and dispatch to main thread on macOS? It's more change but less error prone in the usage of that API.

MatkovIvan: > Shouldn't we make it async and dispatch to main thread on macOS?

No. The main thread is blocked at that moment.

It's fine to return `MSApplicationStateUnknown` at this case, since it's will happen only in one place where we can proceed without this check. I've left the comment there.

PR #1717, opened 2019-08-05
Enable MSAL logging

jaelim-ms:
* [ ] Did you add unit tests?

## Description

Enabling MSAL logging in App Center Auth.

## Related PRs or issues

[AB#65915](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65915)


AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -262,6 +264,23 @@ - (void)signOut {
 
 #pragma mark - Private methods
 
+- (void)enableMSALLogging {
+  MSALGlobalConfig.loggerConfig.logLevel = MSALLogLevelVerbose;
+  [MSALGlobalConfig.loggerConfig setLogCallback:^(MSALLogLevel level, NSString *message, BOOL containsPII) {
```
achocron: PII -> Pii?

jaelim-ms: I'm using the same convention as MSAL defined.
We usually don't change this kind of things, do we?
[MSAL doc](https://github.com/AzureAD/microsoft-authentication-library-for-objc/wiki/Logging#logging-callback)

jaelim-ms:
Running unit test individually is fine but it fails on running entire `MSAuthTests`. It is interfered by something else when `setUp` instantiate `MSAuth`. Looking into it now.

AppCenterAuth/AppCenterAuth/MSAuth.m
guperrot: camelCase is the first letter of every word upper case, I and I are initials for 2 separate words @achocron , thus to me it's PII.

AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -262,6 +264,23 @@ - (void)signOut {
...
+    if (!containsPII) {
```
guperrot: are we missing something important? if yes any way to obfuscate?

jaelim-ms: Basically we don't know what should be obfuscate. We can figure this out but I think this would be good for the first step.

AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -262,6 +264,23 @@ - (void)signOut {
 
 #pragma mark - Private methods
 
+- (void)enableMSALLogging {
+  MSALGlobalConfig.loggerConfig.logLevel = MSALLogLevelVerbose;
```
jwallra: Would it make sense to match the MSAL log level to the App Center log level?

jaelim-ms: Then it would be a bigger work. We need to add a delegate to `MSAppCenter` to listen to it whenever the log level changed. Log level can be changed anytime.

jwallra: I don't know how noisy MSAL is but it might be worth it.  If someone sets their AC log level very low and then MSAL starts spitting out tons of logs when they just wanted some basic ones, then it would be a bad experience

jwallra: Also, for servicing and debugging, it would mean the customer has to do 2 different things to get the most detailed logs, but if we did create the delegate then the troubleshooting steps remain the same for everyone

jaelim-ms: We've talked about this one offline. Should be good, right?

PR #1726, opened 2019-08-13
Update MSAL to 0.5.0

clpolet:

## Description

Update MSAL to v.0.5.0.


AppCenterAuth/AppCenterAuthTests/MSAuthTests.m
```diff
@@ -232,9 +232,9 @@ - (void)testCleanUpOnDisabling {
   [[MSAuthTokenContext sharedInstance] setAuthTokenHistory:authTokenHistory];
   [[MSAuthTokenContext sharedInstance] setAuthToken:fakeToken withAccountId:fakeAccountId expiresOn:nil];
   [self.sut setEnabled:YES];
-  id accountMock = OCMPartialMock([MSALAccount new]);
+  id accountMock = OCMClassMock(MSALAccount.class);
```
MatkovIvan: Add stopMocking call for it

clpolet: My understanding is we don't have to stop mocking ourselves if not stubbing class methods. But if it's confusing or seems error prone or I'm wrong I can add the stop mocking.

clpolet: I am trying to keep things simple: Do I mock class methods YES > stop mocking  otherwise don't. But I can be wrong.

jaelim-ms: Need to call stop mocking whenever it uses class mock. It will affect initialize when it instantiate it

clpolet:
Added the "do not merge" tag because we need to do some deep regression/ app upgrade tests on Auth.

AppCenterAuth/AppCenterAuthTests/MSAuthTests.m
```diff
@@ -1197,7 +1200,7 @@ - (void)testRefreshNeededTriggersRefresh {
   OCMStub([authMock configAuthenticationClient]).andDo(^(NSInvocation *__unused invocation) {
     self.sut.clientApplication = self.clientApplicationMock;
   });
-  id accountMock = OCMClassMock([MSALAccount class]);
+  id accountMock = OCMClassMock(MSALAccount.class);
```
jaelim-ms: Any reason of changing this one?

clpolet: Just for consistency as I used the class property for this class mock in other tests.

clpolet:
Cancelling this PR at the moment as it requires to much work for this release cycle. Keeping the branch so it can be resumed later.

AppCenterAuth/AppCenterAuthTests/MSAuthTests.m
clpolet: I am not so sure about that. To me the OCMClassMock just subclass and return an instance of that subclass, nothing really bad. It's only when you stub a class method that OCMMock creates a meta-class and substitute the meta class of the mocked class. Only at that point it can get bad if the mock stubbing the class method is not explicitly stopped or deallocated.

PR #1736, opened 2019-08-21
Add static framework targets to all modules

russelarms:

* [ ] Did you add unit tests?

## Description

Add 3 OS-specific targets to each module (21 total).
For now we only add new static framework targets for iOS, tvOS, macOS, to each module.
Note that the changes should be contained in .xconfig files, otherwise build would fail.
Those targets should contain static librarty target and public headers. Another option is to add sources instead of static library targets - but that would increase complexity of maintaining.

## Related PRs or issues

[AB#66952](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/66952)


AppCenterCrashes/AppCenterCrashesStaticIOS copy-Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MatkovIvan: ` copy-Info.plist` 😄

Scripts/AppCenter/build-appcenter-ios-framework.sh
```diff
@@ -3,7 +3,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-source $(dirname "$0")/../build-ios-framework.sh AppCenter
+source $(dirname "$0")/../build-ios-framework.sh AppCenterFramework
```
MatkovIvan: `AppCenterFramework.framework`? Are you sure?

Config/Static Framework.xcconfig
```diff
@@ -0,0 +1,10 @@
+CURRENT_PROJECT_VERSION = 1
+DEFINES_MODULE = YES
+DYLIB_COMPATIBILITY_VERSION = 1
+DYLIB_CURRENT_VERSION = 1
```
MatkovIvan: Should we use our version instead?

clpolet: Good point, please use our version.

PR #1738, opened 2019-08-26
Carthage support

russelarms:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [X] Are tests passing locally?
* [X] Are the files formatted correctly?
~* [ ] Did you add unit tests?~
* [X] Did you check UI tests on the sample app? They are not executed on CI.
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description
**Note: the old build pipeline _is supposed to fail_. It has not been updated yet.**
**Note: This PR _does not contain_ any Carthage-specific stuff.**

If you want to look at the changes, the following is necessary after checkout:
`git clean -xfd`
`rm -rf ~/Library/Developer/Xcode/DerivedData`

**Introduced changes:**
- The static library targets are deleted;
- A new `<Module> <Platform> Framework` target has been added for each module and platform;
- All the targets names have been renamed in a more clear and straightforward way;
- Each framework target contains implementation and header files;
- Header files are separated between **Public** and **Project** groups as there should be no **Private** headers;
- Build scripts rewritten and fixed;
- `.xcconfig` files also modified and fixed. The current type of the framework (`MACH_O_TYPE`) is **staticlib**, so far dynamic doesn’t work fully. 
- `CrashReporter.framework` has been replaced with a static library for iOS and tvOS. It saves us the trouble of building CrashReporter from a custom script and Carthage customers of having to link it manually.

**The new project structure:**

image

## Related PRs or issues

[AB#66963](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/66963)
[AB#66952](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/66952)
[AB#66955](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/66955)

## Misc

In order for the CI to be able to build this branch with `carthage build`, this PR has some changes that need to be discussed:
https://github.com/microsoft/appcenter-sdk-apple/commit/994f31ca196f78f6b9237ffdbc09053418ae4c82
https://github.com/microsoft/appcenter-sdk-apple/commit/e2b44b6d8e8a36c59600917bd2912f920c6fd48e

PR #1746, opened 2019-09-04
Carthage feature feedback

annakocheshkova:
## Description

Feedback on #1738


Config/macOS.xcconfig
```diff
@@ -4,4 +4,6 @@ ARCHS = x86_64
 VALID_ARCHS = $(ARCHS) i386
 SKIP_INSTALL = YES
 
+TARGETED_DEVICE_FAMILY = 1, 2
```
clpolet: On macOS config? Isn't 1 for iPhones and 2 for iPads? It's probably wrong right?

annakocheshkova: What's the value for macOS then?

jaelim-ms: I don't think TARGETED_DEVICE_FAMILY is used for mac build.

PR #1753, opened 2019-09-10
Bring your own Identity - Implement setAuthToken and delegate in core

jwallra:

* [x] Did you add unit tests?

## Description

Adding the `setAuthTokenDelegate` method for use with bring-your-own identity.

## Misc

[AB#67953](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/67953)


AppCenter/AppCenter/AppCenter.h
```diff
@@ -6,6 +6,8 @@
 #import "MSAbstractLog.h"
 #import "MSAppCenter.h"
 #import "MSAppCenterErrors.h"
+#import "MSAuthProvider.h"
```
guperrot: This was renamed in the spec.

MSAuthTokenDelegate is the new name.

AppCenter/AppCenter/Internals/MSAppCenterInternal.h
```diff
@@ -74,6 +77,8 @@ static NSString *const kMSTrueEnvironmentString = @"1";
 
 @property(nonatomic, copy) void (^maxStorageSizeCompletionHandler)(BOOL);
 
+@property(nonatomic, copy) void (^authProviderCompletionBlock)(NSString *);
```
guperrot: authTokenCompletionHandler is the new spec

AppCenter/AppCenter/Internals/Util/MSJwtClaims.h
```diff
@@ -0,0 +1,51 @@
...
+
+/*
+ * Initializes the MSJwtClaims with the subject and expiration
+ *
+ * @param subject The JWT subject.
+ * @param expirationDate the JWT expiration date.
+ *
+ * @return MSJwtClaims instance.
+ */
+- (instancetype)initWithClaims:(NSString *)subject expirationDate:(NSDate *)expirationDate;
```
guperrot: expirationDate is also a claim, objective C is a bit weird but name has to be initWithSubject

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegateWrapper.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAuthTokenContextDelegate.h"
+#import "MSAuthTokenDelegate.h"
+
+@interface MSAuthTokenContextDelegateWrapper : NSObject <MSAuthTokenContextDelegate>
```
achocron: Please add a comment explaining why this is important

AppCenter/AppCenter/Internals/Util/MSJwtClaims.h
```diff
@@ -0,0 +1,51 @@
...
+ * The JWT subject.
+ */
+@property(nonatomic, copy, readonly) NSString *subject;
+
+/*
+ * The JWT expiration date.
+ */
+@property(nonatomic, copy, readonly) NSDate *expirationDate;
+
+/*
```
achocron:
```suggestion
/**
```

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,42 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
 }
 #endif
 
++ (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
+  [[MSAppCenter sharedInstance] setAuthTokenDelegate:authTokenDelegate];
+}
+
+- (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
```
achocron: does this logic need to be synchronized?

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -33,6 +33,22 @@
 // NSUUID can return this nullified InstallId while creating a UUID from a nil string, we want to avoid this.
 static NSString *const kMSNullifiedInstallIdString = @"00000000-0000-0000-0000-000000000000";
 
+static NSString *const kMSJwtFormat = @"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.%@";
+
+@interface MSJwtHelper : NSObject
++ (NSString *)createJwtWithUserId:(NSString *)userId
+                            expiration:(int)expiration;
+@end
+@implementation MSJwtHelper
```
achocron:
```suggestion

@implementation MSJwtHelper
```

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -975,4 +991,65 @@ - (void)testNoUserIdWhenSetUserIdIsNotCalledInNextVersion {
...
+}
+
+- (void)testSetAuthTokenDelegateExpiredToken {
+  
+  // If
+  // Create a delegate
+  // Create expired auth token in authTokenContext
+  
+  // When
+  [MSAppCenter setAuthTokenDelegate:<#(id<MSAuthTokenDelegate>)#>]
```
achocron: What are these `#` symbols?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,42 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
 }
 #endif
 
++ (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
```
achocron: Where is the set auth token method? just not written yet?

michaelsmitha: This is written now

AppCenter/AppCenter.xcodeproj/project.pbxproj.orig
```diff
@@ -0,0 +1,4171 @@
+// !$*UTF8*$!
```
dhei: *.orig files are generated by Git merge. We don't want to check in such file.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegateWrapper.h
```diff
@@ -0,0 +1,24 @@
...
+
+@property id<MSAuthTokenDelegate> delegate;
+
+@property MSAuthTokenCompletionHandler authTokenCompletionHandler;
+
+/**
+ * Create a delegate wrapper which will call the delegates for both MSAuthTokenContextDelegate and MSAuthTokenDelegate.
+ *
+ * @param authTokenDelegate The Auth token delegate to call when refreshing tokens.
+ * @param authTokenCompletionHandler The complation handler to call after getting refresh tokens.
```
dhei:
```suggestion
 * @param authTokenCompletionHandler The completion handler to call after getting refresh tokens.
```

AppCenter/AppCenter/MSAppCenter.h
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#import <Foundation/Foundation.h>
-
+#import "MSAuthTokenDelegate.h"
 #import "MSConstants.h"
+#import <Foundation/Foundation.h>
```
dhei: Sort imports.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,59 @@
...
+  @try {
+    NSError *error;
+    NSData *claimsPartData = [[NSData alloc] initWithBase64EncodedString:base64ClaimsPart options:0];
+    NSDictionary *claims = [NSJSONSerialization JSONObjectWithData:claimsPartData options:0 error:&error];
+    if ([claims objectForKey:kMSExpirationClaim] == nil || [claims objectForKey:kMSSubjectClaim] == nil) {
+      MSLogError([MSAppCenter logTag], @"Deserialized claims is missing `sub` or `exp`");
+      return nil;
+    }
+
+    // We will treat invalid expiration times as an interval of 0 because there is not an easy way to determine the
```
dhei: What does it mean by `an interval of 0`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegateWrapper.h
jwallra:
```suggestion
/**
 * MSAuthTokenContext:addDelegate takes a MSAuthTokenContextDelegate, but our new MSAuthTokenDelegate has a different form, 
 * so we need to crate a wrapper which conforms to MSAuthTokenContextDelegate.
 */
@interface MSAuthTokenContextDelegateWrapper : NSObject <MSAuthTokenContextDelegate>
```

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
jwallra: Just FYI, this is a behavior difference from Android.  Does anyone have feedback on why this could be problematic?  Or is it OK?

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,55 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
...
+
+- (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
+  if (authTokenDelegate != nil) {
+    self.authTokenContextDelegateWrapper = [[MSAuthTokenContextDelegateWrapper alloc]
+         initWithAuthTokenDelegate:authTokenDelegate
+        authTokenCompletionHandler:^(NSString *jwt) {
+          MSJwtClaims *claims = [MSJwtClaims parse:jwt];
+          if (claims != nil) {
+            MSLogDebug(MSAppCenter.logTag, @"Token has been refreshed.");
```
jwallra:
```suggestion
            MSLogDebug(MSAppCenter.logTag, @"The auth token has been refreshed.");
```

AppCenter/AppCenter/MSAppCenter.m
jwallra: The delegates are already run in a `@synchronized` block in `MSAuthTokenContext`, is there some other way that using this code could cause deadlocks?

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,227 @@
...
+
+  // If
+  // Valid Token
+  NSString *userId = @"original_user_id";
+  NSString *currentJwt = [MSJwtHelper createJwtWithUserId:userId expiration:86400];
+  MSAuthTokenValidityInfo *validityInfo =
+      [[MSAuthTokenValidityInfo alloc] initWithAuthToken:currentJwt
+                                               startTime:nil
+                                                 endTime:[[NSDate alloc] initWithTimeIntervalSinceNow:86400]];
+
```
jwallra:
```suggestion
```

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,227 @@
...
+  // If
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Completion handler invoked."];
+
+  // Expired Token
+  NSString *userId = @"original_user_id";
+  NSString *currentJwt = [MSJwtHelper createJwtWithUserId:userId expiration:0];
+  MSAuthTokenValidityInfo *validityInfo =
+      [[MSAuthTokenValidityInfo alloc] initWithAuthToken:currentJwt startTime:nil endTime:[[NSDate alloc] initWithTimeIntervalSince1970:0]];
+
```
jwallra:
```suggestion
```

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,227 @@
...
+  [MSAppCenter setAuthToken:currentJwt];
+
+  // Set up invalid JWT (missing exp claim)
```
jwallra: Remove comment and line above, since the method name says what it does now :)

AppCenter/AppCenterTests/MSJwtClaimsTests.m
```diff
@@ -0,0 +1,111 @@
...
+
+@implementation MSJwtClaimsTests
+
+static NSString *const kMSJwtFormat = @"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.%@";
+
+- (void)testGetValidJwt {
+
+  // If
+  NSString *userId = @"some_user_id";
+  int expiration = 1426420800;
```
dhei: Let's not use a magic number. Maybe leave a comment or rename the variable.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
jwallra: Interval of 0 since the epoch (1970)

AppCenter/AppCenterTests/MSJwtClaimsTests.m
jwallra: It doesn't actually need to be a specific value.  Would you prefer to just use 0 for everything?

AppCenter/AppCenter/Internals/Util/MSJwtClaims.h
```diff
@@ -0,0 +1,51 @@
...
+ * @return MSJwtCLaims instance if the string is a valid JWT.
+ */
++ (MSJwtClaims *)parse:(NSString *)jwt;
+
+/**
+ * Returns the subject.
+ *
+ * @return NSString The subject.
+ */
+- (NSString *)getSubject;
```
dhei: Objective-C naming convention doesn't use `get` prefix in the getter method.

Because `subject` is already a readonly property, you don't need this getter method. You can call the `subject` property directly.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,59 @@
...
+    int expirationTimeIntervalSince1970 = [[claims objectForKey:kMSExpirationClaim] intValue];
+    return [[MSJwtClaims alloc] initWithSubject:[claims objectForKey:kMSSubjectClaim]
+                                 expirationDate:[[NSDate alloc] initWithTimeIntervalSince1970:expirationTimeIntervalSince1970]];
+  } @catch (NSException *e) {
+    MSLogError([MSAppCenter logTag], @"Failed to parse JWT: %@", e);
+    return nil;
+  }
+}
+
+- (NSString *)getSubject {
```
dhei: As mentioned [here](https://github.com/microsoft/appcenter-sdk-apple/pull/1753/files#r323505930), I don't think this getter method is necessary.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSJwtClaims.h"
+#import "MSAppCenterInternal.h"
+#import "MSLogger.h"
+#import <Foundation/Foundation.h>
```
dhei: Sort imports.

AppCenter/AppCenterTests/MSJwtClaimsTests.m
dhei: 0 works for me.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
michaelsmitha: @achocron Android is able to actually throw in the try/catch 
https://github.com/microsoft/appcenter-sdk-android/pull/1240/files#diff-c86e6f6d9a168af1ce6ac45f63d50114R57. You can see in our test [here](https://github.com/microsoft/appcenter-sdk-apple/pull/1753/files#diff-23f775a2e0f4731debe8611846ecb5bfR84) that we return a claim with an old interval, android is able to return empty claims [here](https://github.com/microsoft/appcenter-sdk-android/pull/1240/files#diff-43f6e0c4ac3bab52fe59f69bcf67988bR67)

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,224 @@
...
+  [super setUp];
+
+  // Auth token context.
+  [MSAuthTokenContext resetSharedInstance];
+  self.authTokenContextMock = OCMPartialMock([MSAuthTokenContext new]);
+  OCMStub([self.authTokenContextMock sharedInstance]).andReturn(self.authTokenContextMock);
+
+  // System Under Test.
+  [MSAppCenter resetSharedInstance];
+  self.sut = [[MSAppCenter alloc] init];
```
jaelim-ms: Is this used?

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,224 @@
...
+  MSAuthTokenValidityInfo *validityInfo =
+      [[MSAuthTokenValidityInfo alloc] initWithAuthToken:currentJwt
+                                               startTime:nil
+                                                 endTime:[[NSDate alloc] initWithTimeIntervalSinceNow:86400]];
+  [MSAppCenter setAuthToken:currentJwt];
+
+  // Stub delegate mock.
+  id delegateMock = OCMProtocolMock(@protocol(MSAuthTokenDelegate));
+  __block int numCalls = 0;
+  OCMStub([delegateMock appCenter:OCMOCK_ANY acquireAuthTokenWithCompletionHandler:OCMOCK_ANY]).andDo(^(__unused NSInvocation *invocation) {
```
jaelim-ms: Use `OCMReject` for verification if you are verifying the method not to be called.

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,224 @@
...
+    numCalls++;
+  });
+
+  // When
+  [MSAppCenter setAuthTokenDelegate:delegateMock];
+  [self.authTokenContextMock checkIfTokenNeedsToBeRefreshed:validityInfo];
```
jaelim-ms: Don't you need to wait by setting up expectation?

AppCenter/AppCenter/Internals/MSAppCenterInternal.h
```diff
@@ -52,6 +53,8 @@ static NSString *const kMSTrueEnvironmentString = @"1";
 
 @property(nonatomic) NSNumber *requestedMaxStorageSizeInBytes;
 
+@property(nonatomic) MSAuthTokenContextDelegateWrapper *authTokenContextDelegateWrapper;
```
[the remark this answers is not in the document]
jwallra: The comment would essentially be the same as the description for the wrapper itself, no?  it might be useful but it could be bloat

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,51 @@
...
+  if (self) {
+    _subject = subject;
+    _expirationDate = expirationDate;
+  }
+  return self;
+}
+
++ (MSJwtClaims *)parse:(NSString *)jwt {
+  NSArray *parts = [jwt componentsSeparatedByString:kMSJwtPartsSeparator];
+  if ([parts count] < 2) {
```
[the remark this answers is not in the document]
jwallra: Hmm, looks like you're right.  I thought that JWTs didn't always require all three sections but it looks like they do: https://tools.ietf.org/id/draft-jones-json-web-token-01.html (Search for "When validating a JWT the following steps MUST be taken")

@guperrot we have the same bug in Android: https://github.com/microsoft/appcenter-sdk-android/pull/1240/files#diff-c86e6f6d9a168af1ce6ac45f63d50114R48

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,55 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
...
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
+  MSJwtClaims *claims = nil;
+  if (authToken != nil) {
+    claims = [MSJwtClaims parse:authToken];
+  }
+  if (claims != nil) {
+    MSLogDebug(MSAppCenter.logTag, @"Authentication token has been refreshed.");
+    [authTokenContext setAuthToken:authToken withAccountId:claims.subject expiresOn:claims.expirationDate];
+  } else {
+    MSLogDebug(MSAppCenter.logTag, @"Removing authentication token (sign out).");
```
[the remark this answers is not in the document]
jwallra: yes, the message is the same in android: https://github.com/microsoft/appcenter-sdk-android/pull/1240/files#diff-918508e799dc541fd29d897160e9b0a5R1198 (@guperrot FYI)

I'll revise

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,51 @@
...
+    MSLogError([MSAppCenter logTag], @"Failed to parse JWT, not enough parts.");
+    return nil;
+  }
+  NSString *base64ClaimsPart = parts[1];
+  @try {
+    NSError *error;
+    NSData *claimsPartData = [[NSData alloc] initWithBase64EncodedString:base64ClaimsPart options:0];
+    NSDictionary *claims = [NSJSONSerialization JSONObjectWithData:claimsPartData options:0 error:&error];
+    if ([claims objectForKey:kMSExpirationClaim] == nil || [claims objectForKey:kMSSubjectClaim] == nil) {
```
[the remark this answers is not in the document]
jwallra: The logic is app center specific, but I prefer to keep all the validation in the same place.  It gets a lot messier and less maintainable when we have to have parsing in multiple places

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContextDelegateWrapper.h
```diff
@@ -0,0 +1,28 @@
...
+ * so we need to create a wrapper which conforms to MSAuthTokenContextDelegate.
+ */
+@interface MSAuthTokenContextDelegateWrapper : NSObject <MSAuthTokenContextDelegate>
+
+@property id<MSAuthTokenDelegate> delegate;
+
+@property MSAuthTokenCompletionHandler authTokenCompletionHandler;
+
+/**
+ * Create a delegate wrapper which will call the delegates for both MSAuthTokenContextDelegate and MSAuthTokenDelegate.
```
achocron: I wouldn't block on this, but mind tense when writing documentation - in other methods there are sentences like "initializes" rather than "initialize." So maybe this could be "Creates"

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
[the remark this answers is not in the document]
jwallra: Hmm, actually looking at this again, it seems that we only check for 2 parts because we don't care about what's in the last part because we don't do anything with it.  Is this OK?  @jaelim-ms

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,51 @@
...
+}
+
++ (MSJwtClaims *)parse:(NSString *)jwt {
+  NSArray *parts = [jwt componentsSeparatedByString:kMSJwtPartsSeparator];
+  if ([parts count] < 2) {
+    MSLogError([MSAppCenter logTag], @"Failed to parse JWT, not enough parts.");
+    return nil;
+  }
+  NSString *base64ClaimsPart = parts[1];
+  @try {
```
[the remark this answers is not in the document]
jwallra: Exceptions could be thrown in the `NSJSONSerialization` call.  We discussed using the try-catch pattern here, and it seems like it should be OK

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,55 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
...
+  [[MSAppCenter sharedInstance] setAuthTokenDelegate:authTokenDelegate];
+}
+
+- (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
+  if (authTokenDelegate != nil) {
+    self.authTokenContextDelegateWrapper = [[MSAuthTokenContextDelegateWrapper alloc]
+         initWithAuthTokenDelegate:authTokenDelegate
+        authTokenCompletionHandler:^(NSString *jwt) {
+          MSJwtClaims *claims = [MSJwtClaims parse:jwt];
```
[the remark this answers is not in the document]
jwallra: Fixed no null check in `MSJwtClaims`.  I don't see where the code is duplicated though

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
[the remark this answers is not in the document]
guperrot: It's not a full fledge jwt lib. We don't check headers and signature blocks client side, this will be done by ingestion. In SDK we just do the minimum to get sub and exp for auth token context.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,57 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
...
+}
+
+- (void)setAuthTokenDelegate:(id<MSAuthTokenDelegate>)authTokenDelegate {
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
+  @synchronized(self.authTokenContextDelegateWrapper) {
+    if (authTokenDelegate != nil) {
+      self.authTokenContextDelegateWrapper = [[MSAuthTokenContextDelegateWrapper alloc]
+           initWithAuthTokenDelegate:authTokenDelegate
+          authTokenCompletionHandler:^(NSString *jwt) {
+            MSJwtClaims *claims = [MSJwtClaims parse:jwt];
```
guperrot: The completion block should use the instance auth set token method, supposed to be same logic. At least on Android that's what we did.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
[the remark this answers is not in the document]
jaelim-ms: If Ingestion requires signature block, we can check it here and fail it earlier. It would things make harder to investigate any issues caused by this block. I would rather displaying error message in device console than looking up error messages in Ingestion if signature block is missing.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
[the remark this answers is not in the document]
jaelim-ms: You should be able to read `error` instead of catching it.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
```diff
@@ -0,0 +1,56 @@
...
+- (instancetype)initWithSubject:(NSString *)subject expiration:(NSDate *)expiration {
+  self = [super init];
+  if (self) {
+    _subject = subject;
+    _expiration = expiration;
+  }
+  return self;
+}
+
++ (MSJwtClaims *)parse:(NSString *_Nullable)jwt {
```
jaelim-ms: Why is this `nullable`? It will always return `nil` with error message.

AppCenter/AppCenterTests/MSByoiTests.m
```diff
@@ -0,0 +1,218 @@
...
+
+  // Stub delegate mock.
+  id delegateMock = OCMProtocolMock(@protocol(MSAuthTokenDelegate));
+  OCMReject([delegateMock appCenter:OCMOCK_ANY acquireAuthTokenWithCompletionHandler:OCMOCK_ANY]);
+
+  // When
+  [MSAppCenter setAuthTokenDelegate:delegateMock];
+  [self.authTokenContextMock checkIfTokenNeedsToBeRefreshed:validityInfo];
+  
+  // Then - nothing, we just need to make sure the method wasn't called
```
jaelim-ms: You must call `OCMVerifyAll` to confirm `OCMReject`.

AppCenter/AppCenterTests/MSByoiTests.m
jaelim-ms: This one hasn't been addressed or answered.

AppCenter/AppCenter/Internals/Util/MSJwtClaims.m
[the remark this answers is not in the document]
jaelim-ms: Never mind. We need the log even though JWT isn't valid.

AppCenter/AppCenterTests/MSByoiTests.m
jaelim-ms: The call doesn't have any async calls so it seems okay.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -174,6 +178,51 @@ + (void)setCustomProperties:(MSCustomProperties *)customProperties {
...
 
++ (void)setAuthToken:(NSString *)authToken {
+  [[MSAppCenter sharedInstance] setAuthToken:authToken];
+}
+
+- (void)setAuthToken:(NSString *)authToken {
+  MSAuthTokenContext *authTokenContext = [MSAuthTokenContext sharedInstance];
+  MSJwtClaims *claims = nil;
+  if (authToken != nil) {
+    claims = [MSJwtClaims parse:authToken];
```
jaelim-ms: Error message in `MSJwtClaims` is a dead code. Please remove `nil` check here or move the log message from `MSJwtClaims` to here.

PR #1764, opened 2019-09-17
Add aad authentication to iOS sdk

manuchandel:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

This PR is for supporting AAD Enabled Authentication for iOS SDK.

### Prerequisites

1. MSAL requires **authority_url** for authentication.
2. For B2C **authority_url** is generated by Azure when a user registers his app on Azure Portal.
3. For AAD **authority_url** is different and is not returned by Azure as is.

The JSON Payload for AAD looks something like this

```
{
   "client_id" : "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX",
   "authorization_user_agent" : "DEFAULT",
   "authorities" : [
      {
         "type": "AAD",
         "audience": {
            "type": "AzureADandPersonalMicrosoftAccount",
            "tenant_id": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX"
         }
      }
   ]
}
```

To calculate authority_url we use following logic

1. If **type** in **audience** is **AzureADMyOrg** then uri is *https://login.microsoftonline.com/{tenantId}*
1. If **type** in **audience** is **AzureADMultipleOrgs** then uri is *https://login.microsoftonline.com/organizations*
1. If **type** in **audience** is **AzureADandPersonalMicrosoftAccount** then uri is *https://login.microsoftonline.com/common*

### MSAuthority
1. **MSAuthority** is base class of all Authorities and contains a static method **AuthorityWithDictionary** which creates instance of Authority.
1. It processes each authority and based on type initiates objects.
1. Common code processing of reading type and default is done in this class.

### MSB2CAuthority
1. **MSB2CAuthority** implements **MSAuthority** and reads **redirect_uri** as before.
1. IsValid method now has one additional check from before. it checks for the type B2C.

### MSAADAuthority
1. **MSAADAuthority** implements **MSAuthority** and constructs **authority_url** by the logic explained above.
1. IsValid method now has one additional check from before. it checks for the type AAD.

### MSAuthConfig
1. Instead of directly initiating object of MSAuthority type as before, it now calls **AuthorityWithDictionary** static method to get proper object according to authentication type.
[...]

PR #1781, opened 2019-09-20
change B2C authentication to MSAuthentication and change error message

manuchandel:

* [ ] Did you add unit tests?

## Description

In MSAuth.m we were initializing MSB2CAuthority through this line of code

```
MSALB2CAuthority *auth = [[MSALB2CAuthority alloc] initWithURL:(NSURL * __nonnull) self.authConfig.authorities[0].authorityUrl error:nil]
```

Since we are also initializing AAd Authority now depending on the app I changed it to another method from MSAL which automatically decides which authority object to create based on authority url

```
MSALAuthority *auth = [MSALAuthority authorityWithURL:(NSURL * __nonnull)self.authConfig.authorities[0].authorityUrl error:nil];
```

I also changed error message in next line. Earlier it read

```
Auth config doesn't contain a valid default B2C authority.
```

Now it will give message specific to Authority

```
Auth config doesn't contain a valid default %@ authority.", self.authConfig.authorities[0].type)
```

## Related PRs or issues

#1764 

## Misc

### DIFF
--- a/AppCenterAuth/AppCenterAuth/MSAuth.m
+++ b/AppCenterAuth/AppCenterAuth/MSAuth.m
@@ -259,9 +259,8 @@ static dispatch_once_t onceToken;
                 withErrorCode:(NSInteger)errorCode
                       message:(NSString *)errorMessage {
   if (completionHandler) {
-    NSError *error = [[NSError alloc] initWithDomain:kMSACAuthErrorDomain
-                                                code:errorCode
-                                            userInfo:@{NSLocalizedDescriptionKey : errorMessage}];
+    NSError *error =
+        [[NSError alloc] initWithDomain:kMSACAuthErrorDomain code:errorCode userInfo:@{NSLocalizedDescriptionKey : errorMessage}];
     completionHandler(nil, error);
   }
 }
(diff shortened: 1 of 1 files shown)


AppCenterAuth/AppCenterAuth/MSAuth.m
```diff
@@ -394,13 +391,13 @@ - (void)configAuthenticationClient {
...
-  MSALB2CAuthority *auth = [[MSALB2CAuthority alloc] initWithURL:(NSURL * __nonnull) self.authConfig.authorities[0].authorityUrl error:nil];
+  MSALAuthority *auth = [MSALAuthority authorityWithURL:(NSURL * __nonnull)self.authConfig.authorities[0].authorityUrl error:nil];
   MSALPublicClientApplicationConfig *config =
-      [[MSALPublicClientApplicationConfig alloc] initWithClientId:(NSString * __nonnull) self.authConfig.clientId
+      [[MSALPublicClientApplicationConfig alloc] initWithClientId:(NSString * __nonnull)self.authConfig.clientId
                                                       redirectUri:self.authConfig.redirectUri
                                                         authority:auth];
   if (!auth) {
-    MSLogError([MSAuth logTag], @"Auth config doesn't contain a valid default B2C authority.");
+    MSLogError([MSAuth logTag], @"Auth config doesn't contain a valid default %@ authority.", self.authConfig.authorities[0].type);
```
jaelim-ms: As far as I know some authorities might not contain type. Are you sure the first authority always contains type?
Isn't there a chance that `self.authConfig.authorities` could be `nil` or empty?

manuchandel: In MSAuthConfig we make sure that the default authority is processed and it contains a valid type either B2C or AAD or we throw error. This preprocessing steps make sure we always have a valid type.

manuchandel: First authority here will always be an authority whose default is true and has type = AAD or B2C. MSAuthConfig makes sure of it. If you look at a couple of lines above this code, we send 1st authority to MSAL for authentication

PR #1782, opened 2019-09-23
[Feature] iOS and macOS extensions

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Add crash support for:
- iOS extensions
- macOS extensions

## Related PRs or issues

[AB#64886](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64886)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # App Center SDK for iOS, macOS and tvOS Change Log
 
+## Version 2.5.0 (Under development)
+
+### App Center Crashes
+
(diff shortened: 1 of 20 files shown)


clpolet:
We need to remove the mac app extension from the mac app because AC build doesn't support building macOS app extensions (can't upload the prov profile). We'll have to build it separately unless I am missing something.

MatkovIvan:
I've created #1787 PR with iOS only to avoid revert of revert of revert commits

PR #1793, opened 2019-09-25
Mbaas/storage network offline

MehrdadMzfr:

* [ ] Did you add unit tests?

## Description

Use separate HTTP clients for online and offline operations. This is to fail fast on online operations and still keep retrying on offline ones.


nikitag-ms:
Guillaume suggested updating the changelog in the Android version. You can just copy it from there: https://github.com/microsoft/appcenter-sdk-android/pull/1258/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e.

guperrot:
Any SDK behavior change or signature change has to be documented in the changelog.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ ___
 ### App Center Data
 
 * **[Breaking change]** Rename delegate method `data:didCompletePendingOperation:forDocument:withError:` from `MSRemoteOperationDelegate` to `data:didCompleteRemoteOperation:forDocumentMetadata:withError:`.
+* **[Fix]** Reduced retries on Data-related operations to fail fast and avoid the perception of calls "hanging".
```
```suggestion
* **[Fix]** Reduce retries on Data-related operations to fail fast and avoid the perception of calls "hanging".
```

AppCenterData/AppCenterData/MSData.m
```diff
@@ -661,6 +664,7 @@ - (void)deleteFromCosmosDbWithPartition:(NSString *)partition
   [self performCosmosDbOperationWithPartition:partition
                                    documentId:documentId
                                    httpMethod:kMSHttpMethodDelete
+                                   httpClient:self.httpClientWithRetrier
```
nikitag-ms: This one isn't processing a pending delete, right? Shouldn't this one be without a retryer?

AppCenterData/AppCenterData/MSData.m
```diff
@@ -556,6 +557,7 @@ - (void)upsertFromCosmosDbWithPartition:(NSString *)partition
       performCosmosDbOperationWithPartition:partition
                                  documentId:documentId
                                  httpMethod:kMSHttpMethodPost
+                                 httpClient:self.httpClientWithRetrier
```
nikitag-ms: I might be reading this incorrectly, but this seems like an online operation (not a pending one). If so, then shouldn't this be the no retrier instance?

AppCenterData/AppCenterData/MSData.m
MehrdadMzfr: this is actually coming from `processPendingOperations` call. for regular delete it uses:
```- (void)deleteDocumentWithID:(NSString *)documentID
                   partition:(NSString *)partition
                writeOptions:(MSWriteOptions *_Nullable)writeOptions
           completionHandler:(MSDocumentWrapperCompletionHandler)completionHandler {```

File: AppCenterData/AppCenterData/MSData.m (hunk 3 above)
Comment (MehrdadMzfr) [reply AppCenterData/AppCenterData/MSData.m]: good point, refactored the code, this method can be called from both CreateOrReplace and ProcessPendingOperations. therefore, it's required to path the desired httpClient as a parameter.

File: CHANGELOG.md [hunk 4]
```diff
@@ -24,6 +24,7 @@ ___
 ### App Center Data
 
 * **[Breaking change]** Rename delegate method `data:didCompletePendingOperation:forDocument:withError:` from `MSRemoteOperationDelegate` to `data:didCompleteRemoteOperation:forDocumentMetadata:withError:`.
+* **[Fix]** Reduce retries on Data-related operations to fail fast and avoid the perception of calls "hanging".
```
jaelim-ms: We've already released 2.4.0. Please move this to a newer version.

PR #1796, opened 2019-09-26
Fix/token refresh check

achocron:

* [x] Did you add unit tests?

## Description

This fixes an issue where the token refresh can be invoked multiple times for a single token.
[AB#69442](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/69442)


achocron:
Don't merge, realized there is a bug.

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -485,6 +485,31 @@ - (void)testRefreshAuthTokenForAccountId {
...
+  [self.sut setAuthToken:expectedAuthToken withAccountId:expectedAccountId expiresOn:expiresDate];
+  MSAuthTokenValidityInfo *authToken = [[MSAuthTokenValidityInfo alloc] initWithAuthToken:expectedAuthToken
+                                                                                startTime:startDate
+                                                                                  endTime:expiresDate];
+
+  // When
+  [self.sut checkIfTokenNeedsToBeRefreshed:authToken];
+
+  // Then
+  OCMVerify([delegateMock authTokenContext:OCMOCK_ANY refreshAuthTokenForAccountId:expectedAccountId]);
```
jaelim-ms: These are dead code. It doesn't verify/reject anything in the test.

AppCenterAuth/AppCenterAuthTests/MSAuthTests.m
```diff
@@ -1383,7 +1383,7 @@ - (void)testRefreshWithExpiredTokenWhileMsalNotConfigured {
   [authTokenContextMock stopMocking];
 }
 
-- (void)testRefreshRetriedAfterMsalConfigured {
+- (void)testSignOutIfRefreshWithNoMsalConfiguration {
```
jaelim-ms: Why are you replacing the unit test?

jwallra: Yeah this seems kind of confusing, is this not changing the intent of the test case?  Do we not need both?

AppCenter/AppCenterTests/MSAuthTokenContextTests.m
```diff
@@ -485,6 +485,37 @@ - (void)testRefreshAuthTokenForAccountId {
   OCMVerify([delegateMock authTokenContext:OCMOCK_ANY refreshAuthTokenForAccountId:expectedAccountId]);
 }
 
+- (void)testSameAuthTokenNotRefreshedTwice {
+
+  // If
+  NSString *expectedAuthToken = @"authToken1";
+  NSString *expectedAccountId = @"account1";
+  NSDate *startDate = [NSDate dateWithTimeIntervalSinceNow:-((60.0f * 60.0f * 24.0f) * 2)];
+  NSDate *expiresDate = [NSDate dateWithTimeIntervalSinceNow:+500];
```
dhei:
```suggestion
  NSDate *expiresDate = [NSDate dateWithTimeIntervalSinceNow:500];
```
I don't think we need the `+` sign. 😄

AppCenterAuth/AppCenterAuthTests/MSAuthTests.m
achocron: @jaelim-ms That was the case we discussed where we should actually sign out in that case because in the valid flow that couldn't happen.

Issue #1734, opened 2019-08-19
Duplicate interface definition for class 'MSAbstractLog.h' and 'MSServiceAbstract.h'


madhurima09:
Hello,

I'm trying to integrate AppCenter SDK into my project. I have added both the frameworks i.e 'App Center' and 'App Center Crashes'. Both these frameworks have 2 exactly same header files i.e 'MSAbstractLog.h' and 'MSServiceAbstract.h' , as a result of which I'm getting the below compile time errors

- AppCenter/AppCenter/MSAbstractLog.h:6:1: error: duplicate interface definition for class 'MSAbstractLog'

- AppCenter/AppCenter/MSServiceAbstract.h:14:1: error: duplicate interface definition for class 'MSServiceAbstract'

- AppCenter/AppCenter/MSServiceAbstract.h:19:35: error: property has a previous declaration

I have tried adding the frameworks once using binary integration and once via cocoapods. Both ways I got the same issue.
Also, as per the app center sdk documentation, I need to include both the frameworks App Center and App Center Crashes. 

Any one else faced same issue? 
Also, could anyone help to figure out how to resolve this problem?

Thanks.

AnastasiaKubova:
Hi, madhurima09 ! Thank you for contacting us! I didn't manage to reproduce your issue. Could you please clarify some details:
- What the version SDK are you use?
- Could you please share the code snippet with use our SDK?

madhurima09:
Hi @AnastasiaKubova 
Thanks for taking a look at this.
I'm on the latest version of app center sdk. The below implementation is in AppDelegate class.

#import <AppCenter/AppCenter.h>
#import <AppCenterCrashes/AppCenterCrashes.h>

+ (void)initAppCenterCrashService
{
    [MSCrashes setDelegate: self];
    [MSAppCenter start:@"key" withServices:@[[MSCrashes class]]];
}

+ (void)appCenterCrashReportingService
{
static dispatch_once_t onceToken1 = 0;
        dispatch_once(&onceToken1, ^{
[self initAppCenterCrashService];
}
}

To reproduce the issue
1) Once app center is installed via cocoapods or by copying the binaries. 
2)And on building the project there after, the above reported build issues start to show.

Attached is the structure of how the framework looks once installed via cocoapods.

You will notice in the attached image that both AppCenter framework and AppCenterCrashes framework has these 2 same header files'MSAbstractLog.h' and 'MSServiceAbstract.h'.
[...]

PR #1822, opened 2019-10-08
Fix compatibility with projects that don't use clang modules

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Fix compatibility with projects that don't use clang modules.

## Related PRs or issues

#1734
[AB#66994](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/66994)


Sasquatch/SasquatchObjC/AppCenterDelegateObjC.h
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#import <UIKit/UIKit.h>
```
jaelim-ms: Why do you need these?

MatkovIvan: to compile without modules enabled. Swift bindings contains some symbols from this

AppCenter/AppCenter/MSChannelProtocol.h
```diff
@@ -1,9 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#ifndef MS_CHANNEL_PROTOCOL_H
-#define MS_CHANNEL_PROTOCOL_H
```
clpolet: What was this used for? I feel like I am discovering this now. It looks like a hack. I wonder if that was at the time of the event filter service.

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -481,7 +481,7 @@
 		F8936CB5230C24D9006A330F /* MSChannelGroupProtocol.h in Headers */ = {isa = PBXBuildFile; fileRef = 3542741A2012AF0500BE766F /* MSChannelGroupProtocol.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		F8936CB6230C24D9006A330F /* MSChannelProtocol.h in Headers */ = {isa = PBXBuildFile; fileRef = 354274162012AE4000BE766F /* MSChannelProtocol.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		F8936CB7230C24D9006A330F /* MSConstants.h in Headers */ = {isa = PBXBuildFile; fileRef = 6E04013F1D1C99AC0051BCFA /* MSConstants.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		F8936CB8230C24D9006A330F /* MSConstants+Flags.h in Headers */ = {isa = PBXBuildFile; fileRef = 04B525B82194D44E00FA37FD /* MSConstants+Flags.h */; settings = {ATTRIBUTES = (Private, ); }; };
+		F8936CB8230C24D9006A330F /* MSConstants+Flags.h in Headers */ = {isa = PBXBuildFile; fileRef = 04B525B82194D44E00FA37FD /* MSConstants+Flags.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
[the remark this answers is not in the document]
MatkovIvan: I guess it already is. I'll re-check

AppCenter/AppCenter/MSChannelProtocol.h
MatkovIvan: it was for carthage build to deal with core dependency conflicts in modules. It's not a root cause and the wrong way to solve the problem. It isn't required now because I've cleaned up the dependencies setup.
Even when/if we move to proper framework references during modules build, it shouldn't be there

MatkovIvan:
Closing this as it's required more changes

Issue #1817, opened 2019-10-04
Unexpected Xcode Warnings 


phoney:
### **Description**

I'm currently seeing a number of warnings when the appcenter cocoapod is included in my swift iOS Xcode project. I believe that these warnings started recently but I'm not sure the exact version. 

I'm using Appcenter 2.4.0.

Similar warnings were mentioned in issue #1789 but I don't think I'm being prevented from archiving my apps.

These appear in the GenerateDSYMFile step.
warning: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/C/org.llvm.clang.vsts/ModuleCache/1B76CAT92BTRK/Foundation-OIS75AKBYQKF.pcm: No such file or directory
note: while processing /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/C/org.llvm.clang.vsts/ModuleCache/1B76CAT92BTRK/Foundation-OIS75AKBYQKF.pcm
note: Linking a static library that was built with -gmodules, but the module cache was not found.  Redistributable static libraries should never be built with module debugging enabled.  The debug experience will be degraded due to incomplete debug information.

There are about 15 more "No such file or directory" warnings that are very similar.

### **Repro Steps**

My podfile simply says

    pod 'AppCenter'

among the other pods I use. 

### **Details**

1. Which SDK version are you using?
    - 2.4.0
2. Which OS version did you experience the issue on?
    - iOS 11, macOS 10.14.6
3. Which Cocoapods version are you using (run `pod --version`)?
    - 1.8.1 but I think it was in 1.7.x aso
4. What device version did you see this error on?  Were you using an emulator or a physical device?
    na
5. What language are you using?
    - Swift

This is a compile time issue. I'm using Xcode 11.0.

I'm guessing those files that can't be found are in the ModuleCache on your build server and are somehow referenced from the pre-built library.

Jamminroot:
Hey, @phoney! Thanks for bringing this to our attention.
Please give us some time to investigate the issue you are facing.

alexcosmin:
I'm also seeing this.

SDK version: 2.4.0
OS version where issue is experienced: iOS 13.1.2
Cocoapods version: 1.8.3
Language used: Swift
Seeing this on both the iOS simulator and a physical device (iPhone XR)
Compiling with Xcode 11.0 on macOS 10.14.6

[...]

PR #1836, opened 2019-10-14
Fix warnings in Xcode 11 when SDK is installed via CocoaPods

olmisa:

* [ ] Did you add unit tests?

#1817
[AB#70133](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/70133)


Sasquatch/Config/Sasquatch ObjC.xcconfig
```diff
@@ -2,3 +2,4 @@ GCC_PREPROCESSOR_DEFINITIONS = $(inherited) GCC_PREPROCESSOR_MACRO_SASQUATCH_OBJ
 
 // For verifying projects with disabled modules.
 CLANG_ENABLE_MODULES = NO
+OTHER_LDFLAGS = $(OTHER_LDFLAGS) -framework UserNotifications
```
jaelim-ms: Why do we need this one in the app?

olmisa: To avoid the following error that appears during 'SusquatchObjC' build:

   ```
Undefined symbols for architecture x86_64:
      "_OBJC_CLASS_$_UNUserNotificationCenter", referenced from:
         objc-class-ref in AppDelegate.o
         objc-class-ref in AppCenterPush(MSPush.o)
         objc-class-ref in AppCenterPush(MSUserNotificationCenterDelegateForwarder.o)
```

File: Config/Tests.xcconfig [hunk 2]
```diff
@@ -10,4 +10,9 @@ WARNING_CFLAGS = -Wno-objc-messaging-id
 
 OTHER_CFLAGS = $(inherited) -iframework "$(SRCROOT)/../Vendor/$(APPCENTER_BUILD_PLATFORM)/OCMock" -iframework "$(SRCROOT)/../Vendor/$(APPCENTER_BUILD_PLATFORM)/OCHamcrest" -iframework "$(SRCROOT)/../Vendor/$(APPCENTER_BUILD_PLATFORM)/OHHTTPStubs" -iframework "$(PLATFORM_DIR)/Developer/Library/Frameworks"
 
+OTHER_LDFLAGS = $(inherited) -ObjC -lz
```
clpolet: Why do you use `$(inherited)` here instead of `$(OTHER_LDFLAGS)`like in other places?

Sasquatch/Config/Sasquatch ObjC.xcconfig
clpolet: Is the SDK properly declaring the dependency to this system framework? I would think the SDK would infer this dependence. Why don't we have to do that on Sasquatch Swift?

clpolet: That makes sense

MatkovIvan: Wait. Do we really need to add `CLANG_ENABLE_MODULE_DEBUGGING` if we set `CLANG_ENABLE_MODULES = NO` ?

olmisa: Yes, you are right. This property should be removed.

PR #1850, opened 2019-10-24
core delegate handling for the data change notifications

nalankr:
* [ No ] Has `CHANGELOG.md` been updated?
* [ Yes ] Are tests passing locally?
* [ Yes ] Are the files formatted correctly?
* [ Yes ] Did you add unit tests?
* [ No ] Did you check UI tests on the sample app? They are not executed on CI.
* [ No ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

This PR adds the core delegate handling for the data change notifications. 
https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/70610

The overall feature design
https://msmobilecenter.visualstudio.com/Mobile-Center/_wiki/wikis/Mobile-Center.wiki/3337/Design-iOS-SDK-for-Data-Change-Notifications


AppCenter/AppCenterTests/MSServiceNotificationDelegateTests.m
```diff
@@ -0,0 +1,119 @@
...
+#import "MSTestFrameworks.h"
+
+@interface MSServiceNotificationDelegateTests : XCTestCase
+
+@end
+
+@implementation MSServiceNotificationDelegateTests
+
+- (void)testNoRegisteredDelegate {
+  id appCenterMock = OCMPartialMock([MSAppCenter sharedInstance]);
```
clpolet: You'll need to set/reset the AppCenter instance inside the `setup` `teardown` method to run each tests with a clean instance.

AppCenter/AppCenterTests/MSServiceNotificationDelegateTests.m
```diff
@@ -0,0 +1,119 @@
...
+  // creating delegates
+  id delegate = OCMProtocolMock(@protocol(MSServiceNotificationDelegate));
```
clpolet: If you don't set the delegate then I don't think you need this delegate at all in this test. Can delete this line and the `OCMReject` line 27. Just by not crashing we confirm that not having delegates registered works.

AppCenter/AppCenterTests/MSServiceNotificationDelegateTests.m
```diff
@@ -0,0 +1,119 @@
...
+#import "MSAppCenter.h"
+#import "MSAppCenterInternal.h"
+#import "MSServiceNotificationDelegate.h"
+#import "MSTestFrameworks.h"
+
+@interface MSServiceNotificationDelegateTests : XCTestCase
+
+@end
+
+@implementation MSServiceNotificationDelegateTests
```
clpolet: You can probably add a test for when a delegate is registered but doesn't implement the optional method.

AppCenter/AppCenterTests/MSServiceNotificationDelegateTests.m
clpolet: Note you can also add these tests to the MSAppCenterTests.m file as everything is setup there.

nalankr: I am using MSAppCenterTests.m now for the new tests that I am adding

AppCenter/AppCenterTests/MSServiceNotificationDelegateTests.m
nalankr: There is no optional method in our serviceNotificationdelegate. There is only one required method. Please let me know if you there are other tests that we can add.

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -975,4 +975,106 @@ - (void)testNoUserIdWhenSetUserIdIsNotCalledInNextVersion {
...
+  [self.sut removeServiceNotificationDelegate:delegate2];
+
+  // Notification data.
+  NSDictionary<NSString *, NSString *> *data = @{@"key" : @"value"};
+
+  // When
+  [self.sut forwardServiceNotification:data];
+
+  // Then
+  OCMReject([delegate1 appCenter:_sut didReceiveServiceNotification:data]);
```
clpolet: The rejects here should go line 1018 after assigning the mocked delegates. Reject actually crashes the test at the time the rejected method is called. This is why it should be set early. Then if the test simply doesn't crash then it succeeds.

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -975,4 +975,106 @@ - (void)testNoUserIdWhenSetUserIdIsNotCalledInNextVersion {
...
+  [self.sut addServiceNotificationDelegate:delegate2];
+
+  // Notification data.
+  NSDictionary<NSString *, NSString *> *data = @{@"key" : @"value"};
+
+  // When
+  [self.sut forwardServiceNotification:data];
+
+  // Then
+  OCMVerify([delegate1 appCenter:_sut didReceiveServiceNotification:data]);
```
clpolet: We usually postfix mocks names (it's not consistent everywhere though). But that would make `delegate1` and `delegate2` renamed to `delegate1Mock` `delegate2Mock`

AppCenter/AppCenterTests/MSAppCenterTests.m
```diff
@@ -975,4 +975,106 @@ - (void)testNoUserIdWhenSetUserIdIsNotCalledInNextVersion {
...
+  OCMReject([delegate2 appCenter:_sut didReceiveServiceNotification:data]);
```
clpolet: Same comment about this reject here.

AppCenter/AppCenter/Internals/MSServiceNotificationDelegate.h
```diff
@@ -0,0 +1,17 @@
...
+@class MSAppCenter;
+
+@protocol MSServiceNotificationDelegate <NSObject>
+
+/**
+ * A callback that is called when an App Center service notification is received.
+ * @param appCenter The app center instance.
+ * @param notificationData The service notification data.
+ */
+- (void)appCenter:(MSAppCenter *)appCenter didReceiveServiceNotification:(NSDictionary<NSString *, NSString *> *)notificationData;
```
clpolet: Does it actually make sense to keep this method optional? If it makes sense we need to add a test for delegates that don't implement this method. If it doesn't make sense and we make it required we don't need to check `respondsToSelector` on delegates before calling this method on them.

nalankr: Made it required and removed respondsToSelector.

AppCenterAuth/AppCenterAuth.xcodeproj/project.pbxproj
```diff
@@ -925,20 +933,27 @@
 			remoteRef = 04E9760C226A50BF007E6DD1 /* PBXContainerItemProxy */;
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
-		04E9760F226A50BF007E6DD1 /* libMSAL.a */ = {
+		04E9760F226A50BF007E6DD1 /* libMSAL (iOS Static Library).a */ = {
```
jwallra: I'm curious about these file name changes.  I don't think they should be present, there shouldn't be any reason to change them

PR #1855, opened 2019-10-29
Implement didReceiveServiceNotification in MSData and add delegate me…

thomasf7:

* [x] Did you add unit tests?

## Description

Implement didReceiveServiceNotification in MSData and add delegate method didReceiveDocumentChangeNotification to check and forward document change notifications to + tests

## Related PRs or issues

## Misc

### DIFF
--- a/AppCenterData/AppCenterData/Internal/MSDataPrivate.h
+++ b/AppCenterData/AppCenterData/Internal/MSDataPrivate.h
@@ -4,6 +4,7 @@
 #import "MSAuthTokenContextDelegate.h"
 #import "MSDataDelegate.h"
 #import "MSServiceInternal.h"
+#import "MSServiceNotificationDelegate.h"
 #import "MS_Reachability.h"
 
 NS_ASSUME_NONNULL_BEGIN
@@ -13,7 +14,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 static NSString *const kMSDefaultApiUrl = @"https://tokens.appcenter.ms/v0.1";
 
-@interface MSData () <MSAuthTokenContextDelegate>
+/**
+ * Key for payload in service notification
+ */
+static NSString *const kMSServiceNotificationDocumentChangeKey = @"documentsChange";
+
+@interface MSData () <MSAuthTokenContextDelegate, MSServiceNotificationDelegate>
 
 /**
  * Method to reset the singleton when running unit tests only. So calling sharedInstance returns a fresh instance.
@@ -90,6 +96,11 @@ static NSString *const kMSDefaultApiUrl = @"https://tokens.appcenter.ms/v0.1";
  * Process pending operations and sync with CosmosDb.
  */
 - (void)processPendingOperations;
+
+/**
+ * MSServiceNotificationDelegate.
+ */
+- (void)appCenter:(MSAppCenter *)appCenter didReceiveServiceNotification:(NSDictionary<NSString *, NSString *> *)notificationData;
 
 @end
 
--- a/AppCenterData/AppCenterData/Internal/MSDataPrivate.h
+++ b/AppCenterData/AppCenterData/Internal/MSDataPrivate.h
@@ -97,11 +97,6 @@ static NSString *const kMSServiceNotificationDocumentChangeKey = @"documentsChan
  */
 - (void)processPendingOperations;
 
-/**
- * MSServiceNotificationDelegate.
- */
-- (void)appCenter:(MSAppCenter *)appCenter didReceiveServiceNotification:(NSDictionary<NSString *, NSString *> *)notificationData;
-
 @end
 
 NS_ASSUME_NONNULL_END
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -838,23 +838,22 @@ static dispatch_once_t onceToken;
                                                                        error:&error];
     if (error) {
       MSLogError([MSData logTag], @"Couldn't parse json data for service notification: %@", error.localizedDescription);
-    } else {
-      NSMutableArray *documentChanges = [NSMutableArray new];
-      for (NSDictionary *dictionary in documentChangePayload) {
-        MSDocumentChange *documentChange = [[MSDocumentChange alloc] initWithDictionary:dictionary];
-        [documentChanges addObject:documentChange];
-      }
-
-      if ([documentChanges count] > 0) {
-        id<MSDataDelegate> strongDelegate;
-        @synchronized(self) {
-          strongDelegate = self.delegate;
-          if (strongDelegate) {
-            if ([strongDelegate respondsToSelector:@selector(data:didReceiveDocumentChangeNotification:)]) {
-              [strongDelegate data:self didReceiveDocumentChangeNotification:documentChanges];
-            }
-          }
-        }
+      return;
+    }
+    NSMutableArray *documentChanges = [NSMutableArray new];
+    for (NSDictionary *dictionary in documentChangePayload) {
+      MSDocumentChange *documentChange = [[MSDocumentChange alloc] initWithDictionary:dictionary];
+      [documentChanges addObject:documentChange];
+    }
+    if ([documentChanges count] == 0) {
+      MSLogDebug([MSData logTag], @"Empty document change notification received.");
+      return;
+    }
+    id<MSDataDelegate> strongDelegate;
+    @synchronized(self) {
+      strongDelegate = self.delegate;
+      if (strongDelegate && [strongDelegate respondsToSelector:@selector(data:didReceiveDocumentChangeNotification:)]) {
+        [strongDelegate data:self didReceiveDocumentChangeNotification:documentChanges];
       }
     }
   }
--- a/AppCenterData/AppCenterData/MSData.m
+++ b/AppCenterData/AppCenterData/MSData.m
@@ -15,6 +15,7 @@
 #import "MSDataInternal.h"
 #import "MSDataPrivate.h"
 #import "MSDictionaryDocument.h"
+#import "MSDocumentChange.h"
 #import "MSDocumentMetadata.h"
 #import "MSDocumentMetadataInternal.h"
 #import "MSDocumentStore.h"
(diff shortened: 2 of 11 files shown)


AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2819,4 +2821,125 @@ - (void)testRemoteOperationDelegateMethodIsCalledWithNoExistenceOfImplementation
...
+  [dataMock appCenter:nil didReceiveServiceNotification:notificationData];
+
+  OCMVerifyAll(dataMock);
+  [dataMock stopMocking];
+}
+
+- (void)testDocumentChangeDelegateMethodIsNotCalledForDelegateThatDoesNotRespondToSelector {
+  id dataMock = OCMPartialMock(self.sut);
+  OCMStub([dataMock sharedInstance]).andReturn(dataMock);
+  id dataDelegateMock = OCMPartialMock([MSMockDataDelegate new]);
```
jaelim-ms: You can use protocol mock then you don't need `MSMockDataDelegate`.

thomasf7: How do I use protocol mock and have it not implement the optional method? I added this because I couldn't find another way to have respondsToSelector return false.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -823,6 +826,40 @@ - (void)authTokenContext:(MSAuthTokenContext *)__unused authTokenContext didUpda
...
+- (void)appCenter:(MSAppCenter *)__unused appCenter didReceiveServiceNotification:(NSDictionary<NSString *, NSString *> *)notificationData {
+  NSString *documentChangePayloadJSON = notificationData[kMSServiceNotificationDocumentChangeKey];
+  if (documentChangePayloadJSON) {
+    NSData *documentChangePayloadData = [documentChangePayloadJSON dataUsingEncoding:NSUTF8StringEncoding];
+    NSError *error = nil;
+    NSArray *documentChangePayload = [NSJSONSerialization JSONObjectWithData:documentChangePayloadData
+                                                                     options:NSJSONReadingMutableContainers
+                                                                       error:&error];
+    if (error) {
+      MSLogError([MSData logTag], @"Couldn't parse json data for service notification: %@", error.localizedDescription);
```
jaelim-ms: Please do not nest `if` statements. You can simply return here and remove at least one `else` statement below.

AppCenterData/AppCenterData/Model/MSDocumentChange.m
```diff
@@ -0,0 +1,26 @@
...
+static NSString *const kMSDocumentChangeId = @"id";
+static NSString *const kMSDocumentChangePartition = @"ptn";
+static NSString *const kMSDocumentChangeOperation = @"op";
+static NSString *const kMSDocumentChangeTimestamp = @"_ts";
+
+- (instancetype)initWithDictionary:(NSDictionary *)dictionary {
+  if (!dictionary) {
+    return nil;
+  }
+  if ((self = [super init]) != nil) {
```
jaelim-ms:
```suggestion
  if ((self = [super init])) {
```

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2819,4 +2821,125 @@ - (void)testRemoteOperationDelegateMethodIsCalledWithNoExistenceOfImplementation
...
+- (void)testDocumentChangeWithNoRegisteredDelegate {
+  id dataMock = OCMPartialMock(self.sut);
+  OCMStub([dataMock sharedInstance]).andReturn(dataMock);
+
+  NSString *const documentChangesJson = @"[{\"op\":\"replace\",\"id\":\"BHGVGD5\",\"ptn\":\"user\",\"_ts\":1000}]";
+  NSDictionary<NSString *, NSString *> *notificationData = @{kMSServiceNotificationDocumentChangeKey : documentChangesJson};
+
+  [dataMock appCenter:nil didReceiveServiceNotification:notificationData];
+
+  OCMVerifyAll(dataMock);
```
jaelim-ms: What are you verifying here?

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2819,4 +2821,125 @@ - (void)testRemoteOperationDelegateMethodIsCalledWithNoExistenceOfImplementation
...
+  XCTAssertTrue([documentChanges[0].operation isEqualToString:doc1Op]);
+  XCTAssertTrue([documentChanges[0].documentId isEqualToString:doc1Id]);
+  XCTAssertTrue([documentChanges[0].partition isEqualToString:doc1Ptn]);
+  XCTAssertEqual(documentChanges[0].timestamp, doc1Ts);
+  XCTAssertTrue([documentChanges[1].operation isEqualToString:doc2Op]);
+  XCTAssertTrue([documentChanges[1].documentId isEqualToString:doc2Id]);
+  XCTAssertTrue([documentChanges[1].partition isEqualToString:doc2Ptn]);
+  XCTAssertEqual(documentChanges[1].timestamp, doc2Ts);
+
+  OCMVerifyAll(dataMock);
```
jaelim-ms: Do you need this line?

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2819,4 +2821,125 @@ - (void)testRemoteOperationDelegateMethodIsCalledWithNoExistenceOfImplementation
                                }];
 }
 
+- (void)testDocumentChangeDelegateMethodIsCalled {
```
jaelim-ms: Please follow `// If`, `// When`, and `// Then` convention.

AppCenterData/AppCenterData/MSData.m
```diff
@@ -823,6 +826,40 @@ - (void)authTokenContext:(MSAuthTokenContext *)__unused authTokenContext didUpda
   }
 }
 
+#pragma mark - MSServiceNotificationDelegate
+
+- (void)appCenter:(MSAppCenter *)__unused appCenter didReceiveServiceNotification:(NSDictionary<NSString *, NSString *> *)notificationData {
```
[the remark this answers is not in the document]
clpolet: This is how the delegation design works. You can think of it as the "event" emitter. I like to think that it's also good to avoid method signature conflicts in whatever class it's adopted.

AppCenterData/AppCenterData/Model/MSDocumentChange.m
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+#import "MSDocumentChange.h"
+
+@implementation MSDocumentChange
+
+static NSString *const kMSDocumentChangeId = @"id";
+static NSString *const kMSDocumentChangePartition = @"ptn";
+static NSString *const kMSDocumentChangeOperation = @"op";
+static NSString *const kMSDocumentChangeTimestamp = @"_ts";
```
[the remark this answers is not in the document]
thomasf7: In the design review discussions we abbreviated the properties to reduce the message size since they are repeated for every document included in the change notification and the more overhead means the smaller the batch sizes we can support. The underscore for timestamp is because this is mapped directly from the document in cosmosdb and this is the format for timestamp there.

Yes partition is exactly that - a value to differentiate user and public/readonly data. In the future this will also support the name of shared partitions

AppCenterData/AppCenterData/Internal/MSDataPrivate.h
```diff
@@ -91,6 +97,11 @@ static NSString *const kMSDefaultApiUrl = @"https://tokens.appcenter.ms/v0.1";
  */
 - (void)processPendingOperations;
 
+/**
+ * MSServiceNotificationDelegate.
```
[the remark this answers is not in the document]
thomasf7: Removed this method from the header as it is not needed - I had added before MSServiceNotificationDelegate had been added in the previous PR and forgot to remove.

Requested doc comments are in MSServiceNotificationDelegate already.

AppCenterData/AppCenterDataTests/MSDataTests.m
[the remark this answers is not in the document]
thomasf7: Updated tests to use If/When/Then (+ `//Clean up` for stop mocking which I saw in other files)

AppCenterData/AppCenterDataTests/MSDataTests.m
```diff
@@ -2819,4 +2821,125 @@ - (void)testRemoteOperationDelegateMethodIsCalledWithNoExistenceOfImplementation
...
+  id dataDelegateMock = OCMProtocolMock(@protocol(MSDataDelegate));
+  __block NSArray<MSDocumentChange *> *documentChanges = nil;
+  OCMStub([dataDelegateMock data:self.sut didReceiveDocumentChangeNotification:OCMOCK_ANY]).andDo(^(NSInvocation *invocation) {
+    [invocation getArgument:&documentChanges atIndex:3];
+  });
+  [MSData setDelegate:dataDelegateMock];
+
+  NSString *const doc1Op = @"replace";
+  NSString *const doc1Id = @"BHGVGD5";
+  NSString *const doc1Ptn = @"user";
```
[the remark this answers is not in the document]
thomasf7: Updated all variables to full names e.g. `document1Partition`.

AppCenterData/AppCenterData/MSData.m
thomasf7: I won't argue about methods having multiple return points vs nesting :) 

I have:
1. Added a return to remove the else block (as suggested)
1. Changed the check of documentChanges count to `== 0` and added a LogDebug before returning to reduce another level of nesting. (Open to suggestions on different log level - this is a no-op, but may indicate something odd going on in the backend so could maybe go higher - thoughts?)
1. Combined check of strongDelegate and `[strongDelegate respondsToSelector...` to remove another level of nesting.

AppCenterData/AppCenterDataTests/MSDataTests.m
thomasf7: Good point - removed

AppCenterData/AppCenterDataTests/MSDataTests.m
thomasf7: Nothing - removed

AppCenterData/AppCenterDataTests/MSDataTests.m
thomasf7: Updated to follow convention

AppCenterData/AppCenterDataTests/MSDataTests.m
jaelim-ms: Ah.. OK. Then you can declare
```
@interface MSMockDataDelegate : NSObject <MSDataDelegate>
@end
```
in this file right below imports. I don't think `MSMockDataDelegate` is used by any other classes.

thomasf7: No it isn't used anywhere else.

I have removed the files and added both the interface and implementation to the top of this file. I can't see a way to mock just the interface without declaring an implementation as it causes compile errors.

PR #1858, opened 2019-11-01
Update AppCenterDistribute.xcconfig

GrahamFlood:
Need to output strings as UTF-16 so we can process them later.

Tested build and validated encoding on exported strings files.


jwallra:
Hi @GrahamFlood - please make sure your branch is up-to-date with the latest development branch.

There's also a failing test, but it shouldn't be related to your changes.  Do the tests pass locally?  They will be re-run when you update your branch with `develop`.

Issue #1725, opened 2019-08-12
Crash reports for uncaught exceptions are missing important information


hypherion2:
### **Description**
Crash reports for exceptions thrown by the OS frameworks do not include exception information or a proper backtrace.

As per documentation, I've enabled the NSApplicationCrashOnExceptions to catch AppKit exception then I forced a test crash by calling the method substringFromIndex with an invalid index on a NSString.

The following information shows up on AppCenter portal for the crash report (in the RAW format):

> Exception Type:  SIGILL
Exception Codes: ILL_NOOP at 0x0
Crashed Thread:  0

>Thread 0 Crashed:
0   AppKit                               0x00007fff3723ca67 -[NSApplication _crashOnException:] + 109
1   AppKit                               0x00007fff3723c948 -[NSApplication reportException:] + 915
2   AppKit                               0x00007fff36e318f9 -[NSApplication run] + 887
3   AppKit                               0x00007fff36e20d7c NSApplicationMain + 776
4   MyApp                               0x000000011061087b main (main.mm:29)
5   libdyld.dylib                        0x00007fff656e13d5 start + 0

This report gives me absolutely no information on where or what caused my application to crash. 
It also includes some undefined values for the timestamps in the raw report:
>Date/Time:       undefined
Launch Time:     undefined

By comparison, the same crash when using the HockeyApp SDK integrated into the app and subclassing BITCrashExceptionApplication per HockeyApp documentation looks like this:

>Exception Type:  SIGABRT
Exception Codes: #0 at 0x7fff6581c2c6
Crashed Thread:  0

>Application Specific Information:
*** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSCFString substringFromIndex:]: Index 50 out of bounds; string length 10'

>Last Exception Backtrace:
0   CoreFoundation                       0x00007fff39869ce5 __exceptionPreprocess + 232
1   libobjc.A.dylib                      0x00007fff63f13a17 objc_exception_throw + 48
2   CoreFoundation                       0x00007fff39869b2f +[NSException raise:format:] + 201
3   Foundation                           0x00007fff3ba514de -[NSString substringFromIndex:] + 126
4   MyApp                               0x000000010235eae7 -[MyModule badMethod:] (MyModule.m:680)
and so on...full stack from my app continues.

### **Details**

1. Which SDK version are you using?
    - 2.2.0
2. Which OS version did you experience the issue on?
    - OS X 10.14.5
[...]

PR #1868, opened 2019-11-08
Fix handling macOS exception details

MatkovIvan:

* [x] Did you add unit tests?

## Description

Fix handling macOS exception details

## Related PRs or issues

#1725
[AB#66586](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/66586)


AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,210 @@
...
+ */
+- (void)ms_reportException:(NSException *)exception {
+
+  // Don't invoke the registered UncaughtExceptionHandler if we are currently debugging this app!
+  if (![MSAppCenter isDebuggerAttached] && exception) {
+
+    /*
+     * We forward this exception to PLCrashReporters UncaughtExceptionHandler.
+     * If the developer has implemented their own exception handler and that one is invoked before PLCrashReporters exception handler and
+     * the developers exception handler is invoking this method it will not finish it's tasks after this call but directly jump into
```
michaelsmitha:
```suggestion
     * the developers exception handler is invoking this method it will not finish its tasks after this call but directly jump into
```

MatkovIvan:
@michaelsmitha you have found a lot of typos in HockeyApp's comment 😅 
https://github.com/bitstadium/HockeySDK-Mac/blob/5.1.1/Classes/CrashReporting/BITCrashExceptionApplication.h#L3

michaelsmitha:
> @michaelsmitha you have found a lot of typos in HockeyApp's comment 😅
> https://github.com/bitstadium/HockeySDK-Mac/blob/5.1.1/Classes/CrashReporting/BITCrashExceptionApplication.h#L3

Yeah :) I know you didn't write it, but if we are putting it in, we might as well fix it :)

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,210 @@
...
+ *
+ *
+ * References:
+ *   https://developer.apple.com/library/mac/documentation/cocoa/Conceptual/Exceptions/Tasks/ControllingAppResponse.html#//apple_ref/doc/uid/20000473-BBCHGJIJ
+ *   http://stackoverflow.com/a/4199717/474794
+ *   http://stackoverflow.com/a/3419073/474794
+ *   http://macdevcenter.com/pub/a/mac/2007/07/31/understanding-exceptions-and-handlers-in-cocoa.html
+ *
+ */
+@implementation NSApplication (CrashException)
```
clpolet: I'd prefix the category and category methods names since this is a category over an Apple class.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,214 @@
...
+ *
+ *   Overwrite and extend specific methods of `NSApplication`. Can be implemented via subclassing NSApplication or
+ *   by using a category.
+ *
+ *   Benefits:
+ *
+ *     1. Solves scenarios 2 (by overwriting `reportException:`) and 3 (by overwriting `sendEvent:`)
+ *
+ *     2. Subclassing approach isn't enforcing the mechanism onto apps and lets developers opt-in.
+ *        (Category approach would enforce it and rather be a problem of this soltuion.)
```
clpolet:
```suggestion
 *        (Category approach would enforce it and rather be a problem of this solution.)
```

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -673,6 +676,9 @@ - (void)configureCrashReporterWithUncaughtExceptionHandlerEnabled:(BOOL)enableUn
 
     // Add a handler for C++-Exceptions.
     [MSCrashesUncaughtCXXExceptionHandlerManager addCXXExceptionHandler:uncaught_cxx_exception_handler];
+
+    // Activate category to handle additional crash details.
+    [MSCrashesCategory activateCategory];
```
clpolet: We need to provide a way for customers to deactivate swizzling but yet with the possibility to use this feature at the cost of some more integration for them. It's a concept adopted by the SDK in other places where we do swizzling.

MatkovIvan: After thinking more about this point, I have more doubts about it.
From our docs:

> App Center uses swizzling to automatically forward the app delegate's methods to App Center services to improve SDK integration. Method swizzling is a way to change the implementation of methods at runtime. There is a possibility of conflicts with other third party libraries or the app delegate itself.

It makes sense to make it configurable for **AppDelegate** to avoid possible conflicts, but this is another case:
* It swizzles NOT delegate's method
* There is a requirement for only one crash reporting library in the project already
* We can't reuse the same config flag
* It's already under another flag and NOT enabled by default

Also, to make it configurable we need to provide a way to do it manually without swizzling. I see two options how:
1. Add **second** implementation of these methods wrapped to custom `NSApplication` class and document how to configure it in `.plist`. This way was used by HockeyApp.
2. Add static public method to `MSCrashes` to forward the exception details from these overrides. (I guess this method will confuse users)

All of above requires adding new public API, so let's discuss it first.

jaelim-ms: This is improving SDK behavior, not providing an optional functionality. It seems odd to have a config to get important information about crashes, I don't think this is the one that we provide a way to opt-in/out.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,62 @@
...
+  MSMockUserDefaults *settings = [MSMockUserDefaults new];
+  [settings setObject:@YES forKey:@"NSApplicationCrashOnExceptions"];
+  id applicationMock = OCMPartialMock([NSApplication sharedApplication]);
+  id appCenterMock = OCMClassMock([MSAppCenter class]);
+  OCMStub([appCenterMock isDebuggerAttached]).andReturn(NO);
+  id crashesMock = OCMPartialMock([MSCrashes sharedInstance]);
+  OCMStub([crashesMock exceptionHandler]).andReturn((NSUncaughtExceptionHandler *)exceptionHandler);
+#endif
+  // When
+  [MSCrashesCategory activateCategory];
```
Jamminroot: This line gets zoned out by `#if`s in case it's not `TARGET_OS_OSX`. Is that intended?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
clpolet: I saw customers just deactivating swizzling simply because of an internal policy. So yes we need this.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
MatkovIvan: yes

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
MatkovIvan: So, what API/way to do it without swizzling looks best for you?

clpolet: OK so what I would do is try to adapt the patterns we used for swizzling in the SDK 1) for coherency and 2) because it has worked great so far. It means exposing a static method in Crashes to process `reportException:` and `sendEvent:` but it's true that it's a bit confusing and generic so I would rename to `applicationReportException:` and `applicationSendEvent:`. The swizzling will just call these methods under the hood so the processing logic is in only one place.

When Swizzling is deactivated it will be almost as documented by HA. I don't think we need to provide an NSApplication subclass, it will just be for the few apps deactivating swizzling. So that would be:
* Swizzling deactivation should follow the same pattern as usual with the plist boolean for deactivation (i.e.: `AppCenterApplicationForwarderEnabled` set to NO). 
* Then customers subclass NSApplication and update the plist accordingly.
* Implement `sendEvent:` and `applicationSendEvent:` to forward to super implementation and the corresponding methods in Crashes.
* Done.

Swizzling activation should be done in `+load` method for additional stability, I am not super worried about `reportException` but `sendEvent` is probably called more often during the app lifecycle.

MatkovIvan: Almost OK with it except exposing API for `sendEvent:` - only `reportException:` is enough

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.h
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSCrashesCategory : NSObject
```
jaelim-ms: Please consider renaming the class.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,72 @@
...
+@interface MSCrashesCategoryTests : XCTestCase
+
+@end
+
+@implementation MSCrashesCategoryTests
+
+- (void)testActivateCategory {
+
+  // If
+#if TARGET_OS_OSX
```
jaelim-ms: This should be for the entire method unless you want to verify something after `[MSCrashesCategory activateCategory]` in iOS/tvOS.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,72 @@
...
+static NSException *lastException;
+static void exceptionHandler(NSException *exception) { lastException = exception; }
+
+@interface MSCrashesCategoryTests : XCTestCase
+
+@end
+
+@implementation MSCrashesCategoryTests
+
+- (void)testActivateCategory {
```
jaelim-ms: There are many combination of `NSBundle`, `NSUserDefaults` and `isDebuggerAttached` values. I can see test coverage at least for `NSBundle`. Please add more unit tests and also consider splitting into two separate tests for this one.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,72 @@
...
+#if TARGET_OS_OSX
+  id bundleMock = OCMClassMock([NSBundle class]);
+  OCMStub([bundleMock objectForInfoDictionaryKey:@"AppCenterApplicationForwarderEnabled"]).andReturn(@NO);
+  OCMStub([bundleMock mainBundle]).andReturn(bundleMock);
+  MSMockUserDefaults *settings = [MSMockUserDefaults new];
+  [settings setObject:@YES forKey:@"NSApplicationCrashOnExceptions"];
+  id applicationMock = OCMPartialMock([NSApplication sharedApplication]);
+  id appCenterMock = OCMClassMock([MSAppCenter class]);
+  OCMStub([appCenterMock isDebuggerAttached]).andReturn(NO);
+  id crashesMock = OCMPartialMock([MSCrashes sharedInstance]);
```
jaelim-ms: You should reset the Crashes instance at the end of tests. Consider adding `tearDown` for that one.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,72 @@
...
+@interface NSApplication (CrashException)
+
+- (void)ms_reportException:(NSException *)exception;
+- (void)ms_sendEvent:(NSEvent *)theEvent;
+
+@end
+
+#endif
+
+static NSException *lastException;
```
jaelim-ms: Why is it static? Because there is only one unit test?

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
```diff
@@ -5,6 +5,7 @@
 
 #import "MSAppCenterInternal.h"
 #import "MSAppleErrorLog.h"
+#import "MSCrashReporter.h"
```
jaelim-ms: ??

SasquatchMac/SasquatchMacSwift/AppDelegate.swift
```diff
@@ -53,6 +53,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, MSCrashesDelegate, MSPushDel
       return true
     })
 
+    // Enable catching uncaught exceptions thrown on the main thread.
+    UserDefaults.standard.register(defaults: ["NSApplicationCrashOnExceptions": true])
```
jaelim-ms: Swift app shouldn't use swizzling. Please add the key to info.plist.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.h
clpolet: `MSApplicationForwarder` would align with other swizzling classes in the SDK.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.h
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Activate category for UIViewController.
+ */
++ (void)activateCategory;
```
clpolet: Consider renaming for instance `registerForwarding` or `registerSwizzling`.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,223 @@
...
+    return;
+  }
+#if TARGET_OS_OSX
+  if (isCrashOnExceptionsEnabled()) {
+
+    /*
+     * Solution for Scenario 2
+     *
+     * Catch all exceptions that are being logged to the console and forward them to our
+     * custom UncaughtExceptionHandler
```
clpolet:
```suggestion
     * custom UncaughtExceptionHandler.
```

AppCenterCrashes/AppCenterCrashes/Internals/MSCrashesPrivate.h
```diff
@@ -1,57 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#import <array>
```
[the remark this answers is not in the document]
MatkovIvan: it was required to include this file into NOT objective-c++ file

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,223 @@
...
+  } else {
+    MSLogDebug([MSCrashes logTag], @"Application forwarder for info.plist key '%@' disabled. It won't use swizzling.",
+               kMSAppCenterApplicationForwarderEnabledKey);
+    return;
+  }
+#if TARGET_OS_OSX
+  if (isCrashOnExceptionsEnabled()) {
+
+    /*
+     * Solution for Scenario 2
```
clpolet:
```suggestion
     * Solution for Scenario 2:
```

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * Catch all exceptions that are being logged to the console and forward them to our
+     * custom UncaughtExceptionHandler
+     */
+    swizzleReportException();
+
+    /*
+     * Solution for Scenario 3
```
clpolet:
```suggestion
     * Solution for Scenario 3:
```

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * Exceptions that happen inside an IBAction implementation do not trigger a call to
+     * [NSApplication reportException:] and it does not trigger a registered UncaughtExceptionHandler
+     * Hence we need to catch these ourselves, e.g. by overwriting sendEvent: as done right here
```
clpolet:
```suggestion
     * Hence we need to catch these ourselves, e.g. by overwriting sendEvent: as done right here.
```

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -706,7 +730,6 @@ - (void)startDelayedCrashProcessing {
   // This must be performed asynchronously to prevent a deadlock with 'unprocessedCrashReports'.
   dispatch_time_t delay = dispatch_time(DISPATCH_TIME_NOW, kMSCrashProcessingDelay);
   dispatch_after(delay, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-
```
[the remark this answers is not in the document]
MatkovIvan: it was changed by auto-formatting. Do we need manually edit it back each time?

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
MatkovIvan: static here means that it will NOT create symbol for it to access from other files

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
MatkovIvan: See 68 line

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
MatkovIvan: https://github.com/microsoft/appcenter-sdk-apple/pull/1868/files#diff-572dc6f7abeb81a0ab6603650a60ade5L9

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
```diff
@@ -0,0 +1,72 @@
...
+#import "MSAppCenterInternal.h"
+#import "MSCrashesCategory.h"
+#import "MSCrashesPrivate.h"
+#import "MSMockUserDefaults.h"
+#import "MSTestFrameworks.h"
+#import "MSUtility+Application.h"
+
+#if TARGET_OS_OSX
+
+@interface NSApplication (CrashException)
```
clpolet: Can we we remove this now.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
clpolet: OK fair enough if swizzling + system call to `sendEvent` and `reportException` are guaranteed to all run in main thread then there shouldn't be any problem yes.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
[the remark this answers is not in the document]
jaelim-ms: What's weird..  Let's keep it for now and we need to check why formatter works like that.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
jaelim-ms: I'm saying `[MSCrashes resetSharedInstance]`.

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
jaelim-ms: My point is that this will be used more than one method in this class once you add more unit tests. You don't see any of side effects by using static variable?

MatkovIvan: If you are saying about `static` modifier - it's better to have it to avoid possible name conflicts at linking.

if about global var in general - we need it to have a pointer to pure c function and it's required to check the behavior. There is no way to mock it locally

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.h
MatkovIvan: @clpolet not really. It's already aligned to analytics

AppCenterCrashes/AppCenterCrashesTests/MSCrashesCategoryTests.m
MatkovIvan: > also consider splitting into two separate tests for this one.

Actually I can't really separate it into separate tests because we can swizzle only once

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSCrashesCategory.m
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSCrashesCategory.h"
+#import "MSAppCenterInternal.h"
+#import "MSCrashesInternal.h"
+#import "MSCrashesPrivate.h"
+#import "MSUtility+Application.h"
+#import <objc/runtime.h>
```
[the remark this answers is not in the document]
MatkovIvan: > Then this line will be placed at the top after reformat.

it won't, I'll do it manually

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.h
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSApplicationForwarder : NSObject
+
+/**
+ * Activate category for UIViewController.
```
clpolet:
```suggestion
 * Register forwarding on `NSApplication` via swizzling. 
```

SasquatchMac/SasquatchMacSwift/Info.plist
```diff
@@ -28,5 +28,7 @@
 	<string>SasquatchMac</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
+	<key>AppCenterApplicationForwarderEnabled</key>
```
clpolet: We need to set up the manual integration on this app to test it.

PR #1852, opened 2019-10-28
[Fix] MSAlertControler on iOS 13

russelarms:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [ ] Are tests passing locally?
* [ ] Are the files formatted correctly?
* [ ] Did you add unit tests?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

MSAlertController doesn't appear when a project is built with Xcode 11.
This PR adds UIScene support to MSAlertController to fix it.

## Related PRs or issues

List related PRs and other issues.

## Misc

https://developer.apple.com/design/human-interface-guidelines/ios/system-capabilities/multiple-windows/

PR #1873, opened 2019-11-12
Fix missing alert dialogs on iOS 13 when project uses new scenes API

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Fix missing alert dialogs on iOS 13 when project uses new scenes API

## Related PRs or issues

#1852
[AB#72440](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72440)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### App Center Distribute
 
-* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new UIScenes API.
+* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new `UIScene` API (multiple scenes are not yet supported).
 
 ___
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center Distribute
 
-* **[Fix]** Fix missing alert dialogs on iOS 13 when project uses new scenes API.
+* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new UIScenes API.
 
 ___
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 2.5.2 (Under development)
 
+### App Center Distribute
+
+* **[Fix]** Fix missing alert dialogs on iOS 13 when project uses new scenes API.
+
 ___
 
 ## Version 2.5.1
--- a/Vendor/iOS/MSAlertController/MSAlertController.m
+++ b/Vendor/iOS/MSAlertController/MSAlertController.m
@@ -146,19 +146,19 @@ static dispatch_queue_t alertsQueue;
   }
 }
 
-#define DISPATCH_SELECTOR(class, selectorName, ...) ({ \
+#define MS_DISPATCH_SELECTOR(result, class, selectorName, ...) ({ \
   SEL selector = NSSelectorFromString(@#selectorName); \
   IMP impl = [class methodForSelector:selector]; \
-  ((id (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
+  ((result (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
 })
 
 + (void)makeKeyAndVisible {
   if (@available(iOS 13.0, *)) {
-    UIApplication *application = (UIApplication *)DISPATCH_SELECTOR([UIApplication class], sharedApplication);
-    NSSet *scenes = (NSSet *)DISPATCH_SELECTOR(application, connectedScenes);
+    UIApplication *application = MS_DISPATCH_SELECTOR(UIApplication *, [UIApplication class], sharedApplication);
+    NSSet *scenes = MS_DISPATCH_SELECTOR(NSSet *, application, connectedScenes);
     NSObject *windowScene = nil;
     for (NSObject *scene in scenes) {
-      NSInteger activationState = (NSInteger)DISPATCH_SELECTOR(scene, activationState);
+      NSInteger activationState = MS_DISPATCH_SELECTOR(NSInteger, scene, activationState);
       if (activationState == 0 /* UISceneActivationStateForegroundActive */) {
         windowScene = scene;
         break;
@@ -167,7 +167,7 @@ static dispatch_queue_t alertsQueue;
     if (!windowScene) {
       windowScene = scenes.anyObject;
     }
-    DISPATCH_SELECTOR(window, setWindowScene:, windowScene);
+    MS_DISPATCH_SELECTOR(void, window, setWindowScene:, windowScene);
   }
   [window makeKeyAndVisible];
 }
--- a/Vendor/iOS/MSAlertController/MSAlertController.m
+++ b/Vendor/iOS/MSAlertController/MSAlertController.m
@@ -137,13 +137,34 @@ static dispatch_queue_t alertsQueue;
       [alertsToBePresented removeObjectAtIndex:0];
     });
     dispatch_async(dispatch_get_main_queue(), ^{
-      [window makeKeyAndVisible];
+      [MSAlertController makeKeyAndVisible];
       [window.rootViewController presentViewController:nextAlert animated:animated completion:nil];
     });
   } else {
     window.hidden = YES;
     alertIsBeingPresented = NO;
   }
+}
+
+#define DISPATCH_SELECTOR(class, selectorName, ...) ({ \
+  SEL selector = NSSelectorFromString(@#selectorName); \
+  IMP impl = [class methodForSelector:selector]; \
+  ((id (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
+})
+
++ (void)makeKeyAndVisible {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 2.5.2 (Under development)
 
+### App Center Distribute
+
+* **[Fix]** Fix missing alert dialogs on iOS 13 when project uses new scenes API.
```
```suggestion
* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new UIScenes API.
```

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,4 +146,25 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+  ((id (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
+})
+
++ (void)makeKeyAndVisible {
+  if (@available(iOS 13.0, *)) {
+    UIApplication *application = (UIApplication *)DISPATCH_SELECTOR([UIApplication class], sharedApplication);
+    NSSet *scenes = (NSSet *)DISPATCH_SELECTOR(application, connectedScenes);
+    for (NSObject *scene in scenes) {
+      NSInteger activationState = (NSInteger)DISPATCH_SELECTOR(scene, activationState);
+      if (activationState == 0 /* UISceneActivationStateForegroundActive */) {
```
michaelsmitha: `UISceneActivationStateForegroundActive` should be a constant

MatkovIvan: @michaelsmitha it is, but it's not available in old XCode and has name conflict in new

MatkovIvan:
> If no active scenes is found what happened?

Nothing. It won't show.

 > how the states of the scenes are monitored in this case.

Just refresh the active one on showing each alert

> When will it reattempt to display the alert.

Never. It doesn't have this logic.

> Would be ideal if we can reproduce in one of our test apps.

Do we agree to bump XCode to 11 for one of our app?

Vendor/iOS/MSAlertController/MSAlertController.m
clpolet: The issue I see here is when there is no active scenes to begin with. For instance in Distribute the alert can pop up anytime, likely at the beginning of the user session but it's triggered by an HTTP call result so it's async. At that time all of the scenes can be hidden by, let's say, the drop down notification center, in this case all scenes are inactive. We need to find a way to cover this case too.

jaelim-ms: Would be good to add a short description there. 😉

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,4 +146,30 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
   }
 }
 
+#define MS_DISPATCH_SELECTOR(result, class, selectorName, ...) ({ \
```
jaelim-ms: Can you move this to the Utility class? I could see this in another PR, let's avoid having duplicated code here and there.

jaelim-ms:
@MatkovIvan 
> Do we agree to bump XCode to 11 for one of our app?

We only use Xcode 10.1 for SDK build. Ideally the test app should be built on all of Xcode versions.

jaelim-ms:
The code looks good to me but I wonder if we need to test various scenarios including multiple windows on iPadOS.

Vendor/iOS/MSAlertController/MSAlertController.m
MatkovIvan: No, it's in a `Vendor` folder

MatkovIvan:
@jaelim-ms I know, that's why it's a question about bumping.
A test app with new scenes requires XCode 11. Anyway test app changes will be made in separate PR.

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,4 +146,30 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+    NSObject *windowScene = nil;
+    for (NSObject *scene in scenes) {
+      NSInteger activationState = MS_DISPATCH_SELECTOR(NSInteger, scene, activationState);
+      if (activationState == 0 /* UISceneActivationStateForegroundActive */) {
+        windowScene = scene;
+        break;
+      }
+    }
+    if (!windowScene) {
+      windowScene = scenes.anyObject;
```
annakocheshkova: Are you sure that's not a broken state?

MatkovIvan: https://github.com/microsoft/appcenter-sdk-apple/pull/1873#discussion_r345953000

clpolet: Taking any scene seems hazardous. I haven't looked deeply into this but it seems that observing the connected scenes seems to be the only reliable way to be notified of the next active window although quite cumbersome. Or maybe a way to intercept messages to scene delegates? Is there notifications to listen to, similar to the app lifecycle notifications?

MatkovIvan: Well, I guess this is OK for **this** fix in the `Vendor/` utility folder, but as long term solution we need to make more changes not only in this alerts.

Based on [this](https://stackoverflow.com/a/56508769/1703290) good description:
> the UIApplication won't send background and foreground lifecycle messages to your app delegate. That means the code in these methods won't run:
> - `applicationDidBecomeActive`
> - `applicationWillResignActive`
> - `applicationDidEnterBackground`
> - `applicationWillEnterForeground`

So this means that most likely everywhere we use it (like session tracking, distribute, etc) we need some changes.

> Is there notifications to listen to, similar to the app lifecycle notifications?

Yes, `NSNotificationCenter` has `UISceneDidActivateNotification` for example that replaces application's one.

So, I guess we should have separate work item for:
- Adding new scenes to one of our app
- Refactor and centralize the work with app state somewhere in the core
- Probably raise the question to bump XCode version

Vendor/iOS/MSAlertController/MSAlertController.m
clpolet: Right, but can you move it up in this file so it's not in between two instance or class methods :) ?

Vendor/iOS/MSAlertController/MSAlertController.m
clpolet: From what I could test this summer (would need to double check today) on the classic application lifecycle notifications still works even though scenes are implemented but of course it only captures life cycle of the application which is what we want so I am not very worried.

clpolet: This is a partial fix meaning it will fix the default 1 scene implementation provided by XCode but it won't support the multiple windows feature regardless of the number of scenes. For instance, beside the lifecycle cases there is another concern: in some cases we want the alert to be modal across all windows (mandatory update) but at this point it's broken. This is indeed bigger than just this PR we'll need design decisions and how to move forward. For now let's make the scope of this PR clear. It seems to be just fixing the "the default 1 scene implementation provided by XCode but not multiple windows apps".

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 * **[Fix]** Fix sending crashes if an application is launched in background.
 * **[Fix]** Validate error attachment size to avoid server error or out of memory issues (using the documented limit which is 7MB).
 
+### App Center Distribute
+
+* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new UIScenes API.
```
MatkovIvan:
```suggestion
* **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new `UIScene` API (multiple scenes are not yet supported).
```

jaelim-ms:
@MatkovIvan What I meant there was just testing locally. You don't need to add a functionalities to test on the test apps, just create a new app or modify the exist apps (but you don't need to commit) and see if it works without any issues with multiple windows.

russelarms:
@jaelim-ms I tested this class with multiple windows project from the [sample](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/supporting_multiple_windows_on_ipad) and it seems to work fine.
I called an alert dialog with _DispatchQueue.main.asyncAfter_.
For cases where any of the app's windows are on the screen, the dialog is shown in an active window.
For the case when I call a dialog and switch to a different app: after returning to the test app the dialog is shown inside the window in which it was originally invoked.
Let me know if we have to test any specific multiple windows case.

jaelim-ms:
@russelarms That should be enough. Thanks for the verification.

jaelim-ms:
@russelarms @MatkovIvan I don't see any answers for [this](https://github.com/microsoft/appcenter-sdk-apple/pull/1873#discussion_r345953000) feedback. Please let me know how you resolved the feedback.

MatkovIvan:
@jaelim-ms Yes, it has been resolved. I added the fallback for it. See https://github.com/microsoft/appcenter-sdk-apple/pull/1873/commits/a3c73acce01dcbf03d1fe5f07c748e886e20ea57

PR #1875, opened 2019-11-12
change warning to info for a unnecessary and confusing log

CmonsterC:

## Description

change warning to info for a unnecessary and confusing log

## Related PRs or issues

[AB#71820](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71820)


AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -222,7 +222,7 @@ - (void)setAuthTokenHistory:(nullable NSArray<MSAuthTokenInfo *> *)authTokenHist
     [MS_USER_DEFAULTS setObject:encryptedData forKey:kMSAuthTokenHistoryKey];
     MSLogDebug([MSAppCenter logTag], @"Saved new history state.");
   } else {
-    MSLogWarning([MSAppCenter logTag], @"Failed to save new history state.");
+    MSLogInfo([MSAppCenter logTag], @"Failed to save new history state.");
```
dhei: I don't think we want to change the log level when failed to save token in storage. Warning level is still appropriate.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Version 2.5.2 (Under development)
 
+* **[Fix]** Fix warnings in when make authTokenHistory operations.
```
dhei: `authTokenHistory` is an implementation detail of the SDK. It would be better if we provide a more meaningful description of the bug we fixed in the changelog.

Maybe something along the lines of:
```
* **[Fix]** Remove confusing warnings about "Failed to retrieve history state or none was found" when it doesn't make sense.  
```

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: In some customer complains, this is also complained, should we fix that as well?

dhei: okay, that's fair, then we need to address this code path as well.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -207,7 +207,7 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
   if (history) {
     MSLogDebug([MSAppCenter logTag], @"Retrieved history state.");
   } else {
-    MSLogWarning([MSAppCenter logTag], @"Failed to retrieve history state or none was found.");
+    MSLogInfo([MSAppCenter logTag], @"Failed to retrieve history state or none was found.");
```
clpolet: This is sometimes can be a real issue, I am thinking about dissociating "a decryption" issue from "there is nothing in the storage". The first one can stay as a warning and the second one can be an info.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
clpolet: I don't remember exactly but since authTokenHistory can be null it sounds to me that passing nil or empty array to this method is indeed not a warning and can be expected though if the encryption or persisting a valid value is failing then it's a warning.

jaelim-ms:
Please use `AB#{number}` format for work item link. You don't need to copy a link, just type it (e.g. `AB#71820`). Webhook will automatically add a link in description.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
jaelim-ms: We should split the log message with a proper log level.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: > I don't remember exactly but since authTokenHistory can be null it sounds to me that passing nil or empty array to this method is indeed not a warning and can be expected though if the encryption or persisting a valid value is failin

Yeah, I do see some caller is setting nil. will seperate that out.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -207,7 +207,14 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
   if (history) {
     MSLogDebug([MSAppCenter logTag], @"Retrieved history state.");
   } else {
-    MSLogWarning([MSAppCenter logTag], @"Failed to retrieve history state or none was found.");
+    if (decryptedData) {
```
jaelim-ms: Isn't it `!decryptedData`?
Do this before the line 206.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -207,7 +207,14 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
...
+      MSLogWarning([MSAppCenter logTag], @"Failed to unarchive decryptedData.");
+    } else if (encryptedData) {
```
jaelim-ms: Isn't it `!encryptedData`?
Please check this condition before the line 205.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -207,7 +207,14 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
...
+      MSLogWarning([MSAppCenter logTag], @"Failed to get decryptedData.");
+    } else {
+      MSLogInfo([MSAppCenter logTag], @"No token history was found.");
```
jaelim-ms: Line 202 should also need this one.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -207,7 +207,14 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
   if (history) {
     MSLogDebug([MSAppCenter logTag], @"Retrieved history state.");
   } else {
```
jaelim-ms: Do not nest `if` statement. You can do `else if` here.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -221,8 +228,12 @@ - (void)setAuthTokenHistory:(nullable NSArray<MSAuthTokenInfo *> *)authTokenHist
     self.authTokenHistoryArray = authTokenHistory;
     [MS_USER_DEFAULTS setObject:encryptedData forKey:kMSAuthTokenHistoryKey];
     MSLogDebug([MSAppCenter logTag], @"Saved new history state.");
+  } else if (decryptedData) {
```
jaelim-ms: Isn't it `!decryptedData`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -221,8 +228,12 @@ - (void)setAuthTokenHistory:(nullable NSArray<MSAuthTokenInfo *> *)authTokenHist
...
+    MSLogWarning([MSAppCenter logTag], @"Failed to encrypt data.");
+  } else if ([authTokenHistory count] > 0) {
```
jaelim-ms: Isn't it `[authTokenHistory count] == 0`?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: no.  when code comes here. means `history` is null  and we need to log  warning if `decrpytedData` is NOT null, meaning `[NSKeyedUnarchiver unarchiveObjectWithData:decryptedData] ` is not returning correctly.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: no. when code comes here. means history is null and we need to log warning if decrpytedData is NOT null, meaning [NSKeyedUnarchiver unarchiveObjectWithData:decryptedData] is not returning correctly.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: Line 202 is the positive path. is that necessary?

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: no. Line 218 is something extra and common to all branch in the nested `if` and `else`

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: no. Line 218 is something extra and common to all branch in the nested `if` and `else`

CmonsterC: no. same as the previous function. we should not simply check null and log warning by each step. because people can pass in null.  we should log warning only when previous steps are not returning null but current step is returning null.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: no. same as the previous function. we should not simply check null and log warning by each step. because people can pass in null.  we should log warning only when previous steps are not returning null but current step is returning null.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: we should not simply check null and log warning by each step. because people can pass in null. we should log warning only when previous steps are not returning null but current step is returning null.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: we should not simply check null and log warning by each step. because people can pass in null. we should log warning only when previous steps are not returning null but current step is returning null.

CmonsterC:
@jaelim-ms if could be a little bit confusing to do the logging in a intuitive direction, we changed it to let it be easier to be understood. Please review again. Thanks.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: but we will try to think it in a more intuitive way

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: but we will try to think it in a more intuitive way

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: but we will try to think it in a more intuitive way

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
CmonsterC: but we will try to think it in a more intuitive way

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -202,28 +202,61 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
     return (NSArray<MSAuthTokenInfo *> *)self.authTokenHistoryArray;
   }
   NSData *encryptedData = [MS_USER_DEFAULTS objectForKey:kMSAuthTokenHistoryKey];
+
```
dhei: Our convention of in SDK is to only add a blank newline before comments. Let's remove all the unnecessary newlines for consistency.

CmonsterC:
@jaelim-ms Thanks! This makes thing looks better.

nisteuve:
@clpolet Would you mind having another look? I believe your feedback has been resolved at this stage.

AppCenter/AppCenter/Internals/Context/AuthToken/MSAuthTokenContext.m
```diff
@@ -201,49 +201,52 @@ - (void)removeAuthToken:(nullable NSString *)authToken {
...
-  return (NSArray<MSAuthTokenInfo *> *)self.authTokenHistoryArray;
+- (NSArray<MSAuthTokenInfo *> *)retrieveAuthTokenHistory {
+  NSData *encryptedData = [MS_USER_DEFAULTS objectForKey:kMSAuthTokenHistoryKey];
+  if (!encryptedData) {
+    MSLogInfo([MSAppCenter logTag], @"No auth token history was found.");
+    return nil;
+  }
+  NSData *decryptedData = [self.encrypter decryptData:encryptedData];
+  if (!decryptedData) {
+    MSLogWarning([MSAppCenter logTag], @"Failed to decrypt auth token history.");
```
dhei: Should this be an `MSLogError` instead of `MSLogWarning` as this is an error case?

CmonsterC: sure.

PR #1883, opened 2019-11-18
Fix targets initialization in iOS/macOS apps

annakocheshkova:

## Description

Problem: we were initialising transmission targets too early (before `startFromLibrary` call), as a result at the moment of initialisation there were no `channelGroup` to subscribe `propertyConfigurator` to.

_NOTE:_ on macOS, `startFromLibrary` is called in `viewWillAppear` (due to the fact that `viewDidLoad` is called for all the tabbed app at its start, and we need the logic of starting Analytics at **Transmission** tab only). 
It made the fix more complicated, because the start call at `viewWillAppear` and targets initialisation (which is required to load the table view and fill out `isEnabled` values) are performed concurrently, creating a "race condition". 
Fix: I had to move the target initialisation to `viewWillAppear`, after the start from library call, and then reload the tables again with the actual `isEnabled` values.

## Related PRs or issues

[AB#72259](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72259)

### DIFF
--- a/SasquatchMac/SasquatchMac/TransmissionTargets.swift
+++ b/SasquatchMac/SasquatchMac/TransmissionTargets.swift
@@ -18,7 +18,7 @@ class TransmissionTargets {
     case skip
   }
     
-  static let defaultTransmissionTargetIsEnabled: Bool = TransmissionTargets.startTarget == StartupMode.oneCollector.rawValue || startTarget == StartupMode.both.rawValue
+  static let defaultTransmissionTargetWasEnabled: Bool = TransmissionTargets.startTarget == StartupMode.oneCollector.rawValue || startTarget == StartupMode.both.rawValue
 
   private init() {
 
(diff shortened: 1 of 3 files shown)


SasquatchMac/SasquatchMac/TransmissionTargets.swift
```diff
@@ -18,12 +17,12 @@ class TransmissionTargets {
     case none
     case skip
   }
+    
+  static let defaultTransmissionTargetIsEnabled: Bool = TransmissionTargets.startTarget == StartupMode.oneCollector.rawValue || startTarget == StartupMode.both.rawValue
```
MatkovIvan: It will be computed only once. Fix or rename

annakocheshkova:
Yes I saw @guperrot. Assumed that's irrelevant to the task, but forgot to mention in the channel.

guperrot:
> Yes I saw @guperrot. Assumed that's irrelevant to the task, but forgot to mention in the channel.

I created a bug for that.

PR #1889, opened 2019-11-20
Fix device info is attached to crashes

ilyashumihin:

* [x] Did you add unit tests?

## Description

Fix device info is attached to crashes.
After reading the crash reporter's logs, we link a specific log and a device from the history and after we're deleting the entire history of the devices from storage. We need to save the information about the last device so that we can always link the last log to the device on which the crash occurred.
Actions that caused the error (Before fix):
1. Run the application, invoke the crash
2. Restart the application (AppCenter will try to find crash, link his information with the device from the history and clear the entire history of the devices)
3. Cause the app to crash again
4. Upgrade your app version
5. Restart the application (AppCenter will try to link the crash with the device from the history, but the history has been cleared and we will take the device with an updated version of the application, which is incorrect)

## Related PRs or issues

[AB#70439](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/70439)

### DIFF
--- a/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
+++ b/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
@@ -250,9 +250,13 @@ static MSDeviceTracker *sharedInstance = nil;
 
 - (void)clearDevices {
   @synchronized(self) {
+    // Remember the information about the current device.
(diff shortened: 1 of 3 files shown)


MatkovIvan:
Reopened not from a fork. See #1890

PR #1890, opened 2019-11-20
Fix device info is attached to crashes

ilyashumihin:

* [x] Did you add unit tests?

## Description

Fix device info is attached to crashes.
After reading the crash reporter's logs, we link a specific log and a device from the history and after we're deleting the entire history of the devices from storage. We need to save the information about the last device so that we can always link the last log to the device on which the crash occurred.
Actions that caused the error (Before fix):
1. Run the application, invoke the crash
2. Restart the application (AppCenter will try to find crash, link his information with the device from the history and clear the entire history of the devices)
3. Cause the app to crash again
4. Upgrade your app version
5. Restart the application (AppCenter will try to link the crash with the device from the history, but the history has been cleared and we will take the device with an updated version of the application, which is incorrect)

## Related PRs or issues

[[AB#70439](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/70439)]

### DIFF
--- a/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
+++ b/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
@@ -250,6 +250,7 @@ static MSDeviceTracker *sharedInstance = nil;
 
 - (void)clearDevices {
   @synchronized(self) {
+      
     // Remember the information about the current device.
     MSDeviceHistoryInfo *currentDeviceHistory = [[MSDeviceHistoryInfo alloc] initWithTimestamp:[NSDate date]
                                                                                      andDevice:self.device];
@@ -259,8 +260,10 @@ static MSDeviceTracker *sharedInstance = nil;
     [MS_USER_DEFAULTS setObject:[NSKeyedArchiver archivedDataWithRootObject:deviceHistory] forKey:kMSPastDevicesKey];
 
     // Clear cache.
-    self.device = nil;
     [self.deviceHistory removeAllObjects];
+
+    // Keep the latest information about the device in memory.
+    [self.deviceHistory addObject:currentDeviceHistory];
   }
 }
 
--- a/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
+++ b/AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
@@ -250,9 +250,13 @@ static MSDeviceTracker *sharedInstance = nil;
 
 - (void)clearDevices {
   @synchronized(self) {
+    // Remember the information about the current device.
+    MSDeviceHistoryInfo *currentDeviceHistory = [[MSDeviceHistoryInfo alloc] initWithTimestamp:[NSDate date]
+                                                                                     andDevice:self.device];
+    NSArray* deviceHistory = [NSArray arrayWithObject:currentDeviceHistory];
 
-    // Clear persistence.
-    [MS_USER_DEFAULTS removeObjectForKey:kMSPastDevicesKey];
+    // Clear persistence, but keep the latest information about the device.
+    [MS_USER_DEFAULTS setObject:[NSKeyedArchiver archivedDataWithRootObject:deviceHistory] forKey:kMSPastDevicesKey];
 
     // Clear cache.
     self.device = nil;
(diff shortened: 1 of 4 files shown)


AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
```diff
@@ -250,9 +250,13 @@ - (MSDevice *)deviceForTimestamp:(NSDate *)timestamp {
 
 - (void)clearDevices {
   @synchronized(self) {
+    // Remember the information about the current device.
```
russelarms: add a new line before the comment

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### App Center Crashes
 
 * **[Fix]** Fix sending crashes if an application is launched in background.
+* **[Fix]** Fix sending device info is attached to crashes.
```
MatkovIvan: Add more details here

AppCenter/AppCenterTests/MSDeviceTrackerTests.m
```diff
@@ -472,14 +472,12 @@ - (void)testClearingDeviceHistoryWorks {
   MSMockUserDefaults *defaults = [MSMockUserDefaults new];
 
   // When
+  // If the storage is empty, remember the current device.
+  [self.sut device];
   [self.sut clearDevices];
 
   // Then
   XCTAssertTrue([self.sut.deviceHistory count] == 0);
```
MatkovIvan: Why is it 0 if we always keep at least 1 now?

guperrot:
Do we have the same bug on Android?

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
```diff
@@ -250,13 +250,20 @@ - (MSDevice *)deviceForTimestamp:(NSDate *)timestamp {
 
 - (void)clearDevices {
   @synchronized(self) {
+      
+    // Remember the information about the current device.
+    MSDeviceHistoryInfo *currentDeviceHistory = [[MSDeviceHistoryInfo alloc] initWithTimestamp:[NSDate date]
```
jaelim-ms: I guess the history already contains current device info. Can we just remove all except the latest one?

MatkovIvan:
> Do we have the same bug on Android?

@guperrot no

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
```diff
@@ -250,13 +250,17 @@ - (MSDevice *)deviceForTimestamp:(NSDate *)timestamp {
 
 - (void)clearDevices {
   @synchronized(self) {
+      
+    // Refresh current device info if needed before archive.
+    [self device];
```
jaelim-ms: Why do you refresh the current device info? I think it is already there when device tracker was started.

ilyashumihin: No, only if the history of the devices was empty, the current device will be added to the history (based on the constructor). Therefore, we need to make sure that the current device is really added to the history.

jaelim-ms: > only if the history of the devices was empty

I think your assumption is wrong here. It will add device info whenever device tracker is started. Can you see the issue without this line?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * **[Fix]** Fix sending crashes if an application is launched in background.
 * **[Fix]** Validate error attachment size to avoid server error or out of memory issues (using the documented limit which is 7MB).
+* **[Fix]** Fix an issue where crash might contain incorrect data if there was a 2nd crash in a row in the previous version of the application.
```
```suggestion
* **[Fix]** Fix an issue where crash might contain incorrect data if two consecutive crashes occurred in a previous version of the application.
```

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
ilyashumihin: I fixed the solution, now when initializing the tracker, we always save the current device.

PR #1891, opened 2019-11-20
Fix/handle keychain errors

achocron:

* [x] Did you add unit tests?

## Description

This PR fixes an issue where, in certain cases, the app does not have access to the update token stored in keychain. This happens when the phone is locked with a passcode but the flow is executing in the background. We've seen it a lot when there are silent pushes.

The side effects of this is that when in app updates fails to retrieve the token but tries to get release details anyways, a 403 is received because the authentication fails without the token. Then, all state is cleared and the user must sign in to app center on relaunch.

This PR fixes that by
1. enabling a mode where the update token *can* be accessed in locked mode (simple keychain flag)
2. there could still be an edge case where it is not accessed so when we see the access error, don't clear state

[AB#72031](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72031)

### DIFF
--- a/AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
+++ b/AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
@@ -105,6 +105,7 @@ static NSString *kMSDefaultServiceName = @"DefaultServiceName";
 - (void)stopMocking {
   [stringsDictionary removeAllObjects];
   [arraysDictionary removeAllObjects];
+  [statusCodes removeAllObjects];
   [self.mockKeychainUtil stopMocking];
 }
 
--- a/AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
+++ b/AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
@@ -35,10 +35,10 @@ static NSString *kMSDefaultServiceName = @"DefaultServiceName";
     OCMStub([_mockKeychainUtil deleteStringForKey:[OCMArg any]]).andCall([self class], @selector(deleteStringForKey:));
     OCMStub([_mockKeychainUtil deleteStringForKey:[OCMArg any] withServiceName:[OCMArg any]])
         .andCall([self class], @selector(deleteStringForKey:withServiceName:));
-    OCMStub([_mockKeychainUtil stringForKey:[OCMArg any] withStatusCode:[OCMArg anyPointer]])
-        .andCall([self class], @selector(stringForKey:withStatusCode:));
-    OCMStub([_mockKeychainUtil stringForKey:[OCMArg any] withServiceName:[OCMArg any] withStatusCode:[OCMArg anyPointer]])
-        .andCall([self class], @selector(stringForKey:withServiceName:withStatusCode:));
+    OCMStub([_mockKeychainUtil stringForKey:[OCMArg any] statusCode:[OCMArg anyPointer]])
+        .andCall([self class], @selector(stringForKey:statusCode:));
+    OCMStub([_mockKeychainUtil stringForKey:[OCMArg any] withServiceName:[OCMArg any] statusCode:[OCMArg anyPointer]])
+        .andCall([self class], @selector(stringForKey:withServiceName:statusCode:));
     OCMStub([_mockKeychainUtil clear]).andCall([self class], @selector(clear));
   }
   return self;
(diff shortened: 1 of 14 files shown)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -239,7 +239,12 @@ - (void)startUpdate {
   NSString *releaseHash = MSPackageHash();
   if (releaseHash) {
     [self changeDistributionGroupIdAfterAppUpdateIfNeeded:releaseHash];
-    NSString *updateToken = [MSKeychainUtil stringForKey:kMSUpdateTokenKey];
+    OSStatus statusCode;
+    NSString *updateToken = [MSKeychainUtil stringForKey:kMSUpdateTokenKey statusCode:&statusCode];
+    if (statusCode == errSecInteractionNotAllowed) {
```
jaelim-ms: Shouldn't you log errors for all failed status codes?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -439,7 +444,6 @@ - (void)checkLatestRelease:(NSString *)updateToken distributionGroupId:(NSString
                     // If the response payload is MSErrorDetails, consider it as a recoverable error.
                     if (!details || ![kMSErrorCodeNoReleasesForUser isEqualToString:details.code]) {
                       [MSKeychainUtil deleteStringForKey:kMSUpdateTokenKey];
-                      [MS_USER_DEFAULTS removeObjectForKey:kMSSDKHasLaunchedWithDistribute];
```
jaelim-ms: Any reason of removing this line?

AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
```diff
@@ -68,14 +71,31 @@ + (NSString *_Nullable)deleteStringForKey:(NSString *)key withServiceName:(NSStr
...
+  if (statusCode) {
+    *statusCode = placeholderStatus;
+  }
+  if (placeholderStatus != noErr) {
+    return nil;
+  }
   return stringsDictionary[serviceName][key];
 }
 
++ (void)mockStatusCode:(OSStatus)statusCode forKey:(NSString *)key {
```
jaelim-ms: Why do you need this one? Can't you just remove all objects in `stopMocking ` method and use it?

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -2324,6 +2324,56 @@ - (void)testShouldChangeDistributionGroupIdIfStoredIdDoesntMatchDownloadedId {
...
+  OCMReject([distributeMock requestInstallInformationWith:OCMOCK_ANY]);
+  OCMReject([distributeMock checkLatestRelease:OCMOCK_ANY distributionGroupId:OCMOCK_ANY releaseHash:OCMOCK_ANY]);
+
+  // When
+  [distributeMock startUpdate];
+
+  // Stop mocking
+  [distributeMock stopMocking];
+  [utilityMock stopMocking];
+  [MSMockKeychainUtil mockStatusCode:noErr forKey:kMSUpdateTokenKey];
```
jaelim-ms: If you use keychain mock, you must call `stopMocking` to clean it up.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
achocron: Keychain util will log.

AppCenter/AppCenterTests/Util/MSMockKeychainUtil.m
achocron: We specifically need to mock certain status codes - this isn't only used for cleanup

achocron: but we can do that. still need to keep the method though.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
achocron: The user will be re-prompted to sign in after an unrecoverable error. This was causing issues but i am not sure if we need that line to be removed now that we fixed the issue at another level.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
achocron: The logic here is more about returning than logging.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
achocron: I can re-add it. It probably isn't needed anymore.

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
achocron: It is called in tearDown

PR #1906, opened 2019-12-06
Fix crash during SFAuthenticationSession processing request.

ilyashumihin:

* [x] Did you add unit tests?

## Description

During the SFAuthenticationSession request, the method accesses the remote controller SFAuthenticationViewController which is in the process of being released.

Also, when during authorization we switch to a third-party application (for example MS Authenticator) our application forcibly closes the browser (when going into the background). So when you return back, we have again a window appears with the authorization.

**_Preparation to reproduce:_**
1. Test on real device (IOS 11 and above)
2. Open browser settings and clear browser cache
3. Disable debug mode on project
4. You need to install and configure the MS Authenticator app

**_Steps to reproduce:_**
1. Start the sasquatch app (If the pop-up does not appear, close the application and open again)
2. On alert "Wants to use 'AppCenter.ms' to Sign in" click continue
3. On SafariViewController click - sign in via "Company or school"
4. Enter your ms account
5. Click sign in with your phone or token device
6. Open "MS Authenticator" application, enter pin code or Touch ID (on the same device)
7. Return to Sasquatch application
8. The browser was closed, authorization was not made.

Tested on - iphone 6s (IOS 13.1), iPhone XS max (IOS 13.2) - caught the bug
Also tested on - iphone 5s (IOS 10.3.3) - work fine.

## Related PRs or issues

[AB#72230](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72230)
[AB#74160](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Doug-Team/Backlog%20Items/?workitem=74160)


AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -577,7 +577,7 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
     if (!strongSelf) {
       return;
     }
-    strongSelf.authenticationSession = nil;
+    [strongSelf clearAuthenticationSession];
```
prkuruva: Could we remove the try-catch in lines 595-602 with this change?

clpolet: Yes it will be removed.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 
 * **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new `UIScene` API (multiple scenes are not yet supported).
 * **[Fix]** Fix an issue where users would sometimes be prompted multiple times to sign in with App Center.
+* **[Fix]** Fix a crash in some case when SFAuthenticationSession accesses the controller which is in the process of being released.
```
MatkovIvan:
```suggestion
* **[Fix]** Fix a crash in some case when `SFAuthenticationSession` accesses the controller which is in the process of being released.
```

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 
 * **[Fix]** Fix missing alert dialogs in apps that use iOS 13's new `UIScene` API (multiple scenes are not yet supported).
 * **[Fix]** Fix an issue where users would sometimes be prompted multiple times to sign in with App Center.
+* **[Fix]** Fix a crash in some case when `SFAuthenticationSession` accesses the controller which is in the process of being released.
```
MatkovIvan: Move it to next version

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: try-catch is removed.

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -1190,6 +1190,37 @@ - (void)testOpenUrlWithUpdateSetupFailure {
...
+    XCTAssertNotNil([distributeMock authenticationSession]);
+
+    // When
+    [notificationCenterMock postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
+
+    OCMVerify([distributeMock clearAuthenticationSession]);
+    XCTAssertNil([distributeMock authenticationSession]);
+
+    // Clear
+    [distributeMock stopMocking];
```
MatkovIvan: Stop `notificationCenterMock` too

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -1190,6 +1190,37 @@ - (void)testOpenUrlWithUpdateSetupFailure {
...
+    // When
+    [distributeMock openURLInAuthenticationSessionWith:fakeURL];
+
+    // Then
+    XCTAssertNotNil([distributeMock authenticationSession]);
+
+    // When
+    [notificationCenterMock postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
+
+    OCMVerify([distributeMock clearAuthenticationSession]);
```
MatkovIvan: You should check `[session cancel]` instead

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -330,6 +330,11 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
  */
 - (void)closeApp;
 
+/**
+ * Clear currently running SFAuthenticationSession.
+ */
+- (void)clearAuthenticationSession;
```
MatkovIvan: No need to expose it to header

CHANGELOG.md
ilyashumihin: Done

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
ilyashumihin: Done

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # App Center SDK for iOS, macOS and tvOS Change Log
 
+## Version 2.5.3  (Under development)
```
MatkovIvan:
```suggestion
## Version 2.5.3 (Under development)
```

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -30,6 +30,11 @@
  */
 static NSString *const kMSGroupId = @"Distribute";
 
+/**
+ *  Background task to save the browser connection.
```
MatkovIvan:
```suggestion
 * Background task to save the browser connection.
```

MatkovIvan:
@ilyashumihin please resolve conflicts

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -584,21 +585,36 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
     if (error.code == SFAuthenticationErrorCanceledLogin) {
       MSLogError([MSDistribute logTag], @"Authentication session was cancelled by user or failed.");
     }
+    MSLogDebug([MSDistribute logTag], @"Using SFAuthenticationSession got data with URL: %@", callbackUrl);
+
```
MatkovIvan: Add a comment before empty line

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -1097,7 +1113,16 @@ - (void)applicationWillEnterForeground {
   }
 }
 
-- (void)applicationDidEnterBackground {
+/**
+* Clear currently running SFAuthenticationSession.
```
MatkovIvan: Format

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * **[Fix]** Disable module debugging for release mode in the SDK to fix dSYM warnings.
 
+### App Center Distribute
+
+* **[Fix]** Fix a crash in some case when `SFAuthenticationSession` accesses the controller which is in the process of being released.
+* **[Fix]** Fix sign-in with authenticator app during updates activation.
```
jaelim-ms: Can you revise this to be more detailed and generic? Is it only caused by Authenticator app or it could be possible to be crashed with others?

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -1350,15 +1349,8 @@ - (void)testOpenURLInAuthenticationSession API_AVAILABLE(ios(11)) {
   XCTAssertNotNil(self.sut.authenticationSession);
   XCTAssert([self.sut.authenticationSession isKindOfClass:[SFAuthenticationSession class]]);
 
-  // When
-  [notificationCenterMock postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
```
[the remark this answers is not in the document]
ilyashumihin: Going into background no longer has an effect on this code. This test checks the operability of the 'openURLInAuthenticationSessionWith' method.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * **[Fix]** Disable module debugging for release mode in the SDK to fix dSYM warnings.
 
+### App Center Distribute
+
+* **[Fix]** Fix a crash in some case when `SFAuthenticationSession` accesses the controller which is in the process of being released.
```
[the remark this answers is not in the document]
ilyashumihin: Updated

CHANGELOG.md
ilyashumihin: Updated

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -587,18 +588,31 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
+
+  /* Request additional background execution time for authorization. If we authorize using third-party services (MS Authenticator)
+   * then switching to another application will kill the current session. This line fixes this problem.
+   */
+  backgroundAuthSessionTask = [[MSUtility sharedApp] beginBackgroundTaskWithName:@"Safari authentication"
+                                                               expirationHandler:^{
+                                                                 [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
+                                                               }];
+  BOOL success = [session start];
+  if (success) {
```
annakocheshkova: else?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -579,7 +580,7 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
     if (!strongSelf) {
       return;
     }
-    strongSelf.authenticationSession = nil;
+    [strongSelf clearAuthenticationSession];
```
clpolet: This method cancels the session but we don't need to cancel it at this point it's completed already right? Why not just breaking the strong reference as before?

clpolet: I think it makes more sense in reading the code if we break the strong reference at the end of the completion handler before `endBackgroundTask` instead of here.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +590,33 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
-    }
-  } @catch (NSException *exception) {
-    MSLogError([MSDistribute logTag], @"Failed to start authentication session: %@", exception.reason);
+
+  /* Request additional background execution time for authorization. If we authorize using third-party services (MS Authenticator)
+   * then switching to another application will kill the current session. This line fixes this problem.
+   */
+  backgroundAuthSessionTask = [[MSUtility sharedApp] beginBackgroundTaskWithName:@"Safari authentication"
+                                                               expirationHandler:^{
+                                                                 [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
```
clpolet: We should cancel the session here instead of calling `endBackgroundTask`. Then, `endBackgroundTask` and `strongSelf.authenticationSession = nil` will happen inside the completion handler of the session. Although I suppose it would need to be called from the main thread, would that cause a problem at this point of execution?
Also, this should probably be the only place where we need to cancel the session.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +590,33 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
+    [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
   };
   SFAuthenticationSession *session = [[SFAuthenticationSession alloc] initWithURL:url
                                                                 callbackURLScheme:callbackUrlScheme
                                                                 completionHandler:authCompletionBlock];
 
+  if (self.authenticationSession != nil) {
+
+    // Calling 'start' on an existing session crashes the application - clear session.
+    [self clearAuthenticationSession];
```
clpolet: You can just call cancel on the current session right?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -1101,7 +1115,16 @@ - (void)applicationWillEnterForeground {
   }
 }
 
-- (void)applicationDidEnterBackground {
+/**
+ * Clear currently running SFAuthenticationSession.
+ */
+- (void)clearAuthenticationSession API_AVAILABLE(ios(11)) {
```
clpolet: If my feedback is right there is no need to have a method to do cancellation and strong reference breaking both at the same time but we can keep the method for isolating the complexity of the iOS version dependent code to cancel only the session.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: In the documentation of the 'cancel' method is written that if SafariViewController was presented, after the call 'cancel' it will be dismiss. At the time of completion handler execution this ViewController may still exist and we forcibly dismiss it before remove the reference.

ilyashumihin: If we only break reference, we can have a chance to get 'Attempting to load the view of a view controller while it is deallocating'. Since the viewController still exists. According to the cancel documentation, it can fix our problem.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: If we retry to assign the value to authSession without removing the reference we will get a fatal error. So we also need to remove the link.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: Yes, I think you're right, I will correct this point.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
MatkovIvan: @clpolet No, it crashes in this case. it was fixed before by cleaning the reference in background event (that is not fully correct).

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
clpolet: I think we still don't need this method since we don't need to explicitly break the strong reference with `nil` instead we just cancel before reassigning the new session.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
clpolet: Actually, we don't need to do any session clean up in the completion handler if we make sure to cancel the old session and reassign to the new session before starting the new session.

clpolet: No need to break the strong reference in the completion handler so resolving this conversation.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
clpolet: _For info at first I thought cancelling the session programmatically would call the completion handler but it doesn't._

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -579,7 +580,6 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
     if (!strongSelf) {
       return;
```
clpolet: We should still end the bg task before returning here.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +589,33 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
   self.authenticationSession = session;
-  @try {
-    BOOL success = [session start];
-    if (success) {
-      MSLogDebug([MSDistribute logTag], @"Authentication session started, showing confirmation dialog.");
-    }
-  } @catch (NSException *exception) {
-    MSLogError([MSDistribute logTag], @"Failed to start authentication session: %@", exception.reason);
+
+  /* Request additional background execution time for authorization. If we authorize using third-party services (MS Authenticator)
```
clpolet:
```suggestion
  /* 
   * Request additional background execution time for authorization. If we authorize using third-party services (MS Authenticator)
```

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: Fixed

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +590,34 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
+    [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
   };
   SFAuthenticationSession *session = [[SFAuthenticationSession alloc] initWithURL:url
                                                                 callbackURLScheme:callbackUrlScheme
                                                                 completionHandler:authCompletionBlock];
 
+  if (self.authenticationSession != nil) {
+
+    // Calling 'start' on an existing session crashes the application - cancel session.
+    [(SFAuthenticationSession *)self.authenticationSession cancel];
```
MatkovIvan: Why do we still use cast instead just storing the correct type?

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +590,34 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
     if (callbackUrl) {
       [strongSelf openURL:callbackUrl];
     }
+    [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
   };
   SFAuthenticationSession *session = [[SFAuthenticationSession alloc] initWithURL:url
                                                                 callbackURLScheme:callbackUrlScheme
                                                                 completionHandler:authCompletionBlock];
 
+  if (self.authenticationSession != nil) {
```
MatkovIvan: It looks like that we can drop this condition because sending message to `nil` is OK in Objective-C

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -589,20 +590,34 @@ - (void)openURLInAuthenticationSessionWith:(NSURL *)url API_AVAILABLE(ios(11)) {
...
+  /*
+   * Request additional background execution time for authorization. If we authorize using third-party services (MS Authenticator)
+   * then switching to another application will kill the current session. This line fixes this problem.
+   */
+  backgroundAuthSessionTask = [[MSUtility sharedApp] beginBackgroundTaskWithName:@"Safari authentication"
+                                                               expirationHandler:^{
+                                                                 [[MSUtility sharedApp] endBackgroundTask:backgroundAuthSessionTask];
+                                                               }];
+  BOOL success = [session start];
+  if (success) {
```
MatkovIvan:
```suggestion
  if ([session start]) {
```

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
ilyashumihin: SFAuthenticationSession is only used in ios 11 and above

MatkovIvan: Not resolved. The entire method only for iOS 11 (even if not it doesn't change this point)

ilyashumihin: Updated

PR #1928, opened 2019-12-18
Updated testDelegateDeadlock

ilyashumihin:

* [ ] Did you add unit tests?

## Description

@synchronized did not expect the background thread to start

## Related PRs or issues

[AB#74371](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/74371)


AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -543,7 +543,8 @@ - (void)testLogsSentWithSuccess {
 - (void)testDelegateDeadlock {
 
   // If
-  __block NSObject *lock = [NSObject new], *syncCallback = [NSObject new], *syncBackground = [NSObject new];
+  __block NSObject *lock = [NSObject new], *syncCallback = [NSObject new];
+  dispatch_semaphore_t syncBackground = dispatch_semaphore_create(0);
```
MatkovIvan: Please add some comment why we should use it

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -1917,4 +2006,19 @@ - (id)getValidMockLogWithDate:(NSDate *)date {
   return mockLog;
 }
 
+- (dispatch_queue_t)createDispatchQueue {
+  uuid_t uuid;
+  uuid_generate(uuid);
+  char *nameQueue = malloc(100);
```
MatkovIvan: Memory leak

ilyashumihin: Fixed

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -72,17 +67,12 @@ - (void)setUp {
    * dispatch_get_main_queue isn't good option for logsDispatchQueue because
```
MatkovIvan: Update comment

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -72,17 +67,12 @@ - (void)setUp {
...
    * we can't clear pending actions from it after the test. It can cause usages of stopped mocks.
    */
-  self.logsDispatchQueue = dispatch_queue_create("com.microsoft.appcenter.ChannelGroupQueue", DISPATCH_QUEUE_SERIAL);
```
MatkovIvan: `setUp` is called before each test, so the only change is to drop the id. Please revert changes in all tests just because of this

ilyashumihin: The queue ID only displays additional information in the debugger, and it does not affect the test process: https://developer.apple.com/documentation/dispatch/1453030-dispatch_queue_create
Before that, each test depended on the global variable dispatchQueueLogs. if we try to overwrite the value of the queue that started its execution, we run the risk of getting a critical error. Therefore, we wrote initialization of our queue on each test (we caught this situation when debugging).

AppCenter/AppCenterTests/Util/MSDispatchTestUtil.m
```diff
@@ -12,12 +12,13 @@ + (void)awaitAndSuspendDispatchQueue:(dispatch_queue_t)dispatchQueue {
   // Wait for all tasks to complete, then call suspend in the final task.
   dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
   dispatch_async(dispatchQueue, ^{
-    // Suspend the execution of any subsequent tasks.
-    dispatch_suspend(dispatchQueue);
```
jaelim-ms: I'm curious about this change. Can you explain why?

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
jaelim-ms: > we run the risk of getting a critical error
Can you explain?
Is this an intention of dealloc?
I couldn't find any changes that fixes the issue. Please explain.

clpolet: I don't think I understood why we don't create the queue in the setup anymore.

AppCenter/AppCenterTests/Util/MSDispatchTestUtil.m
ilyashumihin: Suspending an executing thread can cause a fatal error. We should suspend it after all tasks in queue are completed.

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
ilyashumihin: All of our tests were dependent on the global queue variable. There are situations when the test was completed, but the queue still continues its execution. In the next test, the startup method changes the value of the current queue. Old queue continues its execution on the sut object, and causes conflicts.

ilyashumihin: In the deadlock test, I changed the recursive mutex to semaphore. Recursive mutex did not wait for the background queue to start executing.

clpolet: Why does old queue continues its execution on the sut object isn't the old queue suspended? In this case are you referring to sut as in the sut of the new test or the sut of the old test?

AppCenter/AppCenterTests/Util/MSDispatchTestUtil.m
jaeklim: Interesting. The Apple doc says
```
The suspension occurs after completion of any blocks running at the time of the call.
```
Sounds like it will suspend a queue after block completion.

ilyashumihin: Yes, you're right, the problem appeared in a slightly different situation, but also because of the suspend. 
We set the global queue variable to delete, and immediately after the test ends, we set a new value for the global queue variable. After that, the system can generate an error:"BUG IN CLIENT OF LIBDISPATCH: Release of a suspended object". But in this case, the permutation of the suspend method from the block does not really matter, I returned back.
But we still need to isolate the current tests from the global queue variable, that was the problem. We ran tests earlier and every 10 times we caught an error. Now we have run 100 tests and everything is successful.

AppCenter/AppCenterTests/Util/MSDispatchTestUtil.m
```diff
@@ -12,6 +12,7 @@ + (void)awaitAndSuspendDispatchQueue:(dispatch_queue_t)dispatchQueue {
   // Wait for all tasks to complete, then call suspend in the final task.
   dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
   dispatch_async(dispatchQueue, ^{
+
```
jaeklim: Unnecessary change even though it has a correct format. The formatter will anyway remove this line in the future.
```suggestion
```

PR #1935, opened 2019-12-24
Update macro

ilyashumihin:

* [x] Did you add unit tests?

## Description

ARM64 has a distinctive implementation for sending messages, so the last preprocessor crashed the execution. Also arm64 otherwise converts a variadic function.
The solution is rewritten on 2 different preprocessors, as the conversion of primitive types differs from the conversion of object types.

## Related PRs or issues

#AB74394


MatkovIvan:
Temporary changed target branch to get clear diif

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -5,6 +5,7 @@
 #import <UIKit/UIWindow.h>
 
 #import "MSAlertController.h"
+#import "MSPerformSelectorUtil.h"
```
MatkovIvan: It's Vendor folder, it shouldn't depend on SDK utils.
I suggest just move this `MSAlertController` to core module, because it's our helper, not 3rd-party thing

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
   }
 }
 
-// FIXME: This macro is not safe for nil arguments as types in this case cannot be inferred.
+#define ARRAY_FROM_ARGS(...) ({\
+ [NSArray arrayWithObjects: !(sizeof( (char[]){#__VA_ARGS__} ) == 1) ? __VA_ARGS__ : [NSNull null], nil];\
```
MatkovIvan: I don't think that `sizeof` will work correctly here. Also, we need some tests for it

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.m
```diff
@@ -27,4 +27,22 @@ + (void)performSelectorOnMainThread:(NSObject *)source withSelector:(SEL)selecto
   [invocation performSelectorOnMainThread:@selector(invoke) withObject:nil waitUntilDone:NO];
 }
 
++ (NSInvocation *)invoke:(NSObject *)source withSelector:(NSString *)selector withObjects:(NSArray *)objects {
+  SEL selectors = NSSelectorFromString(selector);
```
MatkovIvan: Can we share some code with function above?

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+})
+
+#define MS_DISPATCH_SELECTOR_OBJECT(result, class, selectorName, ...) ({ \
+ NSArray* array = ARRAY_FROM_ARGS(__VA_ARGS__);\
+ id instance = class;\
```
MatkovIvan: Please avoid built-in words usage

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+ NSInvocation *impl = [MSPerformSelectorUtil invoke:instance withSelector:@#selectorName withObjects:array];\
```
MatkovIvan: why `impl`?

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
-  SEL selector = NSSelectorFromString(@#selectorName); \
-  IMP impl = [class methodForSelector:selector]; \
-  ((result (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
+ NSArray* array = ARRAY_FROM_ARGS(__VA_ARGS__);\
+ id instance = class;\
+ NSInvocation *impl = [MSPerformSelectorUtil invoke:instance withSelector:@#selectorName withObjects:array];\
+ void *results = nil;\
+ if(![@#result isEqualToString:@"void"]) {\
+  NSUInteger length = [[impl methodSignature] methodReturnLength];\
+  results = malloc(length);\
```
MatkovIvan: Memory leak

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+  [impl getReturnValue:&results];\
+ } \
+ (result)results; \
```
MatkovIvan: Consider to rename type to highlight that it's a type

Vendor/iOS/MSAlertController/MSAlertController.m
```diff
@@ -146,18 +147,37 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
+
 #define MS_DISPATCH_SELECTOR(result, class, selectorName, ...) ({ \
-  SEL selector = NSSelectorFromString(@#selectorName); \
-  IMP impl = [class methodForSelector:selector]; \
-  ((result (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
+ NSArray* array = ARRAY_FROM_ARGS(__VA_ARGS__);\
+ id instance = class;\
+ NSInvocation *impl = [MSPerformSelectorUtil invoke:instance withSelector:@#selectorName withObjects:array];\
+ void *results = nil;\
+ if(![@#result isEqualToString:@"void"]) {\
```
MatkovIvan: Do it at compile time

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Renamed

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Updated, tried to remove as much as possible

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Fixed

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Moved

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Fixed

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Renamed

Vendor/iOS/MSAlertController/MSAlertController.m
ilyashumihin: Updated the implementation

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.m
ilyashumihin: I moved this method to the preprocessor. I found only such a solution.
We can't use the id or NSObject type for source, so how can we apply invoke on a primitive type. At the same time, we need to return the result of the expression if the returned type is not void. The standard objective-c solution can't do this, so I left on the preprocessor.

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -3,9 +3,70 @@
...
+#define CHECK(...) CHECK_N(__VA_ARGS__, 0, )
+#define PROBE(x) x, 1,
+
+#define IS_PAREN(x) CHECK(IS_PAREN_PROBE x)
+#define IS_PAREN_PROBE(...) PROBE(~)
+
+#define PRIMITIVE_COMPARE(x, y) IS_PAREN(COMPARE_##x(COMPARE_##y)(()))
+
+#define COMPARE_void(x) x
+#define COMPARE_NOT_USABLE(x) x
```
MatkovIvan: Let's just use two separated ones to avoid over-engineering above

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -3,9 +3,70 @@
...
+
+#define EXECUTE_INVOCATION(invoke) ({ [invoke getReturnValue:&results]; })
+
+#define Invocation(type, object, selectorName, objects)                                                                                    \
```
MatkovIvan: Upper case for macros

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -3,9 +3,70 @@
 
 #import <Foundation/Foundation.h>
 
+#define ARRAY_FROM_ARGS(...)                                                                                                               \
+  ({                                                                                                                                       \
+    NSMutableArray *initedArray = [NSMutableArray arrayWithObjects:[NSNull null], ##__VA_ARGS__, nil];                                     \
+    [initedArray removeObjectAtIndex:0];                                                                                                   \
+    initedArray;                                                                                                                           \
+  })
```
MatkovIvan: I guess it won't work. I don't see `\` at the end of lines. Did you tested it?

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -3,9 +3,70 @@
...
+
+#define IS_PAREN(x) CHECK(IS_PAREN_PROBE x)
+#define IS_PAREN_PROBE(...) PROBE(~)
+
+#define PRIMITIVE_COMPARE(x, y) IS_PAREN(COMPARE_##x(COMPARE_##y)(()))
+
+#define COMPARE_void(x) x
+#define COMPARE_NOT_USABLE(x) x
+
+#define EXECUTE_INVOCATION(invoke) ({ [invoke getReturnValue:&results]; })
```
MatkovIvan: Why do you need all these brackets?

SasquatchTV/SasquatchTVObjC/MSAlertController.h
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MatkovIvan: Remove copy

SasquatchTV/SasquatchTVObjC/MSAlertController.m
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MatkovIvan: Remove copy

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
ilyashumihin: clang-format-changed-files moved this to the rightmost edge of the line, it's work

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,34 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #import <Foundation/Foundation.h>
 
+#define ARRAY_FROM_ARGS(...)                                                                                                               \
```
MatkovIvan: Add MS prefix

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,34 @@
...
+  ({                                                                                                                                       \
+    NSMutableArray *initedArray = [NSMutableArray arrayWithObjects:[NSNull null], ##__VA_ARGS__, nil];                                     \
+    [initedArray removeObjectAtIndex:0];                                                                                                   \
+    initedArray;                                                                                                                           \
+  })
+
+#define MS_DISPATCH_SELECTOR_OBJECT(type, object, selectorName, ...)                                                                       \
+  ({                                                                                                                                       \
+    void *results;                                                                                                                         \
```
MatkovIvan:
```suggestion
    void *results = nil;                                                                                                                   \
```

SasquatchTV/SasquatchTVObjC/MSAlertController.m
```diff
@@ -146,18 +146,11 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
   }
```
MatkovIvan: Remove this copy too

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,34 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #import <Foundation/Foundation.h>
 
+#define ARRAY_FROM_ARGS(...)                                                                                                               \
+  ({                                                                                                                                       \
+    NSMutableArray *initedArray = [NSMutableArray arrayWithObjects:[NSNull null], ##__VA_ARGS__, nil];                                     \
+    [initedArray removeObjectAtIndex:0];                                                                                                   \
```
MatkovIvan: Why do you need to remove it just after adding?

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,34 @@
...
+    initedArray;                                                                                                                           \
```
MatkovIvan: What about just
```objc
@[##__VA_ARGS__]
```

MatkovIvan: Well, we can just remove this define entirely

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,34 @@
...
+  })
+
+#define MS_DISPATCH_SELECTOR_OBJECT(type, object, selectorName, ...)                                                                       \
+  ({                                                                                                                                       \
+    void *results;                                                                                                                         \
+    [[MSPerformSelectorUtil performSelector:object withSelector:@ #selectorName                                                            \
+                                withObjects:ARRAY_FROM_ARGS(__VA_ARGS__)] getReturnValue:&results];                                        \
```
MatkovIvan:
```suggestion
                                withObjects:@[##__VA_ARGS__]] getReturnValue:&results];                                                    \
```

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.m
MatkovIvan: Is it still true?

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -1,11 +1,25 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
MatkovIvan: Restore

SasquatchTV/SasquatchTVObjC/MSAlertController.h
MatkovIvan: Not resolved

AppCenter/AppCenter/Internals/Util/MSAlertController.m
```diff
@@ -0,0 +1,171 @@
...
+  return [self actionWithTitle:title style:UIAlertActionStyleDestructive handler:handler];
+}
+
+@end
+
+@interface MSAlertController ()
+
+@end
+
+@implementation MSAlertController
```
MatkovIvan: Add some tests for this

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.h
```diff
@@ -3,9 +3,24 @@
 
 #import <Foundation/Foundation.h>
 
+#define MS_DISPATCH_SELECTOR_OBJECT(type, object, selectorName, ...)                                                                       \
+  ({                                                                                                                                       \
+    void *results;                                                                                                                         \
+    [[MSPerformSelectorUtil performSelector:object withSelector:@ #selectorName withObjects:@[ __VA_ARGS__ ]] getReturnValue:&results];    \
+    (__bridge type) results;                                                                                                               \
```
MatkovIvan: No need a separate define for just one `__bridge` word. It can be passed to params

ilyashumihin: Fixed

AppCenter/AppCenter/Internals/Util/MSPerformSelectorUtil.m
ilyashumihin: No, implementation has been updated using the helper class MSPerformSelectorUtil with new implementation.

AppCenter/AppCenter/Internals/Util/MSAlertController.m
```diff
@@ -168,11 +162,8 @@ + (void)makeKeyAndVisible {
     if (!windowScene) {
       windowScene = scenes.anyObject;
     }
-    
-    // FIXME: Explicit typing to call `setWindowScene:` as windowScene maybe nil and the macro won't support it.
-    SEL selector = NSSelectorFromString(@"setWindowScene:");
-    IMP impl = [window methodForSelector:selector];
-    ((void (*)(id, SEL, typeof(window))) impl)(window, selector, windowScene);
+
```
MatkovIvan: Remove empty line or add some comment

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -897,6 +897,33 @@ - (void)testObfuscateRedirectUri {
...
+  [array addObject:@"test"];
+
+  // When
+  NSArray *result = MS_DISPATCH_SELECTOR(__bridge NSArray *, array, objectsAtIndexes:, indexSet);
+
+  // Then
+  XCTAssertTrue(result.count == 1);
+}
+
+- (void)testDispatchVoidMacro {
```
MatkovIvan: It's not a macro anymore

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,70 @@
...
+
+@interface MSAlertController (Testing)
+
++ (void)makeKeyAndVisible;
+
+@end
+
+@implementation MSAlertControllerTests
+
+- (void)testMSAlertAddAction {
```
MatkovIvan: Please fix naming and consider to add more tests

AppCenter/AppCenterTests/MSUtilityTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Renamed and added some tests

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+- (void)testAlertAddDefaultAction {
+
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addDefaultActionWithTitle:@"testAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
```
MatkovIvan: XCTAssertEqual

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+- (void)testAlertAddCancelAction {
+
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addCancelActionWithTitle:@"cancelAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
```
MatkovIvan: XCTAssertEqual

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addDefaultActionWithTitle:@"testAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
+  XCTAssertTrue([[controller actions] firstObject].style == UIAlertActionStyleDefault);
```
MatkovIvan: XCTAssertEqual

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addCancelActionWithTitle:@"cancelAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
+  XCTAssertTrue([[controller actions] firstObject].style == UIAlertActionStyleCancel);
+  XCTAssertTrue([[[controller actions] firstObject].title isEqual:@"cancelAction"]);
```
MatkovIvan: XCTAssertEqualObjects

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addCancelActionWithTitle:@"cancelAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
+  XCTAssertTrue([[controller actions] firstObject].style == UIAlertActionStyleCancel);
```
MatkovIvan: XCTAssertEqual

AppCenter/AppCenterTests/MSAlertControllerTests.m
```diff
@@ -0,0 +1,151 @@
...
+  // If
+  MSAlertController *controller = [MSAlertController new];
+
+  // When
+  [controller addDefaultActionWithTitle:@"testAction" handler:nil];
+
+  // Then
+  XCTAssertTrue([controller actions].count == 1);
+  XCTAssertTrue([[controller actions] firstObject].style == UIAlertActionStyleDefault);
+  XCTAssertTrue([[[controller actions] firstObject].title isEqual:@"testAction"]);
```
MatkovIvan: XCTAssertEqualObjects

SasquatchTV/SasquatchTVObjC/AppDelegate.m
```diff
@@ -58,30 +57,33 @@ - (void)crashes {
...
-                                                                                                @"report so we can fix the issue?"];
+                                                                                                @"report so we can fix the issue?"
+                                                                                 preferredStyle:UIAlertControllerStyleAlert];
 
                // Add a "No"-Button and call the notifyWithUserConfirmation-callback with MSUserConfirmationDontSend.
-               [alertController addCancelActionWithTitle:@"Don't Send"
-                                                 handler:^(UIAlertAction *action) {
-                                                   [MSCrashes notifyWithUserConfirmation:MSUserConfirmationDontSend];
-                                                 }];
+               [alertController addAction:[UIAlertAction actionWithTitle:@"Don't Send"
```
jwallra: Run formatter - `:`s should align, there shouldn't be 2 parameters on the same line.  Hopefully the formatter will fix it automatically 🤷‍♂

MatkovIvan: @jwallra it's not 2 parameters of the same function.

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

AppCenter/AppCenterTests/MSAlertControllerTests.m
ilyashumihin: Updated

SasquatchTV/SasquatchTVObjC/AppDelegate.m
ilyashumihin: I also ran a script for formatting.

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -230,6 +230,12 @@
 		6E48A5A41D3831FE006E8B5F /* MSChannelUnitConfigurationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6E48A5A31D3831FE006E8B5F /* MSChannelUnitConfigurationTests.m */; };
 		6E48A5A71D383893006E8B5F /* MSChannelGroupDefaultTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6E48A5A61D383893006E8B5F /* MSChannelGroupDefaultTests.m */; };
 		6EB1F40E1D2443B7005F9F99 /* MSChannelUnitDefaultTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6EB1F40D1D2443B7005F9F99 /* MSChannelUnitDefaultTests.m */; };
+		7839727423B39C2A00620142 /* MSAlertController.h in Headers */ = {isa = PBXBuildFile; fileRef = 78686FAB23B390C500CDA4C1 /* MSAlertController.h */; };
```
clpolet: Why did this move? Isn't it only used in Distribute?

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
```diff
@@ -3,8 +3,17 @@
 
 #import <Foundation/Foundation.h>
 
+#define MS_DISPATCH_SELECTOR(type, object, selectorName, ...)                                                                              \
```
clpolet: Actually we don't really dispatch here as it's not on the queue/threading context should it really be part of this util?

AppCenter/AppCenter.xcodeproj/project.pbxproj
jaeklim: I guess it is because alert controller is a kind of utility that could be used for other modules even though it is only used by Distribute at the moment. I don't have any strong opinion on this one as it could stay in Distribute or it could be moved to core for any of upcoming usages. The only concern is that I couldn't think of any future usage of this one. 😅

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
MatkovIvan: the code of it almost does the same thing. So it looks ok to have it together. Probably better naming helps here...

AppCenter/AppCenter.xcodeproj/project.pbxproj
ilyashumihin: I assume this implementation should be in Core of AppCenter. Possible we may need to display the alert in other modules. And this implementation probably does not have an unambiguous relationship to the distribute module.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
ilyashumihin: I think we can keep this method here. Its essentially dispatch receiver’s message to target.

AppCenter/AppCenter.xcodeproj/project.pbxproj
clpolet: It's just I do not want to bloat core size for something only used by one module unless there is some technical/logical good reason. I don't see one at this time.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
clpolet: I think It's more about messaging not dispatching.

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -230,6 +230,12 @@
...
+		7839727623B39C2B00620142 /* MSAlertController.h in Headers */ = {isa = PBXBuildFile; fileRef = 78686FAB23B390C500CDA4C1 /* MSAlertController.h */; };
```
clpolet: On another topic the alert logic has always been in a "vendor" folder. I am not sure why we did this in the past, I don't strongly about this but I want to make sure we are not missing something by moving it. Basically just raising awareness.

MatkovIvan: (moved because we need some SDK utils as dependency there)

AppCenter/AppCenter.xcodeproj/project.pbxproj
ilyashumihin: I moved this controller to Distribute

AppCenterDistribute/AppCenterDistribute/MSAlertController.m
```diff
@@ -168,11 +162,7 @@ + (void)makeKeyAndVisible {
     if (!windowScene) {
       windowScene = scenes.anyObject;
     }
-    
-    // FIXME: Explicit typing to call `setWindowScene:` as windowScene maybe nil and the macro won't support it.
-    SEL selector = NSSelectorFromString(@"setWindowScene:");
-    IMP impl = [window methodForSelector:selector];
-    ((void (*)(id, SEL, typeof(window))) impl)(window, selector, windowScene);
+    [MSDispatcherUtil performSelector:window withSelector:@"setWindowScene:" withObjects:@[ windowScene ]];
```
clpolet: Why are we not using `MS_DISPATCH_SELECTOR` here? Also in some weird scenario `windowScene` may be nil which would cause a crash here. I think the `FIXME` is remains unaddressed at this time.

ilyashumihin: MS_DISPATCH_SELECTOR macro cannot accept the returned void type, so performSelector is used. We tried to make a universal macro, but it turns out to be quite large. Since it is practically not used anywhere else, we decided to simplify the implementation
Why should setWindowScene cause a crash? According to the documentation, it accepts nullable as a parameter. I checked locally and there were no errors.

MatkovIvan: @clpolet It's what it was https://github.com/microsoft/appcenter-sdk-apple/pull/1935/commits/72d5446909be46c9088a5e1b15cbda3c3d5631e3#diff-eaebbbd7ad1e6d97b1a06ddd0cbd3677L13

clpolet: @ilyashumihin and @MatkovIvan OK yes the universal macro looked quite big 😅. I haven't tried but would it be possible to split into too macros one with return value and another one with explicit void return value. I am trying to find some coherency.

@ilyashumihin I am not worried about `setWindowScene:` itself but how we call it with possible nil parameters. If `windowScene` is nil this line crashes because it creates an array containing a nil object.

AppCenter/AppCenter.xcodeproj/project.pbxproj
```diff
@@ -230,6 +230,7 @@
 		6E48A5A41D3831FE006E8B5F /* MSChannelUnitConfigurationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6E48A5A31D3831FE006E8B5F /* MSChannelUnitConfigurationTests.m */; };
 		6E48A5A71D383893006E8B5F /* MSChannelGroupDefaultTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6E48A5A61D383893006E8B5F /* MSChannelGroupDefaultTests.m */; };
 		6EB1F40E1D2443B7005F9F99 /* MSChannelUnitDefaultTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6EB1F40D1D2443B7005F9F99 /* MSChannelUnitDefaultTests.m */; };
+		78686FAE23B390C600CDA4C1 /* MSAlertController.m in Sources */ = {isa = PBXBuildFile; fileRef = 78686FAC23B390C600CDA4C1 /* MSAlertController.m */; };
```
MatkovIvan: It is not in core now

AppCenterDistribute/AppCenterDistribute/MSAlertController.m
```diff
@@ -146,18 +147,11 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
-  IMP impl = [class methodForSelector:selector]; \
-  ((result (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
-})
-
 + (void)makeKeyAndVisible {
   if (@available(iOS 13.0, tvOS 13.0, *)) {
-    UIApplication *application = MS_DISPATCH_SELECTOR(UIApplication *, [UIApplication class], sharedApplication);
-    NSSet *scenes = MS_DISPATCH_SELECTOR(NSSet *, application, connectedScenes);
-    id windowScene = nil;
+    UIApplication *application = MS_DISPATCH_SELECTOR(__bridge UIApplication *, [UIApplication class], sharedApplication);
```
MatkovIvan: Please add a comment why `__bridge` is used here

AppCenterDistribute/AppCenterDistribute/MSAlertController.m
ilyashumihin: @clpolet I made changes. Now we have 2 separate macros and the app doesn't crash when sending nil as an parameter.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
```diff
@@ -3,8 +3,28 @@
 
 #import <Foundation/Foundation.h>
 
+#define ARRAY_FROM_ARGS(...)                                                                                                               \
+  ({                                                                                                                                       \
+    NSMutableArray *initedArray = [NSMutableArray arrayWithObjects:[NSNull null], ##__VA_ARGS__, nil];                                     \
```
clpolet: A problem here is if you pass a nil value in the middle of the arguments, it'll ignore any other arguments. For instance `NSArray *array = ARRAY_FROM_ARGS(@1, nil, @2);` here `@2` will be ignored. Although not sure how to fix it 🤔.

ilyashumihin: I think we can't do this with the usual C and Objective-C tools. We can't cycle through parameters using a variadic func and change them. We can try to impose restrictions on the function in this case. Pass an NSNull value instead of nil. In this case we can find out what positions NSNull is in and convert them to nil.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
clpolet: Well you are right, the doc for `NSInvocation` also mention dispatching.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
clpolet: Wait, isn't this solving our issues here:
```
#define MS_DISPATCH_SELECTOR(declaration, object, selectorName, ...)({ \
(declaration [object methodForSelector:NSSelectorFromString(selectorName)])(object, NSSelectorFromString(selectorName), ##__VA_ARGS__); \
})
```
What was problematic was guessing the types of the parameters and returned value but we can actually ask for them. This is how it would work for instance:
`    MS_DISPATCH_SELECTOR((void (*)(id, SEL, typeof(windowScene))), window, @"setWindowScene:", windowScene);`
Just a one liner.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
```diff
@@ -3,6 +3,12 @@
 
 #import <Foundation/Foundation.h>
 
+#define MS_DISPATCH_SELECTOR(declaration, object, selectorName, ...)                                                                       \
+  ({                                                                                                                                       \
+    (declaration [object methodForSelector:NSSelectorFromString(selectorName)])(object, NSSelectorFromString(selectorName),                \
```
MatkovIvan: Please make it multiline like in original macro

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -897,6 +897,19 @@ - (void)testObfuscateRedirectUri {
   XCTAssertFalse([obfuscatedString rangeOfString:kMSRedirectUriObfuscatedTemplate].location == NSNotFound);
 }
 
+- (void)testDispatchObjectMacro {
```
clpolet: Please also add a test where `nil` is passed as parameter but is not the last parameter.

AppCenter/AppCenter/Internals/Util/MSDispatcherUtil.h
clpolet: Yes, sorry @ilyashumihin my example was really minimal.

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -897,6 +897,28 @@ - (void)testObfuscateRedirectUri {
...
+  MS_DISPATCH_SELECTOR((void (*)(id, SEL, id)), array, addObject:, @"test");
+
+  // Then
+  XCTAssertEqual([array count], 1);
+  XCTAssertEqual([array firstObject], @"test");
+}
+
+- (void)testDispatchObjectMacroWithNil {
+  typedef void (^block)(NSString *, NSString *);
+  MS_DISPATCH_SELECTOR((void (*)(id, SEL, NSString *, NSString *, block)), self, methodWithArgs:secondArg:completionHandler:, nil, @"test",
```
ilyashumihin: Formatted with a script.

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -897,6 +897,28 @@ - (void)testObfuscateRedirectUri {
...
+  // When
+  MS_DISPATCH_SELECTOR((void (*)(id, SEL, id)), array, addObject:, @"test");
+
+  // Then
+  XCTAssertEqual([array count], 1);
+  XCTAssertEqual([array firstObject], @"test");
+}
+
+- (void)testDispatchObjectMacroWithNil {
+  typedef void (^block)(NSString *, NSString *);
```
clpolet: Smart test case 👍

clpolet: We could also have used a mock to validate the call with specific arguments but this works too.

AppCenter/AppCenterTests/MSUtilityTests.m
```diff
@@ -897,6 +897,28 @@ - (void)testObfuscateRedirectUri {
...
+  MS_DISPATCH_SELECTOR((void (*)(id, SEL, NSString *, NSString *, block)), self, methodWithArgs:secondArg:completionHandler:, nil, @"test",
+                       ^(NSString *firstArg, NSString *secondArg) {
```
clpolet: 🤔 I think the test will succeed if the completion is not called. OK now I think we should use a mock.

AppCenterDistribute/AppCenterDistribute/MSAlertController.m
```diff
@@ -146,20 +147,14 @@ + (void)presentNextAlertAnimated:(BOOL)animated {
...
-  IMP impl = [class methodForSelector:selector]; \
-  ((result (*)(id, SEL, ...)) impl)(class, selector, ##__VA_ARGS__); \
-})
-
 + (void)makeKeyAndVisible {
   if (@available(iOS 13.0, tvOS 13.0, *)) {
-    UIApplication *application = MS_DISPATCH_SELECTOR(UIApplication *, [UIApplication class], sharedApplication);
-    NSSet *scenes = MS_DISPATCH_SELECTOR(NSSet *, application, connectedScenes);
-    id windowScene = nil;
+
```
clpolet:
```suggestion
```

AppCenter/AppCenterTests/MSUtilityTests.m
clpolet: An expectation also work yes.

PR #1950, opened 2020-01-27
Update CHANGELOG

annakocheshkova:

[AB#75278](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75278)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 ### App Center Distribute
 
-* **[Breaking change]** Add `setUpdateTrack` method to be able to explicitly set either **public** or **private** update track. By default, a public distribution group will be used.
+* **[Breaking change]** Add `setUpdateTrack` method to be able to explicitly set either `MSUpdateTrackPrivate` or `MSUpdateTrackPublic` update track. By default, a public distribution group will be used.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 ### App Center Distribute
 
+* **[Breaking change]** Add `setUpdateTrack` method to be able to explicitly set either `MSUpdateTrackPrivate` or `MSUpdateTrackPublic` update track. By default, a public distribution group will be used.
```
dhei: Getter method is also provided, we need to mention it.

clpolet: Should also mention that we don't support multiple public groups anymore.

guperrot:
```suggestion
* **[Feature]** Add `updateTrack` property to be able to explicitly set either `MSUpdateTrackPrivate` or `MSUpdateTrackPublic` update track. By default, a public distribution group is used. **Breaking change**: To allow users to access releases of private groups you now need to migrate your application to call `Distribute.updateTrack = MSUpdateTrackPrivate`. Please read the documentation for more details.
* **[Behavior change]** The public distribution is simplified to provide only one public group. If you have existing public groups defined for your application your users will receive the latest version of all public groups.
```

Issue #1944, opened 2020-01-21
Some exceptions thrown on main thread are swallowed by AppKit


depth42:
With appcenter v2.5.3, getting information for exceptions on the main thread has been greatly improved. Thanks for this! 
But we are still seing some exception crashes for which we do not get exception backtraces, like this one: 

```objc
AppKit              -[NSApplication _crashOnException:]
AppKit              +[CATransaction(NSCATransaction) NS_setFlushesWithDisplayLink]
AppKit              ___NSRunLoopObserverCreateWithHandler_block_invoke
CoreFoundation      __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__
CoreFoundation      __CFRunLoopDoObservers
CoreFoundation      __CFRunLoopRun
CoreFoundation      CFRunLoopRunSpecific
HIToolbox           RunCurrentEventLoopInMode
HIToolbox           ReceiveNextEventCommon
HIToolbox           _BlockUntilNextEventMatchingListInModeWithFilter
AppKit              _DPSNextEvent
AppKit              -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:]
AppKit              -[NSApplication run]
AppKit              NSApplicationMain
Merlin Project 6    main main.m:17
libdyld.dylib       start
```

Is there any chance to also catch these?

annakocheshkova:
Hi there! Currently, we report all available information. 
Could you please confirm what information exactly is lost and how to reproduce this kind of behavior? There's not much we can do having so little info.

depth42:
Sadly, we only get these reports from our customers and are currently not able to reproduce these crashes. But I think, I understand, why appcenter currently is not able to gather information, when exceptions occur during +[CATransaction NS_setFlushesWithDisplayLink]. Exceptions which occur there are caught by Apple and forwarded to the private method -[NSApplication _crashOnException:] instead of -[NSApplication reportException:]. Appcenter only seems to swizzle the latter method and therefore misses the first one.

depth42:
I look a bit deeper into this. Could it be, that appcenter is somehow dropping the "Application Specific Backtrace" information from the raw crash logs? This information is added inside -[NSApplication _crashOnException:] by calling NSNoteInCrashReports. There is a similar issue being reported for iOS: https://github.com/microsoft/appcenter/issues/857

MatkovIvan:
[...]

PR #1954, opened 2020-01-28
Add more crash cases

MatkovIvan:
* [ ] Did you add unit tests?

## Description

- Throw Custom C++ exception
- Throw Objective-C exception during drawing custom view

## Related PRs or issues

[AB#75484](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75484)
#1944


CrashLib/CrashLib/MSCrashCXXCustomException.mm
```diff
@@ -0,0 +1,27 @@
...
+- (NSString *)category {
+  return @"Exceptions";
+}
+
+- (NSString *)title {
+  return @"Throw Custom C++ exception";
+}
+
+- (NSString *)desc {
+  return @"Throw an uncaught C++ exception that cannot be casted to std::exception.";
```
owlstronaut:
```suggestion
  return @"Throw an uncaught C++ exception that cannot be cast to std::exception.";
```

CrashLib/CrashLib/MSCrashCustomView.m
```diff
@@ -0,0 +1,57 @@
...
+
+@implementation MSCustomView
+
+#if TARGET_OS_OSX
+-(void)drawRect:(NSRect)rect {
+#else
+-(void)drawRect:(CGRect)rect {
+#endif
+    [super drawRect:rect];
+    @throw [NSException exceptionWithName:NSGenericException reason:@"Crash from drawing custom view."
```
clpolet: This one doesn't crash on iOS and tvOS, if that's expected we should document it.

MatkovIvan: Fixed

PR #1959, opened 2020-01-31
Fixed wrapper sdk info which attached to crash

ilyashumihin:

* [x] Did you add unit tests?

## Description

Fixed a bug where sdkWrapper information was not assigned to crash.

Situation:
When the constructor was called, we initialized the device without wrapperInfo and saved it in history. Next was the installation of wrapperSdk, but the device in history was saved without information about wrapper. Now after each installation of wrapper we replace the current (last saved) device with a device with information about wrapper.

[AB#75474](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Doug-Team/Backlog%20Items/?workitem=75474)

Also screenshot with console log after local test:
Screen Shot 2020-01-31 at 1 05 17 PM


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### App Center Crashes
 
 * **[Fix]** Fix an issue to call Crashes delegate callbacks even though a modal window is being shown on macOS apps.
+* **[Fix]** Fix an issue when React Native SDK did not send wrapperSdk information.
```
guperrot: We should move this to the RN repo changelog. Even if the fix is in native, from customer point of view we fixed RN.

guperrot: Also did we check symbols / stack trace correct after that fix?

ilyashumihin: Added [PR](https://github.com/microsoft/appcenter-sdk-react-native/pull/798) for RN repo

ilyashumihin: Screenshot 2020-02-03 at 15 04 32
JS stack is not displayed on the portal, I think because it is sent in the message field. Created a separate bug on the board.

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
```diff
@@ -428,6 +428,31 @@ - (void)assertIsOtherWithImagePath:(NSString *)imagePath processPath:(NSString *
   XCTAssertEqual(imageType, MSBinaryImageTypeOther, @"Test other image %@ with process %@", imagePath, processPath);
 }
 
+- (void)testErrorLogFromCrashReportWithWrapper {
+
+  // If
+  MSMockUserDefaults *defaults = [MSMockUserDefaults new];
+  [MSDeviceTracker resetSharedInstance];
+  NSData *crashData = [MSCrashesTestUtil dataOfFixtureCrashReportWithFileName:@"live_report_exception"];
+  XCTAssertNotNil(crashData);
```
jwallra: This line should be in the `// Then` block

CHANGELOG.md
guperrot: No that is normal stack trace.

guperrot: The backend will parse the message string actually, that's how it works in RN.

guperrot: how did you prepare symbols?

ilyashumihin: But I don't see this information on the portal:
Screen Shot 2020-02-04 at 4 20 07 PM

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
```diff
@@ -80,6 +80,10 @@ - (void)setWrapperSdk:(MSWrapperSdk *)wrapperSdk {
   @synchronized(self) {
     wrapperSdkInformation = wrapperSdk;
     needRefresh = YES;
+
+    // Replace the last device without wrapperSdk in the UserDefaults with an updated info.
+    [self.deviceHistory removeLastObject];
```
jaeklim: Do you need this line?

CHANGELOG.md
guperrot: its unsymbolicated, how did you prepare and upload symbols for this build?
You need to make sure to include the index.ios.map inside the dsym archive.

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
ilyashumihin: Yes, because at the moment the last device in the history does not contain wrapperSdk. But when calling [self device] we save all items stored in deviceHistory.

CHANGELOG.md
russelarms: I've tested it with proper js source map and it works.
Screenshot 2020-02-05 at 17 25 43

ilyashumihin: Yes, I did send the source map incorrectly...

AppCenter/AppCenter/Internals/Context/Device/MSDeviceTracker.m
jaeklim: What I meant was that you don't need to delete the last one. This will lead to pick a wrong device info for logs that are created before this line instead of device info that doesn't have wrapper SDK info. Both are not good but wrong device info could be worse.

PR #1972, opened 2020-02-12
Implement [MSDistribute checkForUpdate]

thyeggman:

* [x] Did you add unit tests?

## Description

Create a checkForUpdate method to force manual check of new release.

## Misc

[AB#75299](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75299)


AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -37,8 +43,12 @@ - (void)testCheckForUpdateAuthenticateWhenItHasNotAuthenticated {
   MSDistribute *distribute = [MSDistribute new];
   id distributeMock = OCMPartialMock(distribute);
   OCMStub([distributeMock canBeUsed]).andReturn(YES);
+  id parserMock = OCMClassMock([MSBasicMachOParser class]);
```
owlstronaut: does this need to stopMocking?

guperrot: we stopMocking for this in other tests indeed

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -162,6 +162,11 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
  */
 @property(nonatomic) MSDistributeFlags distributeFlags;
 
+/**
+ * Flag to indicate whether an update was started with a checkForUpdate call.
+ */
+@property(atomic) bool checkForUpdateFlag;
```
jaeklim: We can get rid of this property and add a parameter to `startUpdate` for a flag that inforce update flow even though automatic check is disabled. We don't care authentication stuff at the moment so we can simplify this one.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -187,17 +187,6 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
...
 - (void)disableAutomaticCheckForUpdate;
 
-/**
- * Check the latest release of the application.
- *
- * @param updateToken The update token stored in keychain. This value can be nil if it is public distribution.
- * @param distributionGroupId The distribution group Id in keychain.
- * @param releaseHash The release hash of the current version.
- */
-- (void)checkForUpdateWithUpdateToken:(nullable NSString *)updateToken
```
jaeklim: This is a rollback as it was added earlier based on different understanding of the spec.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
```diff
@@ -350,9 +339,11 @@ static NSString *const kMSTesterAppUpdateSetupFailedKey = @"MSTesterAppUpdateSet
 - (void)dismissEmbeddedSafari;
 
 /**
- * Start update workflow
+ * Start update workflow.
+ *
+ * @param onStart A flag that indicates whether it is called when Distribute is started or not.
  */
-- (void)startUpdate;
+- (void)startUpdateOnStart:(BOOL)onStart;
```
jaeklim: We probably want to talk about this method signature even though this is internal use only.

thyeggman: If the `onStart` parameter means what I think it means, the signature should just be `- (void)startUpdate:(BOOL)onStart`...

But `onStart` is supposed to indicate if the call is happening when the SDK is starting, right?  I think something like `isSdkStarting` would be clearer

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -4,111 +4,278 @@
...
+  [MSDistribute disableAutomaticCheckForUpdate];
+
+  // Then
+  XCTAssertFalse(self.sut.automaticCheckForUpdateDisabled);
+}
+
+- (void)testCheckForUpdateOpenBrowserEvenThoughAutomaticUpdateIsDisabled {
+
+  // If
+  id utilityMock = OCMClassMock([MSUtility class]);
```
thyeggman: Utility mock should be shared.  It's used by 4+ tests now and seems to be mostly set up the same way.

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -4,111 +4,278 @@
...
 
+@property(nonatomic) MSDistribute *sut;
+@property(nonatomic) id settingsMock;
+@property(nonatomic) id bundleMock;
+@property(nonatomic) id keychainUtilMock;
+@property(nonatomic) id parserMock;
+
 @end
 
 @implementation MSDistributeCheckForUpdateTests
```
thyeggman: I don't see any tests that call `startUpdateOnStart:YES`, I would think there should be some.  Or do these tests verify its use some other way without calling it directly?

jaeklim: Last two tests cover this scenario as those are testing from start.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
owlstronaut: `updateShouldBeginOnStart:(BOOL)shouldBegin` could work. I think it is fine as is.

owlstronaut:
What's been done here?
image

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -3,112 +3,283 @@
...
+}
+
+- (void)tearDown {
+  [super tearDown];
+  [self.settingsMock stopMocking];
+  [self.keychainUtilMock stopMocking];
+  [self.bundleMock stopMocking];
+  [self.parserMock stopMocking];
+  [self.utilityMock stopMocking];
+  self.sut = nil;
```
MatkovIvan: Add `resetSharedInstance` to `tearDown`

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -3,112 +3,283 @@
...
 
   // If
-  NSString *updateToken = @"UpdateToken";
-  NSString *distributionGroupId = @"DistributionGroupId";
-  NSString *releaseHash = @"ReleaseHash";
-  id distributeMock = OCMPartialMock([MSDistribute new]);
+  NSDictionary<NSString *, id> *plist = @{@"CFBundleShortVersionString" : @"1.0", @"CFBundleVersion" : @"1"};
+  OCMStub([self.bundleMock infoDictionary]).andReturn(plist);
+
+  // Distribute Mock
```
MatkovIvan:
```suggestion
  // Distribute mock.
```

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -3,112 +3,283 @@
...
+  OCMStub([distributeMock canBeUsed]).andReturn(YES);
+  OCMStub([distributeMock isEnabled]).andReturn(YES);
+  [self.settingsMock removeObjectForKey:kMSUpdateTokenRequestIdKey];
+  OCMStub([distributeMock checkForUpdatesAllowed]).andReturn(YES);
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Start update processed"];
+
+  // Ingestion Mock
+  __block id ingestionMock = OCMClassMock([MSDistributeIngestion class]);
+  OCMStub([ingestionMock checkForPublicUpdateWithQueryStrings:OCMOCK_ANY completionHandler:OCMOCK_ANY]).andDo(^(NSInvocation *invocation) {
+    void (^handler)(NSString *callId, NSHTTPURLResponse *_Nullable response, NSData *_Nullable data, NSError *_Nullable error);
```
MatkovIvan: Please add typedef to make it more readable

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
```diff
@@ -3,112 +3,283 @@
...
-  OCMStub([bundleMock infoDictionary]).andReturn(plist);
-  
-  // Parser mock
-  id parserMock = OCMClassMock([MSBasicMachOParser class]);
-  OCMStub([parserMock machOParserForMainBundle]).andReturn(parserMock);
-  OCMStub([parserMock uuid]).andReturn([[NSUUID alloc] initWithUUIDString:@"CD55E7A9-7AD1-4CA6-B722-3D133F487DA9"]);
-  
+  OCMStub([self.bundleMock infoDictionary]).andReturn(plist);
+
+  // Distribute Mock
```
MatkovIvan:
```suggestion
  // Distribute mock.
```

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
jaeklim: We shouldn't rely on clean up that's done by other tests. We should do it in `setUp` wherever it is required.

AppCenterDistribute/AppCenterDistributeTests/MSDistributeCheckForUpdateTests.m
jaeklim: We don't do this in unit tests and just one time use only.

AppCenterDistribute/AppCenterDistribute/MSDistributePrivate.h
thyeggman: I don't have a strong preference, if everything else is good we should merge this to unblock wrapper SDKs

jaeklim: This is still open and we will revisit this soon.

PR #1973, opened 2020-02-13
Don't open browser at start if checkForUpdate is manual

guperrot:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Don't open browser at start if checkForUpdate is manual.

## Related PRs or issues

[AB#76577](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/76577)

## Misc

Will require to interact with [AB#75299](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75299) to include checking for `checkForUpdate` manual flag to open browser, maybe in another PR.

PR #1974, opened 2020-02-13
Fix an incorrect use of OCMReject in an existing test

guperrot:

* [ ] Did you add unit tests?

## Description

Discovered while trying to reuse this test for #1973. OCMReject won't complain if used after the method has been invoked, it needs to be configured before the test code runs. All other tests do that already.

### DIFF
--- a/AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
+++ b/AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
@@ -1534,6 +1534,7 @@ static NSURL *sfURL;
   OCMStub([distributeMock buildTokenRequestURLWithAppSecret:OCMOCK_ANY releaseHash:OCMOCK_ANY isTesterApp:true])
       .andReturn([NSURL URLWithString:@"some_url://"]);
   OCMStub([distributeMock openUrlUsingSharedApp:OCMOCK_ANY]).andReturn(YES);
+  OCMReject([distributeMock openUrlInAuthenticationSessionOrSafari:OCMOCK_ANY]);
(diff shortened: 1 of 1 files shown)


guperrot:
> Please add `OCMVerifyAll` calls.
> 
> See [9.2 Re-throwing fail fast exceptions in verify all](http://ocmock.org/reference/) for details

We verified that in the past and already discussed this in the past.
I verified again yesterday and OCMReject works as expected without this.
Your link is about OCMExpect anyway.

MatkovIvan:
From this link:

> ### Re-throwing fail fast exceptions in verify all
> In fail-fast mode an exception **might not cause** the test to fail. This can happen when the call stack for the method does not end in the test. Fail fast exceptions will be re-thrown when `OCMVerifyAll` is called. This makes it possible to ensure that unwanted invocations from notifications etc. can be detected.

So it **will NOT** fail if it's called from queue for example

guperrot:
> From this link:
> 
> > ### Re-throwing fail fast exceptions in verify all
> > In fail-fast mode an exception **might not cause** the test to fail. This can happen when the call stack for the method does not end in the test. Fail fast exceptions will be re-thrown when `OCMVerifyAll` is called. This makes it possible to ensure that unwanted invocations from notifications etc. can be detected.
> 
> So it **will NOT** fail if it's called from queue for example

But the tests are posting a command to the queue that will be executed after the SDK code, in that queue we fullfill the expectation that is waited. That's why it works I think in the current tests.

MatkovIvan:
I just want to make sure that on possible future edits, it remains valid

PR #1992, opened 2020-02-27
Set userID after start is called

alamathe1:
If setUserID is called before the SDK is started, it does not work

* [ ] Did you add unit tests?

## Description

With this PR of the example app, the objective is to ensure that the setUserId method is called only after the SDK is started

## Related PRs or issues

If setUserId is called before SDK start, it shows up on the AppCenter portal as `N/A`

## Misc


jaeklim:
@alamathe1 Thanks for your contribution and fixing the test app.

PR #2010, opened 2020-03-17
Remove prefix in PLCR usage

MatkovIvan:
* [ ] Did you add unit tests?

## Description

The prefix is handled by PLCR preprocessor definitions. No need to use it explicitly everywhere. The prefix itself is still here, so it's fully binary compatible.

## Related PRs or issues

[AB#77610](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/77610)


jaeklim:
What's the relationship between using the latest PLCR and removing prefix usage in the SDK?
I also doubt if any of libraries (using PLCR) that are also used in our customers' apps don't have this prefix would conflict with AC Crashes as it doesn't set PLCR prefix. Can you explain?

MatkovIvan:
> What's the relationship between using the latest PLCR and removing prefix usage in the SDK?

This is just related improvement. It can be applied to any version that we're using in App Center.

> would conflict with AC Crashes as it doesn't set PLCR prefix. Can you explain?

As mentioned in the description, the prefix is still here, but handled properly by PLCR defines

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -671,10 +671,10 @@ - (void)configureCrashReporterWithUncaughtExceptionHandlerEnabled:(BOOL)enableUn
   }
 #endif
   PLCrashReporterSymbolicationStrategy symbolicationStrategy = PLCrashReporterSymbolicationStrategyNone;
-  MSPLCrashReporterConfig *config = [[MSPLCrashReporterConfig alloc] initWithSignalHandlerType:signalHandlerType
+  PLCrashReporterConfig *config = [[PLCrashReporterConfig alloc] initWithSignalHandlerType:signalHandlerType
                                                                          symbolicationStrategy:symbolicationStrategy
                                                         shouldRegisterUncaughtExceptionHandler:enableUncaughtExceptionHandler];
-  self.plCrashReporter = [[MSPLCrashReporter alloc] initWithConfiguration:config];
+  self.plCrashReporter = [[PLCrashReporter alloc] initWithConfiguration:config];
```
iamclement: I synced with @jaeklim about this, let's have a comment somewhere in the code, possibly above the PLCR init, where we explain the prefixing and why we don't need it in the SDK source code. This will avoid future confusion.

Issue #1635, opened 2019-05-27
Namespace NSUserDefaults keys


sindresorhus:
### **Description**

The SDK uses some NSUserDefault keys to store various information. This is fine, but some of the keys used are not namespaced, which makes it harder to filter them out and you also risk the keys being accidentally overwritten by apps. Since this is an SDK, it's your responsibility to be as resilient as possible.

The following keys and what they should be renamed to:

- `pastDevicesKey` => `MSPastDevicesKey`
- `SessionIdHistory` => `MSSessionIdHistory`
- `UserIdHistory` => `MSUserIdHistory`

Ideally, you should make it even more resilient by using the the prefix `MSAppCenter`, for example, `MSAppCenterPastDevicesKey`.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Add the SDK to an app
2. Run it
3. Then look at the NSUserDefaults

### **Details**

1. Which SDK version are you using?
   2.0.1
2. Which OS version did you experience the issue on?
   macOS 10.14.5
3. Which Cocoapods version are you using (run `pod --version`)?
   None
4. What device version did you see this error on?  Were you using an emulator or a physical device?
   N/A
5. What language are you using?
    - [ ] Objective C
    - [x] Swift
6. What third party libraries are you using?
7. Please enable verbose logging for your app using `MSAppCenter.setLogLevel(.verbose)` before your call to `MSAppCenter.start(...)` for Swift, or `[MSAppCenter setLogLevel:MSLogLevelVerbose]` before `[MSAppCenter start: ...]` for Objective C and include the logs here:

N/A

jayMcBee:
This goes against all guidelines for macOS/iOS - please fix!

MatkovIvan:
Hey!
We've included the fix for this in [the latest SDK release](https://github.com/microsoft/appcenter-sdk-apple/releases/tag/3.2.0).

PR #2013, opened 2020-03-19
Unify NSUserDefaults keys

annakocheshkova:

* [X] Did you add unit tests?

## Description

All the keys used in NSUserDefaults should start with "MS". Not "NS", not "kMS", and also not simply without any prefix.
This PR standartises the keys and adds migration from old keys.

## Related PRs or issues

[AB#73543](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/73543)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 3.1.1 (In development)
 
+### App Center
+
+* **[Improvement]** Use standartised `NSUserDefaults` keys
```
MatkovIvan: Can it be breaking change? Dot at the end, add more details about prefix etc

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
```diff
@@ -18,25 +19,89 @@ + (instancetype)shared {
...
+  NSString *keyWithoutMSPrefixStartingWithLowercase = [firstLetterLowercased stringByAppendingString:[keyWithoutMSPrefix substringFromIndex:1]];
+  [keys addObject:keyWithoutMSPrefixStartingWithLowercase];
+  
+  // kMSKey -> MSKey
+  NSString *keyStartingFromK = [@"k" stringByAppendingString:key];
+  [keys addObject:keyStartingFromK];
+  
+  // NSKey -> MSKey
+  NSString *keyWithNSPrefix = [@"NS" stringByAppendingString:keyWithoutMSPrefix];
+  [keys addObject:keyWithNSPrefix];
```
MatkovIvan: I guess we should use fixed list to avoid affecting non-appcenter data

CHANGELOG.md
annakocheshkova: Why it's a breaking change if we have migration?

annakocheshkova:
Looks like codecov is showing incorrect data

sindresorhus:
Is a `MS` prefix really safe enough? What if Microsoft has other SDKs with the same

sindresorhus:
Is a `MS` prefix really safe enough? What if a different Microsoft project has an SDKs with a `MSUserIdHistory` key? That's a pretty generically named key. Or what if a company named ManagerSoftware uses a `MS` prefix too and uses the AppCenter SDK? Per https://github.com/microsoft/appcenter-sdk-apple/issues/1635, an SDK is supposed to be resilient. I would strongly recommend using a more specific prefix like `MSAppCenter` or `MSAC`.

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.h
```diff
@@ -15,6 +15,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (instancetype)shared;
 
++ (void)resetSharedInstance;
+
+/**
+ * Migrates the given array of keys: re-saves the value by the old key to new key.
+ * Deletes the old key.
+ * @param migratedKeys keys to be migrated.
+ * Dictionary format: [oldKey : newKey].
+ * @param serviceName name of the service executing the migration.
```
MatkovIvan: Not sure that it's correct format

CHANGELOG.md
MatkovIvan: Dot at the end, add more details about prefix etc

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.h
annakocheshkova: @MatkovIvan what place exactly?

iamclement: I think he's referring to line 24.

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
```diff
@@ -1,42 +1,84 @@
...
   });
   return sharedInstance;
 }
 
++ (void)resetSharedInstance {
+  onceToken = 0; // resets the once_token so dispatch_once will run again
+  sharedInstance = nil;
+}
+
+- (void)migrateSettingsKeys:(NSDictionary *)migratedKeys andService:(NSString *)serviceName {
```
iamclement: Can we rename to `migrateKeys:forService:`?

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -276,6 +276,13 @@ + (void)setDelegate:(_Nullable id<MSCrashesDelegate>)delegate {
 
 - (instancetype)init {
   if ((self = [super init])) {
+    NSDictionary *migratedKeys = @{
+      @"kMSCrashesIsEnabledKey" : @"MSACCrashesIsEnabledKey",
+      @"MSAppDidReceiveMemoryWarning" : @"MSACAppDidReceiveMemoryWarning",
+      @"NSApplicationCrashOnExceptions" : @"MSACApplicationCrashOnExceptions",
```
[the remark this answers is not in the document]
annakocheshkova: Oh 😅

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.h
annakocheshkova: Still not sure what's wrong. This is how I pass the keys to the method

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
```diff
@@ -1,42 +1,84 @@
...
 
++ (void)resetSharedInstance {
+  onceToken = 0; // resets the once_token so dispatch_once will run again
+  sharedInstance = nil;
+}
+
+- (void)migrateKeys:(NSDictionary *)migratedKeys forService:(NSString *)serviceName {
+  NSString *serviceHasMigratedKey = [NSString stringWithFormat:kMSAppCenterUserDefaultsMigratedKeyFormat, serviceName];
+  NSNumber *serviceHasMigrated = [self objectForKey:serviceHasMigratedKey];
+  if (serviceHasMigrated) {
```
jaeklim: Let's add a verbose log here that says it has already been migrated.

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,118 @@
...
+#import "MSUserDefaults.h"
+#import "MSWrapperLogger.h"
+
+@interface MSUserDefaultsTests : XCTestCase
+
+@end
+
+static NSString *const kMSAppCenterUserDefaultsMigratedKeyFormat = @"MSACCoreUserDefaultsMigratedKey";
+static NSString *const kMSUserDefaultsPrefix = @"MSAC";
+static NSString *const mockServiceName = @"Core";
```
jaeklim: Is this used?

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,118 @@
...
+  [[NSUserDefaults standardUserDefaults] setObject:@(1) forKey:kMSAppCenterUserDefaultsMigratedKeyFormat];
+
+  // When
+  [MSUserDefaults shared];
+
+  // Then
+  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:@"MSACPastDevicesKey"]);
+}
+
+- (void)testMSIsAppendedOnSetAndGet {
```
jaeklim:
```suggestion
- (void)testPrefixIsAppendedOnSetAndGet {
```

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,118 @@
...
+  [userDefaults setObject:value forKey:key];
+
+  // Then
+  XCTAssertEqual(value, [[NSUserDefaults standardUserDefaults] objectForKey:[kMSUserDefaultsPrefix stringByAppendingString:key]]);
+  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:key]);
+  XCTAssertEqual(value, [userDefaults objectForKey:key]);
+
+  // When
+  [userDefaults removeObjectForKey:key];
+  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:[kMSUserDefaultsPrefix stringByAppendingString:key]]);
```
jaeklim:
```suggestion

  // Then
  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:[kMSUserDefaultsPrefix stringByAppendingString:key]]);
```

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,118 @@
...
+  // If
+  NSDictionary *keys =
+      @{@"okeyTest1" : @"MSACKeyTest1", @"okeyTest2" : @"MSACKeyTest2", @"okeyTest3" : @"MSACKeyTest3", @"okeyTest4" : @"MSACKeyTest4"};
+  MSUserDefaults *userDefaults = [MSUserDefaults shared];
+  NSArray *oldKeysArray = [keys allKeys];
+  NSArray *expectedKeysArray = [keys allValues];
+  for (NSUInteger i = 0; i < [keys count]; i++) {
+    [[NSUserDefaults standardUserDefaults] setObject:[NSString stringWithFormat:@"Test %lu", (unsigned long)i] forKey:oldKeysArray[i]];
+  }
+  // Check that in MSUserDefaultsTest the same keys.
```
jaeklim:
```suggestion

  // Check that in MSUserDefaultsTest the same keys.
```

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,118 @@
...
+- (void)testMigrateUserDefaultSettings {
+
+  // If
+  NSDictionary *keys =
+      @{@"okeyTest1" : @"MSACKeyTest1", @"okeyTest2" : @"MSACKeyTest2", @"okeyTest3" : @"MSACKeyTest3", @"okeyTest4" : @"MSACKeyTest4"};
+  MSUserDefaults *userDefaults = [MSUserDefaults shared];
+  NSArray *oldKeysArray = [keys allKeys];
+  NSArray *expectedKeysArray = [keys allValues];
+  for (NSUInteger i = 0; i < [keys count]; i++) {
+    [[NSUserDefaults standardUserDefaults] setObject:[NSString stringWithFormat:@"Test %lu", (unsigned long)i] forKey:oldKeysArray[i]];
```
jaeklim: Just curiosity, can't it be like below?
```suggestion
    [[NSUserDefaults standardUserDefaults] setObject:[NSString stringWithFormat:@"Test %tu", i] forKey:oldKeysArray[i]];
```

AppCenter/AppCenterTests/Util/MSMockUserDefaults.h
```diff
@@ -9,5 +9,5 @@
  * Clear dictionary
  */
 - (void)stopMocking;
-
+- (void)migrateKeys:( NSDictionary *)migratedKeys forService:( NSString *)serviceName;
```
jaeklim: Please run formatter for this change.
Also add a doc for the method.
```suggestion

- (void)migrateKeys:(NSDictionary *)migratedKeys forService:(NSString *)serviceName;

```

AppCenter/AppCenterTests/Util/MSMockUserDefaults.m
```diff
@@ -26,6 +26,10 @@ - (instancetype)init {
   return self;
 }
 
+- (void)migrateKeys:(__unused NSDictionary *)migratedKeys forService:(__unused NSString *)serviceName {
+  // stub
```
jaeklim: Remove it or revise the comment. Don't forget to add a line break before the comment.

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
```diff
@@ -71,7 +71,7 @@ - (void)setUp {
   [MSAppCenter resetSharedInstance];
 
   // Mock NSUserDefaults.
-  self.settingsMock = [MSMockUserDefaults new];
+  self.settingsMock = OCMPartialMock([MSMockUserDefaults new]);
```
jaeklim: Why do you mock this?? If we want to mock it then we should mock `NSUserDefaults` instead of this mock class. I would suggest you to implement `migrateKeys:forService:` method in `MSMockUserDefaults` class that just set a boolean property to true and assert this value in your test.
I can see that you want to verify migration completed, this will do what you want.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
```diff
@@ -24,7 +24,7 @@ static BOOL isApplicationForwarderEnabled() {
 static NSString *const kMSCrashOnExceptionsKey = @"NSApplicationCrashOnExceptions";
 
 static BOOL isCrashOnExceptionsEnabled() {
-  NSNumber *crashOnExceptions = [MS_USER_DEFAULTS objectForKey:kMSCrashOnExceptionsKey];
+  NSNumber *crashOnExceptions = [[NSUserDefaults standardUserDefaults] objectForKey:kMSCrashOnExceptionsKey];
```
jaeklim: Is this expected change?

AppCenterCrashes/AppCenterCrashesTests/MSCrashesTests.mm
```diff
@@ -69,6 +71,7 @@ @implementation MSCrashesTests
 
 - (void)setUp {
   [super setUp];
+  self.settingsMock = OCMPartialMock([MSMockUserDefaults new]);
```
jaeklim: Same here.

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -101,7 +101,7 @@ - (void)setUp {
   [MSLogger setCurrentLogLevel:MSLogLevelVerbose];
   self.keychainUtilMock = [MSMockKeychainUtil new];
   self.sut = [MSDistribute new];
-  self.settingsMock = [MSMockUserDefaults new];
+  self.settingsMock = OCMPartialMock([MSMockUserDefaults new]);
```
jaeklim: Same here.

AppCenterPush/AppCenterPushTests/MSPushTests.m
```diff
@@ -70,7 +70,7 @@ @implementation MSPushTests
 - (void)setUp {
   [super setUp];
   [MSUserIdContext resetSharedInstance];
-  self.settingsMock = [MSMockUserDefaults new];
+  self.settingsMock = OCMPartialMock([MSMockUserDefaults new]);
```
jaeklim: Same here.

AppCenterPush/AppCenterPushTests/Util/MSMockUserDefaults.m
```diff
@@ -26,6 +26,10 @@ - (instancetype)init {
   return self;
 }
 
+- (void)migrateKeys:(__unused NSDictionary *)migratedKeys forService:(__unused NSString *)serviceName {
+    // stub
```
jaeklim: Remove it or revise the comment. Don't forget to add a line break before the comment.

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
annakocheshkova: Are you sure about it? It means user will get this log 99% of the app starts (1 time it has migrated and all the subsequent starts will contain this log)

AppCenterAnalytics/AppCenterAnalyticsTests/MSAnalyticsTests.m
annakocheshkova: Yeah, good point

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
annakocheshkova: Yes. Our implementation `MSUserDefaults` will automatically append "MSAC" to any given key now (after all, it's our overload of `NSUserDefaults` so this is in fact one of the things it's expected to do). This key here (`NSApplicationCrashOnExceptions`), however, is a system key, so I use system `NSUserDefaults` for it.

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
jaeklim: Oops. I missed L45. It should be good then.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
jaeklim: Got it. I actually don't like to add a prefix in the methods in `MSUserDefaults` but didn't comment that based on the usage in this PR. Now we might want to consider adding `MSAC` in all constants instead of appending it in the methods. Better to have the prefix explicitly than implicitly.

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.m
```diff
@@ -1,42 +1,84 @@
...
 // Licensed under the MIT License.
 
+#import "MSUserDefaults.h"
 #import "MSAppCenterInternal.h"
 #import "MSLogger.h"
-#import "MSUserDefaults.h"
 
 static NSString *const kMSUserDefaultsTs = @"_ts";
+static NSString *const kMSAppCenterUserDefaultsMigratedKeyFormat = @"%@UserDefaultsMigratedKey";
+static NSString *const kMSUserDefaultsPrefix = @"MSAC";
```
iamclement: Should we add a more descriptive prefix such as mentioned in this PR `MSAppCenter`. Might be more helpful debugging user defaults value for the app dev.

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
annakocheshkova: I also thought about it but couldn't come to a conclusion. Let's post a poll

AppCenter/AppCenter/Internals/Model/Util/MSUserDefaults.h
iamclement: Something like that:
```suggestion
 * Deletes the old key.
 *
 * @param migratedKeys keys to be migrated. Dictionary format: [oldKey : newKey].
 * @param serviceName name of the service executing the migration.
```

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
annakocheshkova: Based on the poll, I'm keeping this as is for now. Maybe more people will vote

annakocheshkova:
FYI codecov is still lying 😅

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -276,6 +276,12 @@ + (void)setDelegate:(_Nullable id<MSCrashesDelegate>)delegate {
 
 - (instancetype)init {
   if ((self = [super init])) {
+    NSDictionary *migratedKeys = @{
+      @"kMSCrashesIsEnabledKey" : @"MSACCrashesIsEnabledKey",
+      @"MSAppDidReceiveMemoryWarning" : @"MSACAppDidReceiveMemoryWarning",
+      @"MSUserConfirmation" : @"MSACUserConfirmation"
+    };
+    [MS_USER_DEFAULTS migrateKeys:migratedKeys forService:kMSServiceName];
```
iamclement: 🤔 I am thinking `migrateKeys:forService:` is something the service abstraction can do for all services in its init. If services for instance override a `keysToMigrate` method returning the keys to migrate. Actually another reason to do it there is because we are already getting the isEnabled state for the service from the user defaults in the base class and the migration needs to happen before that.

iamclement: Could be a bit convoluted. If we don't do it this way we still need the migration to happen before accessing the `isEnabled` value.

annakocheshkova: I tried this approach before but got stuck. 
Basically it comes down to this: 
1. We don't know what services we have.
2. It means we should use a delegate approach and this delegate has this `getKeysToMigrateMethod`.
3. Assignment of all delegates should happen before `MSUserDefaults` init (at init we migrate).
4. I added all the delegates at module init but that didn't work: some of the modules are initialised earlier than others in our Sasquatch app for example. 
5. Distribute is called -> it sets the delegate -> it then uses MSUserDefaults in the init which triggers the migration of only one module.

annakocheshkova: Probably we could enumerate by modules array somehow..

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+static MSAppCenterUserDefaults *sharedInstance = nil;
+static dispatch_once_t onceToken;
+
+@implementation MSAppCenterUserDefaults
+
++ (instancetype)shared {
+  dispatch_once(&onceToken, ^{
+    sharedInstance = [[MSAppCenterUserDefaults alloc] init];
+    NSDictionary *migratedKeys = @{
+      @"MSChannelStartTimer" : @"MSACChannelStartTimer",
```
MatkovIvan: Let's add comments where each key is used

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+      @"SessionIdHistory" : @"MSACSessionIdHistory",
+      @"UserIdHistory" : @"MSACUserIdHistory"
+    };
+    [sharedInstance migrateKeys:migratedKeys forService:@"Core"];
+  });
+  return sharedInstance;
+}
+
++ (void)resetSharedInstance {
+  onceToken = 0; // resets the once_token so dispatch_once will run again
```
MatkovIvan:
```suggestion

  // Reset the once_token so dispatch_once will run again.
  onceToken = 0;
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+      [[NSUserDefaults standardUserDefaults] removeObjectForKey:oldKey];
+      MSLogVerbose([MSAppCenter logTag], @"Migrating the key %@ -> %@", oldKey, newKey);
+    }
+  }
+  [self setObject:@(1) forKey:serviceHasMigratedKey];
+}
+
+- (NSString *)getAppCenterKeyFrom:(NSString *)key {
+    if ([key hasPrefix:kMSUserDefaultsPrefix]) {
+        return key;
```
MatkovIvan: Can we add some assert here to catch incorrect usage in debug builds?

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+}
+
+- (NSDictionary *)updateDictionary:(NSDictionary *)dict forKey:(NSString *)key expiration:(float)expiration {
+  NSString *keyPrefixed = [self getAppCenterKeyFrom:key];
+  NSMutableDictionary *update = [[NSMutableDictionary alloc] initWithDictionary:dict];
+
+  // Get from local store.
+  NSDictionary *store = [[NSUserDefaults standardUserDefaults] dictionaryForKey:keyPrefixed];
+
+  CFAbsoluteTime ts = [(NSNumber *)store[kMSUserDefaultsTs] doubleValue];
```
MatkovIvan: What is ts? Use full name and add a comment

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+  MSLogVerbose([MSAppCenter logTag], @"Settings:store[%@]=%@", keyPrefixed, store);
+
+  // Force update if timestamp expiration is reached.
+  if (ts <= 0.0 || expiration <= 0.0f || fabs(CFAbsoluteTimeGetCurrent() - ts) < (double)expiration) {
+
+    // Remove if already in store and value is the same.
+    for (NSString *k in [store allKeys]) {
```
MatkovIvan: k -> key

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+        [update removeObjectForKey:k];
+    }
+  }
+
+  // If still values to update.
+  if ([update count] > 0) {
+    MSLogDebug([MSAppCenter logTag], @"Settings:update[%@]=%@", keyPrefixed, update);
+
+    // Copy store as a mutable version.
+    NSMutableDictionary *d = [store mutableCopy];
```
MatkovIvan: d?

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
...
+
+  return update;
+}
+
+- (NSDictionary *)updateDictionary:(NSDictionary *)dict forKey:(NSString *)key {
+  return [self updateDictionary:dict forKey:key expiration:0.0];
+}
+
+- (BOOL)updateObject:(id)o forKey:(NSString *)key expiration:(float)expiration {
+  NSDictionary *update = [self updateDictionary:@{@"v" : o} forKey:key expiration:expiration];
```
MatkovIvan: v?

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
annakocheshkova: I'm not sure what it is exactly... 🤔

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
annakocheshkova: I don't think yet another migration would be a good idea...

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSAppCenterUserDefaults.h"
+#import "MSAppCenterInternal.h"
+#import "MSLogger.h"
+
+static NSString *const kMSUserDefaultsTs = @"_ts";
```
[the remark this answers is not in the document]
annakocheshkova: Again, it will require yet another migration

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
iamclement: The base class migration / child class key definition would work now but is not ideal simply because we will always have to make sure keys are used only after the migration as you said. We had to deal with such issues when we first introduced the swizzling. We had to declare methods and classes to swizzle depending on services used. We might be able to reuse the same mechanism.

iamclement: Here is how it works. We can have the objective-c runtime do the hard work for us thanks to the [load](https://developer.apple.com/documentation/objectivec/nsobject/1418815-load?language=objc) class method. It's called when classes are loaded by the objective-c runtime. It's synchronous and a class is "loaded" before its categories. Since this is happening before the wrapper initialization no App Center user defaults is get/set and the migration happens after all keys are collected. 

So the wrapper will implement its load method and initialize a class level property exposing a mutable dictionary which will receive the keys to migrate. All services would create a category over the user defaults wrapper. These categories will implement their load method. These load methods will add the keys to be migrated to the user default wrapper dictionary property. Then, when the wrapper will initialize himself it will refer to the dictionary property to migrate all the keys at once.  This should be quite safe.

Now there is a little trick with the categories in our code base. Since we don't require the `-ObjC` flag the categories are not wired automatically at compile time. The only thing you have to do to force it is to implement the categories in classes that we know will be loaded for instance the services main class implementations.

[Example of a category implementing the load method. ](https://github.com/microsoft/appcenter-sdk-apple/blob/d3eb6c942c39be49a6363f28c86f799eb5c1f5c0/AppCenterDistribute/AppCenterDistribute/Internals/MSDistributeAppDelegate.m#L40)
DM me if you need help :).

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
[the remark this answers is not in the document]
iamclement: I think it's dead code. Can be removed in another PR. That was for values expiration but I don't this we use it.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
iamclement: That was for values expiration but I don't this we use it.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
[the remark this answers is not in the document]
iamclement: Or can clean up now all the expiration logic as you wish.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
iamclement: Talked in DM. @annakocheshkova good point, the categories might be useless, just implementing the services `load` methods should be enough. Although I don't think the ordering of class loading between classes of the same lib is guaranteed calling the wrapper in the service load method should force load it.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -274,6 +274,14 @@ + (void)setDelegate:(_Nullable id<MSCrashesDelegate>)delegate {
 
 #pragma mark - Service initialization
 
++ (void)load {
+  [MSAppCenterUserDefaults addKeysToMigrate:@{
+    @"kMSCrashesIsEnabledKey" : @"MSAppCenterCrashesIsEnabledKey",              // MSCrashes
```
iamclement: Are these `// MSCrashes` comments useful?

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m
iamclement: I would add a comment to explain that it's a system key so we don't use the App Center wrapper.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
annakocheshkova: That's the format I adopter by [this comment](https://github.com/microsoft/appcenter-sdk-apple/pull/2013#discussion_r398552143)

annakocheshkova: Makes not much sense here, but in other places points to the place where it's used

AppCenterCrashes/AppCenterCrashesTests/MSMockNsUserDefaults.m
```diff
@@ -0,0 +1,46 @@
...
+    OCMStub([_mockNSUserDefaults standardUserDefaults]).andReturn(self);
+  }
+  return self;
+}
+
+- (void)setObject:(id)anObject forKey:(NSString *)aKey {
+
+  // Don't store nil objects.
+  if (!anObject) {
+    return;
```
MatkovIvan: Should we remove the value in this case?

annakocheshkova: I don't think so. It will break many tests and besides, there's a remove functions for that.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,93 @@
...
+}
+
++ (instancetype)shared {
+  dispatch_once(&onceToken, ^{
+    sharedInstance = [[MSAppCenterUserDefaults alloc] init];
+    NSDictionary *changedKeys = @{
+      @"MSChannelStartTimer" : @"MSAppCenterChannelStartTimer",         // MSChannelUnitDefault
+      @"pastDevicesKey" : @"MSAppCenterPastDevicesKey",                 // MSDeviceTrackerPrivate
+      @"MSInstallId" : @"MSAppCenterInstallId",                         // MSAppCenterInternal
+      @"MSAppCenterIsEnabled" : @"MSAppCenterAppCenterIsEnabled",       // MSAppCenter
```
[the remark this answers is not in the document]
annakocheshkova: @jaeklim @iamclement what do you think? For me it's ok

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,93 @@
...
+
++ (NSDictionary<NSString *, NSString *> *)keysToMigrate {
+  return keysToMigrate;
+}
+
++ (void)resetSharedInstance {
+
+  // Reset the once_token so dispatch_once will run again.
+  onceToken = 0;
+  sharedInstance = nil;
```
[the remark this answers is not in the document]
annakocheshkova: Actually this was incorrect. We use `resetSharedInstance` in tests, and we do not need this logic in tests: the `keysToMigrate` are initialized in `load` method. This is required for sdk to work and saves us from multiple nil checks.
I replaced it with removing all objects from the array.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
```diff
@@ -0,0 +1,65 @@
...
+ */
++ (instancetype)shared;
+
+/**
+ * Returns an array of keys to be migrated.
+ */
++ (NSDictionary<NSString *, NSString *> *)keysToMigrate;
+
+/**
+ * Append an array of keys to migrated.
```
iamclement:
```suggestion
 * Append an array of keys to migrate.
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @return the shared settings.
+ */
++ (instancetype)shared;
+
+/**
+ * Returns an array of keys to be migrated.
+ */
++ (NSDictionary<NSString *, NSString *> *)keysToMigrate;
```
iamclement: It sounds like `keysToMigrate` and `migrateKeys` are only used by the `MSAppCenterUserDefaults` implementation and unit tests so should be in a private header by our convention.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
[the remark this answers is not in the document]
iamclement: I don't have a strong opinion on that one.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,94 @@
...
+    return;
+  }
+  MSLogVerbose([MSAppCenter logTag], @"Migrating the old NSDefaults keys to new ones.");
+  for (NSString *oldKey in [migratedKeys allKeys]) {
+    NSString *newKey = migratedKeys[oldKey];
+    id value = [[NSUserDefaults standardUserDefaults] objectForKey:oldKey];
+    if (value != nil) {
+      [[NSUserDefaults standardUserDefaults] setObject:value forKey:newKey];
+      [[NSUserDefaults standardUserDefaults] removeObjectForKey:oldKey];
+      MSLogVerbose([MSAppCenter logTag], @"Migrating the key %@ -> %@", oldKey, newKey);
```
iamclement: nit:
```suggestion
      MSLogVerbose([MSAppCenter logTag], @"Migrating key %@ -> %@", oldKey, newKey);
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+  [self setObject:@YES forKey:kMSAppCenterUserDefaultsMigratedKey];
+}
+
+- (NSString *)getAppCenterKeyFrom:(NSString *)key {
+  NSAssert(![key hasPrefix:kMSUserDefaultsPrefix], @"Please do not prepend the key with 'MSAppCenter'. It's done automatically.");
```
iamclement: Isn't that a bit overkill? We'll do it every-time we'll compose a key.

iamclement: Ah but it's supposed to be disabled in release builds, should be fine then.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
iamclement: Got it thanks.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 3.1.1 (In development)
 
+### App Center
+
+* **[Improvement]** Use standartised `NSUserDefaults` keys with the prefix **MSAppCenter** for all the keys used in the SDK.
```
iamclement: How about?
```suggestion
* **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few missing namespace.
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
iamclement: Should also add that it's to be called from a `load` method or won't be migrated.

AppCenterCrashes/AppCenterCrashesTests/MSMockNsUserDefaults.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+@interface MSMockNsUserDefaults : NSUserDefaults
+
+/**
+ * Clear dictionary
```
MatkovIvan:
```suggestion
 * Clear dictionary.
```

AppCenter/AppCenterTests/Util/MSMockUserDefaults.m
```diff
@@ -20,12 +21,16 @@ - (instancetype)init {
...
     // Mock MSUserDefaults shared method to return this instance.
-    _mockMSUserDefaults = OCMClassMock([MSUserDefaults class]);
+    _mockMSUserDefaults = OCMClassMock([MSAppCenterUserDefaults class]);
     OCMStub([_mockMSUserDefaults shared]).andReturn(self);
   }
   return self;
 }
 
+- (void)migrateKeys:(__unused NSDictionary *)migratedKeys {
+  [self setObject:@(1) forKey:kMSMockMigrationKey];
```
MatkovIvan:
```suggestion
  [self setObject:@1 forKey:kMSMockMigrationKey];
```

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -62,10 +62,26 @@
...
+    @"kMSDistributeIsEnabledKey" : @"MSAppCenterDistributeIsEnabledKey",              // MSDistributePrivate
+    @"MSPostponedTimestamp" : @"MSAppCenterPostponedTimestamp",                       // MSDistributePrivate
+    @"MSSDKHasLaunchedWithDistribute" : @"MSAppCenterSDKHasLaunchedWithDistribute",   // MSDistributePrivate
+    @"MSMandatoryRelease" : @"MSAppCenterMandatoryRelease",                           // MSDistributePrivate
+    @"MSDistributionGroupId" : @"MSAppCenterDistributionGroupId",                     // MSDistributePrivate
+    @"MSUpdateSetupFailedPackageHash" : @"MSAppCenterUpdateSetupFailedPackageHash",   // MSDistributePrivate
+    @"MSDownloadedReleaseHash" : @"MSAppCenterDownloadedReleaseHash",                 // MSDistributePrivate
+    @"MSDownloadedReleaseId" : @"MSAppCenterDownloadedReleaseId",                     // MSDistributePrivate
+    @"MSDownloadedDistributionGroupId" : @"MSAppCenterDownloadedDistributionGroupId", // MSDistributePrivate
+    @"MSTesterAppUpdateSetupFailed" : @"MSAppCenterTesterAppUpdateSetupFailed"        // MSDistributePrivate
```
MatkovIvan: `// MSDistributePrivate` doesn't not really explain where it's used (it was a main intention for this).
I guess it makes sense to say that "DistributeIsEnabledKey" (for example) is used in `// [MSDistribute isEnabled]`

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -80,11 +96,11 @@ - (instancetype)init {
      * data won't be automatically deleted by uninstall so we should detect it and clean up keychain data when Distribute service gets
      * initialized.
      */
-    NSNumber *flag = [MS_USER_DEFAULTS objectForKey:kMSSDKHasLaunchedWithDistribute];
+    NSNumber *flag = [MS_APP_CENTER_USER_DEFAULTS objectForKey:kMSSDKHasLaunchedWithDistribute];
     if (!flag) {
       MSLogInfo([MSDistribute logTag], @"Delete update token if exists.");
       [MSKeychainUtil deleteStringForKey:kMSUpdateTokenKey];
-      [MS_USER_DEFAULTS setObject:@(1) forKey:kMSSDKHasLaunchedWithDistribute];
+      [MS_APP_CENTER_USER_DEFAULTS setObject:@(1) forKey:kMSSDKHasLaunchedWithDistribute];
```
MatkovIvan:
```suggestion
      [MS_APP_CENTER_USER_DEFAULTS setObject:@1 forKey:kMSSDKHasLaunchedWithDistribute];
```

AppCenterPush/AppCenterPushTests/Util/MSMockUserDefaults.m
```diff
@@ -20,12 +21,16 @@ - (instancetype)init {
...
     // Mock MSUserDefaults shared method to return this instance.
-    _mockMSUserDefaults = OCMClassMock([MSUserDefaults class]);
+    _mockMSUserDefaults = OCMClassMock([MSAppCenterUserDefaults class]);
     OCMStub([_mockMSUserDefaults shared]).andReturn(self);
   }
   return self;
 }
 
+- (void)migrateKeys:(__unused NSDictionary *)migratedKeys {
+  [self setObject:@(1) forKey:kMSMockMigrationKey];
```
MatkovIvan:
```suggestion
  [self setObject:@1 forKey:kMSMockMigrationKey];
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,94 @@
...
++ (void)addKeysToMigrate:(NSDictionary<NSString *, NSString *> *)keys {
+  [keysToMigrate addEntriesFromDictionary:keys];
+}
+
++ (instancetype)shared {
+  dispatch_once(&onceToken, ^{
+    sharedInstance = [[MSAppCenterUserDefaults alloc] init];
+    NSDictionary *changedKeys = @{
+      @"MSChannelStartTimer" : @"MSAppCenterChannelStartTimer",         // MSChannelUnitDefault
+      @"pastDevicesKey" : @"MSAppCenterPastDevicesKey",                 // MSDeviceTrackerPrivate
```
iamclement, line 31: Sometimes we use the `Key` suffix like for this key, no big deal but this is our chance for harmonizing so let's remove all the `key` suffix from the keys string I think it's useless.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,94 @@
...
+
++ (void)addKeysToMigrate:(NSDictionary<NSString *, NSString *> *)keys {
+  [keysToMigrate addEntriesFromDictionary:keys];
+}
+
++ (instancetype)shared {
+  dispatch_once(&onceToken, ^{
+    sharedInstance = [[MSAppCenterUserDefaults alloc] init];
+    NSDictionary *changedKeys = @{
+      @"MSChannelStartTimer" : @"MSAppCenterChannelStartTimer",         // MSChannelUnitDefault
```
iamclement, line 30: This is tricky because it's per channels so the key is dynamic[ look here](https://github.com/microsoft/appcenter-sdk-apple/blob/8e1835b06ae67054a091bb826d7faaed1ad29628/AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m#L409). I'd say you can scan the user defaults, list keys prefixed with `MSChannelStartTimer` and migrate.

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -45,6 +45,13 @@ @implementation MSAnalytics
 
 #pragma mark - Service initialization
 
++ (void)load {
+  [MSAppCenterUserDefaults addKeysToMigrate:@{
```
iamclement, line 49: There's one tricky set of keys to migrate here `AnalyticsIsEnabledKey/xxxx`, the transmission targets enabled state keys. It's dynamically created with the target key from the token. I we'll need to list all of them from the user defaults but not sure about the state of the defaults at load time.

AppCenterCrashes/AppCenterCrashes/MSCrashes.mm
```diff
@@ -274,6 +274,14 @@ + (void)setDelegate:(_Nullable id<MSCrashesDelegate>)delegate {
...
+    @"MSAppDidReceiveMemoryWarning" : @"MSAppCenterAppDidReceiveMemoryWarning", // MSCrashes
+    @"MSUserConfirmation" : @"MSAppCenterUserConfirmation"                      // MSCrashes
```
iamclement, line 281: `MSAppCenterUserConfirmation` could be less generic, for instance: `MSAppCenterCrashesUserConfirmation`

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -27,13 +27,19 @@ + (instancetype)shared {
...
     sharedInstance = [[MSAppCenterUserDefaults alloc] init];
     NSDictionary *changedKeys = @{
-      @"MSChannelStartTimer" : @"MSAppCenterChannelStartTimer",         // MSChannelUnitDefault
-      @"pastDevicesKey" : @"MSAppCenterPastDevicesKey",                 // MSDeviceTrackerPrivate
-      @"MSInstallId" : @"MSAppCenterInstallId",                         // MSAppCenterInternal
-      @"MSAppCenterIsEnabled" : @"MSAppCenterAppCenterIsEnabled",       // MSAppCenter
-      @"MSEncryptionKeyMetadata" : @"MSAppCenterEncryptionKeyMetadata", // MSEncrypterPrivate
-      @"SessionIdHistory" : @"MSAppCenterSessionIdHistory",             // MSSessionContext
-      @"UserIdHistory" : @"MSAppCenterUserIdHistory"                    // MSUserIdContext
+      @"MSChannelStartTimer*" : @"MSAppCenterChannelStartTimer",        // [MSChannelUnitDefault oldestPendingLogTimestampKey]
```
iamclement, line 30: I am not completely sold with the wildcard thing as it introduces parsing and reserves the `*` character. How about you introduce a new simple type (e.g.:`MSUserDefaultsBaseKey`) inheriting from `NSString`. The new class can simply be declared within `MSAppCenterUserDefaults.h`. It doesn't need to do anything special. Then the migration will just filter on this type.

AppCenterDistribute/AppCenterDistribute/MSDistribute.m
```diff
@@ -69,16 +69,38 @@ @implementation MSDistribute
...
-    @"MSMandatoryRelease" : @"MSAppCenterMandatoryRelease",                           // MSDistributePrivate
-    @"MSDistributionGroupId" : @"MSAppCenterDistributionGroupId",                     // MSDistributePrivate
-    @"MSUpdateSetupFailedPackageHash" : @"MSAppCenterUpdateSetupFailedPackageHash",   // MSDistributePrivate
-    @"MSDownloadedReleaseHash" : @"MSAppCenterDownloadedReleaseHash",                 // MSDistributePrivate
-    @"MSDownloadedReleaseId" : @"MSAppCenterDownloadedReleaseId",                     // MSDistributePrivate
-    @"MSDownloadedDistributionGroupId" : @"MSAppCenterDownloadedDistributionGroupId", // MSDistributePrivate
-    @"MSTesterAppUpdateSetupFailed" : @"MSAppCenterTesterAppUpdateSetupFailed"        // MSDistributePrivate
+    @"kMSDistributeIsEnabledKey" : @"MSAppCenterDistributeIsEnabled",   // [MSDistribute isEnabled]
+    @"MSPostponedTimestamp" : @"MSAppCenterPostponedTimestamp",
+                                                                        // [MSDistributePrivate notifyUpdateAction],
```
MatkovIvan, line 74: There is no `MSDistributePrivate` class, it's just a header

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,160 @@
...
+}
+
+- (void)removeObjectForKey:(NSString *)key {
+  NSString *keyPrefixed = [self getAppCenterKeyFrom:key];
+  [[NSUserDefaults standardUserDefaults] removeObjectForKey:keyPrefixed];
+}
+
+@end
+
+@interface MSUserDefaultsWildcardKey ()
```
iamclement, line 136: We probably could use a `pragma` here to delimitate the classes impls.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
```diff
@@ -0,0 +1,57 @@
...
+ * @param key the key to remove.
+ */
+- (void)removeObjectForKey:(NSString *)key;
+
+@end
+
+// A class defining that the instance of this class needs wildcard migration.
+// This means that for instances of this class, MSAppCenterUserDefautls will
+// search for the old keys starting with this key and migrate all of them.
+@interface MSUserDefaultsWildcardKey : NSString
```
MatkovIvan, line 52: `inheriting from NSString` is a bad idea

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,160 @@
...
+      @"MSEncryptionKeyMetadata" : @"MSAppCenterEncryptionKeyMetadata", // [MSEncrypterPrivate getCurrentKeyTag],
+                                                                        // [MSEncrypterPrivate rotateToNewKeyTag]
+      @"SessionIdHistory" : @"MSAppCenterSessionIdHistory",             // [MSSessionContext init],
+                                                                        // [MSSessionContext setSessionId],
+                                                                        // [MSSessionContext clearSessionHistoryAndKeepCurrentSession]
+      @"UserIdHistory" : @"MSAppCenterUserIdHistory"                    // [MSUserIdContext init], [MSUserIdContext setUserId],
+                                                                        // [MSUserIdContext clearUserIdHistory]
+    };
+    [keysToMigrate addEntriesFromDictionary:changedKeys];
+    [sharedInstance migrateKeys:keysToMigrate];
```
MatkovIvan, line 46: Let's implicit call from creating instance method

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,160 @@
...
+  [keysToMigrate removeAllObjects];
+}
+
+- (void)migrateKeys:(NSDictionary *)migratedKeys {
+  NSNumber *hasMigrated = [self objectForKey:kMSAppCenterUserDefaultsMigratedKey];
+  if (hasMigrated) {
+    return;
+  }
+  MSLogVerbose([MSAppCenter logTag], @"Migrating the old NSDefaults keys to new ones.");
+  for (NSString *oldKey in [migratedKeys allKeys]) {
```
MatkovIvan, line 69: `allKeys` is not needed here

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
```diff
@@ -45,6 +45,14 @@ @implementation MSAnalytics
 
 #pragma mark - Service initialization
 
++ (void)load {
```
MatkovIvan, line 48: `load` called from first usage. It won't work correctly on start services one after another like in RN

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,151 @@
...
+#import "MSLoggerInternal.h"
+#import "MSTestFrameworks.h"
+#import "MSUtility.h"
+#import "MSWrapperLogger.h"
+
+@interface MSUserDefaultsTests : XCTestCase
+
+@end
+
+static NSString *const kMSAppCenterUserDefaultsMigratedKey = @"MSAppCenter310UserDefaultsMigratedKey";
```
MatkovIvan, line 15: it must be stored per module. See the comment about `load`

AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m
iamclement: The load method of our classes are called only one time pre-main so it's really safe but that's because we are a static framework. I am not entirely sure about the behavior if one day we decide to ship a dynamic framework. Let's discuss options.

iamclement: The `load` method as we know in ObjC isn't supported in plain Swift: https://stackoverflow.com/questions/24898453/swift-objective-c-load-class-method. That doesn't make it future proof. Let's go back to migrating in the `init` method. Though we need to pay attention to the bug I mentioned about the need to  migrate before the `isEnabled` state is set in the abstraction.

AppCenter/AppCenter/MSAppCenter.m
```diff
@@ -241,6 +241,30 @@ - (instancetype)init {
...
     _enabledStateUpdating = NO;
+    NSDictionary *changedKeys = @{
+      @"MSAppCenterChannelStartTimer" : MSPrefixKeyFrom(@"MSChannelStartTimer"),
+      // [MSChannelUnitDefault oldestPendingLogTimestampKey]
+      @"MSAppCenterPastDevices" : @"pastDevicesKey",
+      // [MSDeviceTracker init],
+      // [MSDeviceTracker device],
+      // [MSDeviceTracker clearDevices]
+      @"MSAppCenterInstallId" : @"MSInstallId",
+      // [MSAppCenterInternal installId]
```
MatkovIvan, line 252:
```suggestion
      // [MSAppCenter installId]
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,116 @@
...
+  onceToken = 0;
+  sharedInstance = nil;
+}
+
+- (void)migrateKeys:(NSDictionary *)migratedKeys forService:(NSString *)service {
+  NSNumber *serviceMigrated = [self objectForKey:[NSString stringWithFormat:kMSAppCenterUserDefaultsMigratedKeyFormat, service]];
+  if (serviceMigrated) {
+    return;
+  }
+
```
MatkovIvan, line 34: Add a comment or remove empty line

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,116 @@
...
+      for (NSString *userDefaultsKey in userDefaultKeys) {
+        if ([userDefaultsKey hasPrefix:oldKeyPrefix]) {
+          NSString *suffix = [userDefaultsKey stringByReplacingOccurrencesOfString:oldKeyPrefix withString:@""];
+          NSString *newKeyWithSuffix = [newKeyString stringByAppendingString:suffix];
+          id value = [[NSUserDefaults standardUserDefaults] objectForKey:userDefaultsKey];
+          [self swapKeys:userDefaultsKey newKey:newKeyWithSuffix value:value];
+        }
+      }
+    } else {
+      MSLogError([MSAppCenter logTag], @"Unsupported type %@ for key %@", [oldKey class], oldKey);
```
MatkovIvan, line 60: Assert?

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,116 @@
...
+    return;
+  }
+  [[NSUserDefaults standardUserDefaults] setObject:value forKey:newKey];
+  [[NSUserDefaults standardUserDefaults] removeObjectForKey:oldKey];
+  MSLogVerbose([MSAppCenter logTag], @"Migrating key %@ -> %@", oldKey, newKey);
+}
+
+- (NSString *)getAppCenterKeyFrom:(NSObject *)key {
+  if (![key isKindOfClass:[NSString class]]) {
+    MSLogError([MSAppCenter logTag], @"Unsupported type %@ for key %@", [key class], key);
```
MatkovIvan, line 78: Assert?

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,126 @@
...
+  }
+  [MSAppCenterUserDefaults resetSharedInstance];
+}
+
+- (void)testSettingsAlreadyMigrated {
+
+  // If
+  NSString *testValue = @"testValue";
+  [[NSUserDefaults standardUserDefaults] setObject:testValue forKey:@"pastDevicesKey"];
+  [[NSUserDefaults standardUserDefaults] setObject:@(1) forKey:kMSAppCenterUserDefaultsMigratedKey];
```
MatkovIvan, line 32:
```suggestion
  [[NSUserDefaults standardUserDefaults] setObject:@YES forKey:kMSAppCenterUserDefaultsMigratedKey];
```

AppCenter/AppCenterTests/Util/MSMockUserDefaults.m
```diff
@@ -20,12 +21,16 @@ - (instancetype)init {
...
     // Mock MSUserDefaults shared method to return this instance.
-    _mockMSUserDefaults = OCMClassMock([MSUserDefaults class]);
+    _mockMSUserDefaults = OCMClassMock([MSAppCenterUserDefaults class]);
     OCMStub([_mockMSUserDefaults shared]).andReturn(self);
   }
   return self;
 }
 
+- (void)migrateKeys:(__unused NSDictionary *)migratedKeys forService:(nonnull NSString *)service {
+  [self setObject:@1 forKey:[NSString stringWithFormat:kMSMockMigrationKey, service]];
```
MatkovIvan, line 31:
```suggestion
  [self setObject:@YES forKey:[NSString stringWithFormat:kMSMockMigrationKey, service]];
```

AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSAnalyticsTransmissionTarget.m
```diff
@@ -32,7 +32,7 @@ - (instancetype)initWithTransmissionTargetToken:(NSString *)token
 
     // Disable if ancestor is disabled.
     if (![self isImmediateParent]) {
-      [MS_USER_DEFAULTS setObject:@(NO) forKey:self.isEnabledKey];
+      [MS_APP_CENTER_USER_DEFAULTS setObject:@(NO) forKey:self.isEnabledKey];
```
MatkovIvan, line 35:
```suggestion
      [MS_APP_CENTER_USER_DEFAULTS setObject:@NO forKey:self.isEnabledKey];
```

AppCenterDistribute/AppCenterDistributeTests/MSDistributeTests.m
```diff
@@ -2789,7 +2798,7 @@ - (void)testRemoveDownloadedReleaseDetailsIfUpdated {
 
 - (void)testStartUpdateWhenEnabledButDidNotStart {
   NSString *isEnabledKey = @"MSAppCenterIsEnabled";
-  [MS_USER_DEFAULTS setObject:@(YES) forKey:isEnabledKey];
+  [MS_APP_CENTER_USER_DEFAULTS setObject:@(YES) forKey:isEnabledKey];
```
MatkovIvan, line 2801:
```suggestion
  [MS_APP_CENTER_USER_DEFAULTS setObject:@YES forKey:isEnabledKey];
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
iamclement: Well, inheriting from `NSString` with `NSString` being a class cluster doesn't make a lot of sense here. Object composition is the right choice indeed.

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
```diff
@@ -0,0 +1,64 @@
...
+ * @param migratedKeys a dictionary for keys that contains new key as a key of dictionary and old key as a value.
+ * @param service service name.
+ */
+- (void)migrateKeys:(NSDictionary *)migratedKeys forService:(NSString *)service;
+
+@end
+
+// A class defining that the instance of this class needs wildcard migration.
+// This means that for instances of this class, MSAppCenterUserDefautls will
+// search for the old keys starting with this key and migrate all of them.
```
iamclement, line 55:
```suggestion
/**
 * A class defining that the instance of this class needs wildcard migration.
 * This means that for instances of this class, MSAppCenterUserDefautls will
 * search for the old keys starting with this key and migrate all of them.
 */
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,107 @@
...
+      [self swapKeys:(NSString *)oldKey newKey:newKeyString value:value];
+    } else {
+      NSAssert([oldKey isKindOfClass:[MSUserDefaultsPrefixKey class]], @"Unsupported type");
+
+      // List all the keys starting with oldKey.
+      NSString *oldKeyPrefix = ((MSUserDefaultsPrefixKey *)oldKey).keyPrefix;
+      NSArray *userDefaultKeys = [[[NSUserDefaults standardUserDefaults] dictionaryRepresentation] allKeys];
+      for (NSString *userDefaultsKey in userDefaultKeys) {
+        if ([userDefaultsKey hasPrefix:oldKeyPrefix]) {
+          NSString *suffix = [userDefaultsKey stringByReplacingOccurrencesOfString:oldKeyPrefix withString:@""];
```
iamclement, line 50: nit:
```suggestion
          NSString *suffix = [userDefaultsKey substringFromIndex:[oldKeyPrefix length]];
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,107 @@
...
+- (void)swapKeys:(NSString *)oldKey newKey:(NSString *)newKey value:(id)value {
+  if (value == nil) {
+    return;
+  }
+  [[NSUserDefaults standardUserDefaults] setObject:value forKey:newKey];
+  [[NSUserDefaults standardUserDefaults] removeObjectForKey:oldKey];
+  MSLogVerbose([MSAppCenter logTag], @"Migrating key %@ -> %@", oldKey, newKey);
+}
+
+- (NSString *)getAppCenterKeyFrom:(NSObject *)key {
```
iamclement, line 70: Why is key not `NSString *`? Same for `objectForKey:` , `setObject:forKey:`, `removeObjectForKey:`?

MatkovIvan: Based on my feedback https://github.com/microsoft/appcenter-sdk-apple/pull/2013#discussion_r402982589

So, it's to support prefix rule

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
iamclement: Indeed, sorry I misled on that.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 3.1.1 (In development)
 
+### App Center
+
+* **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few missing namespace.
```
iamclement, line 7:
```suggestion
* **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few keys missing namespace.
```

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,108 @@
...
+    NSString *newKeyString = (NSString *)newKey;
+    if ([oldKey isKindOfClass:[NSString class]]) {
+      id value = [[NSUserDefaults standardUserDefaults] objectForKey:(NSString *)oldKey];
+      [self swapKeys:(NSString *)oldKey newKey:newKeyString value:value];
+    } else {
+      NSAssert([oldKey isKindOfClass:[MSUserDefaultsPrefixKey class]], @"Unsupported type");
+
+      // List all the keys starting with oldKey.
+      NSString *oldKeyPrefix = ((MSUserDefaultsPrefixKey *)oldKey).keyPrefix;
+      NSArray *userDefaultKeys = [[[NSUserDefaults standardUserDefaults] dictionaryRepresentation] allKeys];
```
MatkovIvan, line 47: Nitpick: No need for `allKeys` here, you can get value from that dictionary

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.m
```diff
@@ -0,0 +1,108 @@
...
+  if (value == nil) {
+    return;
+  }
+  [[NSUserDefaults standardUserDefaults] setObject:value forKey:newKey];
+  [[NSUserDefaults standardUserDefaults] removeObjectForKey:oldKey];
+  MSLogVerbose([MSAppCenter logTag], @"Migrating key %@ -> %@", oldKey, newKey);
+}
+
+- (NSString *)getAppCenterKeyFrom:(NSString *)key {
+  NSString *keyString = (NSString *)key;
```
MatkovIvan, line 71: Redundant

AppCenter/AppCenter/Internals/Storage/MSAppCenterUserDefaultsPrivate.h
```diff
@@ -0,0 +1,22 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSAppCenterUserDefaults ()
+
+/**
+ * Returns an array of keys to be migrated.
+ */
++ (NSDictionary *)keysToMigrate;
```
MatkovIvan, line 13: I don't see implementation of this method. Is it supposed to return all already migrated keys for all services?

AppCenter/AppCenter/Internals/Model/Util/MSAppCenterUserDefaults.h
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+#define MSPrefixKeyFrom(_key) [[MSUserDefaultsPrefixKey alloc] initWithPrefix:_key]
+
+NS_ASSUME_NONNULL_BEGIN
+
+static NSString *const kMSUserDefaultsPrefix = @"MSAppCenter";
```
MatkovIvan, line 10: Why do you need it in this header? Move it to `*.m` or `*Private.h` at least

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,126 @@
...
+
+static NSString *const kMSAppCenterUserDefaultsMigratedKey = @"MSAppCenter310AppCenterUserDefaultsMigratedKey";
+
+@implementation MSUserDefaultsTests
+
+- (void)setUp {
+  NSArray *userDefaultKeys = [[[NSUserDefaults standardUserDefaults] dictionaryRepresentation] allKeys];
+  for (NSString *key in userDefaultKeys) {
+    [[NSUserDefaults standardUserDefaults] removeObjectForKey:key];
+  }
```
MatkovIvan, line 23: Does [`removePersistentDomain`](https://developer.apple.com/documentation/foundation/userdefaults/1417339-removepersistentdomain) work here?

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,126 @@
...
+  // Then
+  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:[kMSUserDefaultsPrefix stringByAppendingString:key]]);
+}
+
+- (void)testMigrateUserDefaultSettings {
+
+  NSArray *suffixes = @[ @"-suffix1", @"/suffix2", @"suffix3" ];
+  NSString *wildcard = @"okeyTestWildcard";
+  NSString *expectedWildcard = @"MSAppCenterOkeyTestWildcard";
+  // If
```
MatkovIvan, line 68: Empty line before comment

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,126 @@
...
+
+  // When
+  [userDefaults removeObjectForKey:key];
+
+  // Then
+  XCTAssertNil([[NSUserDefaults standardUserDefaults] objectForKey:[kMSUserDefaultsPrefix stringByAppendingString:key]]);
+}
+
+- (void)testMigrateUserDefaultSettings {
+
```
MatkovIvan, line 64: Empty line only before comment

AppCenter/AppCenterTests/MSUserDefaultsTests.m
```diff
@@ -0,0 +1,126 @@
...
+  NSString *expectedWildcard = @"MSAppCenterOkeyTestWildcard";
+  // If
+  NSDictionary *keys = @{
+    @"MSAppCenterKeyTest1" : @"okeyTest1",
+    @"MSAppCenterKeyTest2" : @"okeyTest2",
+    @"MSAppCenterKeyTest3" : @"okeyTest3",
+    @"MSAppCenterKeyTest4" : @"okeyTest4",
+    expectedWildcard : MSPrefixKeyFrom(wildcard)
+  };
+
```
MatkovIvan, line 76: Empty line only before comment

AppCenterCrashes/AppCenterCrashesTests/MSApplicationForwarderTests.m
```diff
@@ -55,9 +55,8 @@ - (void)testRegisterForwarding {
   XCTAssertNil(lastException);
 
   // Enable crash on exeptions
-  MSMockUserDefaults *settings = [MSMockUserDefaults new];
+  MSMockNsUserDefaults *settings = [MSMockNsUserDefaults new];
   [settings setObject:@YES forKey:@"NSApplicationCrashOnExceptions"];
-
```
MatkovIvan, line 60: Empty line before comments

AppCenterCrashes/AppCenterCrashesTests/MSApplicationForwarderTests.m
```diff
@@ -55,9 +55,8 @@ - (void)testRegisterForwarding {
   XCTAssertNil(lastException);
 
   // Enable crash on exeptions
```
MatkovIvan, line 57:
```suggestion
  // Enable crash on exceptions.
```

AppCenterCrashes/AppCenterCrashesTests/MSMockNsUserDefaults.m
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSMockNsUserDefaults.h"
+#import "MSTestFrameworks.h"
+
+@interface MSMockNsUserDefaults ()
```
MatkovIvan, line 7:
```suggestion
@interface MSMockNSUserDefaults ()
```

MatkovIvan: or
```suggestion
@interface MSMockUserDefaults ()
```

AppCenter/AppCenterTests/MSUserDefaultsTests.m
annakocheshkova: I don't think so. we need to remove keys without prefix, too.

iamclement:
This change is going in the next release cycle.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 3.1.1 (In development)
 
+### App Center
+
+* **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few keys missing namespace.
```
MatkovIvan, line 7: Move to the next version

PR #2031, opened 2020-04-17
Fix event filter in test apps

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Event filter didn't work because of signature change. Also, fix compilation error in Xcode 11

The change in `MSConstants+Flags.h` is required to include private header (but it is not finally required for `MSEventFilter`)

## Related PRs or issues

[AB#79201](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/79201)

### DIFF
--- a/SasquatchMac/SasquatchMac/ViewControllers/EventFilterViewController.swift
+++ b/SasquatchMac/SasquatchMac/ViewControllers/EventFilterViewController.swift
@@ -6,16 +6,20 @@ import Cocoa
 class EventFilterViewController: NSViewController {
 
   var appCenter: AppCenterDelegate = AppCenterProvider.shared().appCenter!
+  private var eventFilterStarted = false
 
   @IBOutlet weak var setEnabledButton: NSButton!
 
   @IBAction func setEnabled(_ sender: NSButton) {
+    if !eventFilterStarted {
+      appCenter.startEventFilterService()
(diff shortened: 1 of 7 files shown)


SasquatchMac/SasquatchMac/ViewControllers/EventFilterViewController.swift
```diff
@@ -6,16 +6,20 @@ import Cocoa
 class EventFilterViewController: NSViewController {
 
   var appCenter: AppCenterDelegate = AppCenterProvider.shared().appCenter!
+  private var eventFilterStarted = false
 
   @IBOutlet weak var setEnabledButton: NSButton!
 
   @IBAction func setEnabled(_ sender: NSButton) {
+    if !eventFilterStarted {
```
annakocheshkova, line 14: What would be if the state is being changed to disabled? Would we still start it?

annakocheshkova: I mean if it is already enabled on the first launch

MatkovIvan: There is no way to start in on the launch, it's intended to start it manually here

PR #2032, opened 2020-04-17
Update Unit tests to support Xcode 11

AnastasiaKubova:

## Description

- We need to update unit tests, so they runs correctly on Xcode 11. There are several tasks connected with unit tests: 
- Update **OHHTTPStubs** to the latest version (aka `The following binaries use incompatible versions of Swift: /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/lib/libXCTestSwiftSupport.dylib` error).
- Fix errors related to renaming. Example: `OH prefix was dropped`.
- Fix several kinds of warnings. Example: "Implicit conversion changes signedness: `'UIBackgroundFetchResult' (aka 'enum UIBackgroundFetchResult') to 'NSInteger' (aka 'long')`" Solution for example: replace `equalToInteger(xxx)` with `equalToUnsignedInteger(xxx)`

## Related PRs or issues

[AB#79030](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/79030)


Vendor/tvOS/OHHTTPStubs/OHHTTPStubs.framework/Modules/module.modulemap
```diff
@@ -1,5 +1,5 @@
 framework module OHHTTPStubs {
-  umbrella header "OHHTTPStubsUmbrella.h"
+  umbrella header "OHHTTPStubs.h"
```
MatkovIvan, line 2: Remove frameworks if you already have submodule (and do rebase to avoid keeping another manually built binary in thehistory)

Sasquatch/SasquatchSwift-Preview Extension.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MatkovIvan, line 1: Not related to the PR

Scripts/build-ohttpstubs.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
```
MatkovIvan, line 1: Based on [OHHTTPStubs readme](https://github.com/AliSoftware/OHHTTPStubs#using-the-right-swift-version-for-your-project) - "CocoaPods will compile OHHTTPStubs with the right Swift Version matching the one you use for your project automatically"

So, it already has the logic to choose right parameters based on XCode version. Also, all swift code is placed to separate target, so we don't need to use it at all.

It's not clear why we need this script at all instead of just referencing the project in XCode

AppCenter/AppCenterTests/OHHTTPStubs/include/HTTPStubs.h
```diff
@@ -0,0 +1,219 @@
+/***********************************************************************************
```
MatkovIvan: It should not be copied

PR #2037, opened 2020-04-22
Add Swift Package Manager manifest file

Jamminroot:
* [ ] Did you add unit tests?

## Description

This PR adds manifest file to bring partial SPM support - only core and analytics modules.

## Related PRs or issues

[AB#79418](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/79418)


Package.swift
```diff
@@ -0,0 +1,62 @@
...
+            dependencies: ["AppCenter"],
+            path: "AppCenterAnalytics/AppCenterAnalytics",
+            exclude: ["Support"],
+            cSettings: [
+                .headerSearchPath("**"),
+                .headerSearchPath("../../AppCenter/AppCenter/**"),
+            ],
+            linkerSettings: [
+                .linkedLibrary("z"),
+                .linkedLibrary("sqlite3"),
```
MatkovIvan, line 55: Analytics module doesn't use sqlite3 directly

Package.swift
```diff
@@ -0,0 +1,62 @@
...
+        .target(
+            name: "AppCenter",
+            dependencies: [],
+            path: "AppCenter/AppCenter",
+            exclude: ["Support"],
+            cSettings: [
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.ios\"", .when(platforms: [.iOS])),
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.macos\"", .when(platforms: [.macOS])),
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.tvos\"", .when(platforms: [.tvOS])),
+                .define("APP_CENTER_C_VERSION", to:"\"3.1.1\""),
```
MatkovIvan, line 30: Do we really need extra quotes here?

Package.swift
```diff
@@ -0,0 +1,62 @@
+// swift-tools-version:5.0
```
MatkovIvan, line 1: Can we omit this? We do not depend on any Swift version, so no need to limit it

AppCenter/AppCenter/AppCenter.modulemap
```diff
@@ -0,0 +1,6 @@
+framework module AppCenter {
```
MatkovIvan, line 1: Can we reuse existing one? I'd keep only one modulemap to avoid confusions

Package.swift
Jamminroot: As per [documentation](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#about-the-swift-tools-version) that line must be first. Not sure if we should omit it

Package.swift
Jamminroot: Muotes are mandatory in define values, yes.

Package.swift
Jamminroot: Fixed

AppCenter/AppCenter/AppCenter.modulemap
Jamminroot: Those were not necessary at all - since we provided include symlinks.

Jamminroot:
Note about `UIKit` and `AppKit` usage for reviewers - those are used in Analytics, so we need those lines as `linkedFrameworks`.

Link to dev:

https://github.com/microsoft/appcenter-sdk-apple/blob/develop/AppCenterAnalytics/AppCenterAnalytics/TransmissionTarget/MSPropertyConfigurator.m#L4-L10

https://github.com/microsoft/appcenter-sdk-apple/blob/develop/AppCenterAnalytics/AppCenterAnalytics/Internals/Util/MSAnalyticsCategory.m#L5-L11

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### App Center
 
 * **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few keys missing namespace.
+* **[Feature]** Partial Swift Package Manager distribution support added (for Core and Analytics modules).
```
annakocheshkova, line 8: To me, it sounds better 
```suggestion
* **[Feature]** Swift Package Manager support for Core and Analytics modules.
```

MatkovIvan: I'd move it outside of this subsection

Jamminroot: I don't think it should be somewhere else but this subsection - It doesn't fit any module's section as well as it fits general.
Updated phrase.

PR #2039, opened 2020-04-24
Swift Package Manager support (Analytics & Crashes)

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Swift Package Manager has become the official package manager for the third-party libraries on all platforms. 

## Related PRs or issues

[AB#74009](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/74009)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### App Center
 
 * **[Improvement]** Use namespaced `NSUserDefaults` keys with the **MSAppCenter** prefix for all the keys set by the SDK. Fixed a few keys missing namespace.
+* **[Feature]** Swift Package Manager support for Core and Analytics modules.
```
iamclement, line 8: We should mention SPM support in Core and Analytics separately.

vsts/internal-version-increment.sh
```diff
@@ -75,3 +75,8 @@ fi
 if [ "$BUILD_SOURCEBRANCHNAME" != "master" ]; then
   sed -i '' 's/\(VERSION_STRING[[:space:]]*=[[:space:]]*\).*/\1'$publish_version'/g' $VERSION_FILENAME
 fi
+
+# Update swift package manager config file
+swift_package_file='Package.swift'
```
jaeklim, line 80: Should it be in env variable? Do we have to use `Package.swift`?

Package.swift
```diff
@@ -0,0 +1,60 @@
...
+        .target(
+            name: "AppCenter",
+            dependencies: [],
+            path: "AppCenter/AppCenter",
+            exclude: ["Support"],
+            cSettings: [
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.ios\"", .when(platforms: [.iOS])),
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.macos\"", .when(platforms: [.macOS])),
+                .define("APP_CENTER_C_NAME", to: "\"appcenter.tvos\"", .when(platforms: [.tvOS])),
+                .define("APP_CENTER_C_VERSION", to:"\"3.1.1\""),
```
jaeklim, line 30: You don't need to keep a real version here. Below script doesn't commit this so it seems like `3.1.1` forever.
Can you just use a variable or some of meaningful string and use this in the script to replace to a real version?

jaeklim: Or should it be the version that we need to update on release?

vsts/internal-version-increment.sh
```diff
@@ -75,3 +75,8 @@ fi
...
+sed -i '' 's/\(define("APP_CENTER_C_VERSION",[[:space:]]*to:*\).*/\1''"\\"'$publish_version'\\""),''/g' $swift_package_file
```
jaeklim, line 81: As mentioned above, if you use something like "SWIFT_PACKAGE_VERSION" instead of a real version, you can simplify this line.
```
.define("APP_CENTER_C_VERSION", to:"\"SWIFT_PACKAGE_VERSION\""),
```
```suggestion
sed -i '' 's/SWIFT_PACKAGE_VERSION/'$publish_version'/g' $swift_package_file
```

AppCenter/AppCenter/include/AppCenter.h
```diff
@@ -0,0 +1 @@
+../AppCenter.h
```
jaeklim, line 1: Is it better to move public headers to `./include/`? Does it work?

jaeklim: Otherwise we have to keep in mind whenever we add a new public header, we have to add this in `include`.

vsts/internal-version-increment.sh
MatkovIvan: No, it's a constant, this name is predefined

Package.swift
MatkovIvan: > Or should it be the version that we need to update on release?

Yes, it must be there. A version, not placeholder.

> Below script doesn't commit this so it seems like 3.1.1 forever.

It's existing script for our current release process, it will be committed during next steps.

vsts/internal-version-increment.sh
MatkovIvan: No, the real version must be stored in this file

AppCenter/AppCenter/include/AppCenter.h
MatkovIvan: > Does it work?

No, it doesn't work well. We store headers in not one folder and reuse some Core headers in other modules. Symlinks simplifies things here.

> Otherwise we have to keep in mind whenever we add a new public header, we have to add this in include as well.

Yes, but we prepared integration tests on CI that should detect this kind of issues.

annakocheshkova:
Verified that ios/macos extensions work, but with a little workaround (see https://github.com/MicrosoftDocs/appcenter-docs-pr/pull/1900)

wlisac:
@annakocheshkova Can you share the workaround for macOS extensions? That link does not appear to be public.

MatkovIvan:
@wlisac 

> If you're integrating App Center via SPM and want to use it in your app's extension target as well, make sure that you provide `DISABLE_DIAMOND_PROBLEM_DIAGNOSTIC=YES` in your configuration. This is necessary to avoid SPM limitations in linking a module to multiple targets.

eljeff:
> @wlisac
> 
> > If you're integrating App Center via SPM and want to use it in your app's extension target as well, make sure that you provide `DISABLE_DIAMOND_PROBLEM_DIAGNOSTIC=YES` in your configuration. This is necessary to avoid SPM limitations in linking a module to multiple targets.

@MatkovIvan How / where do you set this?
Thanks y'all

MatkovIvan:
@eljeff Xcode target's build settings (if it cannot be found, add it with "+" menu) or xcconfig file

eljeff:
Thanks a lot @MatkovIvan . I went to build setting, clicked the "+" button for "Add User-Defined Setting", and there I can paste in "DISABLE_DIAMOND_PROBLEM_DIAGNOSTIC", and set value to "YES"

Thanks again

PR #2069, opened 2020-05-21
Move script lines for updating version in Package.swift into another script

vvechkanov:

* ~~[ ] Has `CHANGELOG.md` been updated?~~
* ~~[ ] Are tests passing locally?~~
* ~~[ ] Did you add unit tests?~~
* ~~[ ] Did you check UI tests on the sample app? They are not executed on CI.~~
* ~~[ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?~~

## Description

Version should be updated manually, for that lines for updating it moved into the script, which is called manually.

### DIFF
--- a/vsts/internal-version-increment.sh
+++ b/vsts/internal-version-increment.sh
@@ -75,8 +75,3 @@ fi
 if [ "$BUILD_SOURCEBRANCHNAME" != "master" ]; then
   sed -i '' 's/\(VERSION_STRING[[:space:]]*=[[:space:]]*\).*/\1'$publish_version'/g' $VERSION_FILENAME
 fi
-
-# Update swift package manager config file
-swift_package_file='Package.swift'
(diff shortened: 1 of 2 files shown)


vsts/internal-version-increment.sh
```diff
@@ -75,8 +75,3 @@ fi
 if [ "$BUILD_SOURCEBRANCHNAME" != "master" ]; then
   sed -i '' 's/\(VERSION_STRING[[:space:]]*=[[:space:]]*\).*/\1'$publish_version'/g' $VERSION_FILENAME
 fi
-
-# Update swift package manager config file
-swift_package_file='Package.swift'
```
MatkovIvan, line 80: It looks like the good place for it is `vsts/post-release-version-bump.sh`, isn't it?

Jamminroot: That would still require for manual check then

Jamminroot: Also, I can't find the pipeline using this script 🤔

PR #2089, opened 2020-05-29
Fix sdk name on catalyst

annakocheshkova:

* [ ] ~Are tests passing locally?~

## Description

We had `appcenter.ios` reported with xcframeworks and `appcenter.macos` reported through SPM package. 
The name of the sdk should be one for each case. As discussed offline, this should most likely be "ios" since the sdk being used in "ios".


AppCenter/AppCenter/Internals/Util/MSUtility.m
```diff
@@ -17,8 +17,13 @@
 } ms_info_t;
 
 // SDK versioning.
+#if TARGET_OS_MACCATALYST
+static ms_info_t appcenter_library_info __attribute__((section("__TEXT,__ms_ios,regular,no_dead_strip"))) = {
```
MatkovIvan, line 21: Let's keep this line out of `#if` to avoid duplication

Package.swift
```diff
@@ -32,6 +32,7 @@ let package = Package(
             exclude: ["Support"],
             cSettings: [
                 .define("APP_CENTER_C_NAME", to: "\"appcenter.ios\"", .when(platforms: [.iOS])),
+                .define("APP_CENTER_C_NAME_CATALYST", to: "\"appcenter.ios\"", .when(platforms: [.macOS])),
```
MatkovIvan, line 35: Let's hold this one. See AB#80703

MatkovIvan: btw do we really need this custom define? I guess it will be easier just define it in header with preprocessor instead of xcfiles and SPM. I do not really like introducing new custom define name

Config/iOS.xcconfig
```diff
@@ -1,4 +1,5 @@
 APPCENTER_SDK_NAME = appcenter.ios
+APPCENTER_SDK_NAME_CATALYST = appcenter.maccatalyst
```
MatkovIvan, line 2: https://github.com/microsoft/appcenter-sdk-apple/pull/2089#discussion_r432598860

Package.swift
iamclement: I think in any case we probably don't need a new define.

Issue #2105, opened 2020-06-18
FormatStackFrame in MSErrorLogFormatter does not always work


izackp:
### **Description**

FormatStackFrame in MSErrorLogFormatter does not always work. Seemlying for high address spaces.

In the app center crash report a lot of frames have the address but don't have the accompanying code. They will appear as '0x + 0'. This is because the binary image isn't found even though it definitely is there.

I believe plreport imageForAddress does not normalize addressed in the same way as appcenter

As a side note, I also believe the baseAddress and pcOffset need to be normalized before building the symbol string. I'm currently doing this and app center seems to resolve symbols just fine.

### **Repro Steps**

1. Recreate / submit a crash on an iOS app 
2. Notice missing frame information in reports despite having the address

### **Details**

1. Which SDK version are you using?
    - Dev Branch - 3.2.0 - June 16th
2. Which OS version did you experience the issue on?
    - iOS Simulator
3. Which Cocoapods version are you using (run `pod --version`)?
    - None
4. What device version did you see this error on?  Were you using an emulator or a physical device?
    - iPad Sim 12.9in 3rd gen 13.3 sek
5. What language are you using?
    - [x ] Objective C
    - [x ] Swift

izackp:
A solution I hacked together that works for me:
`
+ (NSString *)formatStackFrame:(PLCrashReportStackFrameInfo *)frameInfo report:(PLCrashReport *)report {

  /*
   * Base image address containing instrumentation pointer, offset of the IP
   * from that base address, and the associated image name.
   */
  uint64_t baseAddress = 0x0;
  uint64_t pcOffset = 0x0;
  NSString *symbolString = nil;

  uint64_t instructionPointer = frameInfo.instructionPointer;
  uint64_t normalizedPointer = [MSErrorLogFormatter normalizeAddress:instructionPointer is64bit:YES];
  
  PLCrashReportBinaryImageInfo *imageInfo = [report imageForAddress:normalizedPointer];
  for (PLCrashReportBinaryImageInfo *imageInfo_ in [report images]) {
      uint64_t normalizedBaseAddress = [MSErrorLogFormatter normalizeAddress:imageInfo_.imageBaseAddress is64bit:true];
      uint64_t max = (normalizedBaseAddress + imageInfo_.imageSize);
      if (normalizedBaseAddress <= normalizedPointer && normalizedPointer < max) {
          imageInfo = imageInfo_;
          break;
      }
  }
  
[...]

PR #2110, opened 2020-06-23
Fix MSErrorLogFormatter

MatkovIvan:
* [x] Did you add unit tests?

## Description

- SDK was not able to find images on Simulator (null pointers in stack trace) because PAC mask was incorrectly applied on Simulator, but it should be only on `arm64`
- App Center should not try to strip PAC on formatting, PLCR takes care about it on a crash time
- App Center tried to find ARM registers for last selector on macOS (it's not Big Sur yet 😉 )
- Update symbol prefix rules in App Center formatter to avoid warning
    ```
    [AppCenterCrashes] WARNING: +[MSErrorLogFormatter formatStackFrame:report:]/492 Symbol prefix rules are unknown for this OS!
    ```

Additional small fixes:

- ~Remove query the binary images for determine CPU type, it's directly available in `report.machineInfo.processorInfo`~
- Remove is64bit parameter from some methods - it was incorrectly used
- Simplify `isCodeType64bit`
- Fail fast on unarchiving `nil` to fix
    ```
    [AppCenter] ERROR: +[MSUtility unarchiveKeyedData:]/71 Unarchiving NSData failed with error: The data couldn’t be read because it isn’t in the correct format.
    ```
- Verify if wrapper exception data exists to fix
    ```
    [AppCenterCrashes] ERROR: +[MSWrapperExceptionManager loadWrapperExceptionWithBaseFilename:]/113 Could not read exception data stored on disk with file name last_saved_wrapper_exception
    ```

## Related PRs or issues

[AB#81394](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/81394)
#2105, #1162


AppCenterCrashes/AppCenterCrashes/Internals/Util/MSErrorLogFormatter.m
```diff
@@ -197,13 +197,7 @@ @implementation MSErrorLogFormatter
...
   MSAppleErrorLog *errorLog = [MSAppleErrorLog new];
 
-  // Map to Apple-style code type, and mark whether architecture is LP64
-  // (64-bit).
-  NSNumber *codeType = [self extractCodeTypeFromReport:report];
-  BOOL is64bit = [self isCodeType64bit:codeType];
-
-  // errorId – Used for de-duplication in case we sent the same crashreport
-  // twice.
+  // errorId – Used for de-duplication in case we sent the same crashreport stwice.
```
Jamminroot, line 200:
```suggestion
  // errorId – Used for de-duplication in case we sent the same crashreport twice.
```

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSErrorLogFormatter.m
```diff
@@ -529,43 +522,47 @@ + (NSString *)extractExceptionReasonFromReport:(PLCrashReport *)report {
...
-    if (foundSelector == NULL)
+    if (foundSelector == NULL) {
       foundSelector = [[self class] selectorForRegisterWithName:@"r2" ofThread:crashedThread report:report];
-  }
-#endif
+    }
+    break;
+
+  case CPU_TYPE_ARM64:
+    foundSelector = [[self class] selectorForRegisterWithName:@"x1" ofThread:crashedThread report:report];
```
annakocheshkova, line 544: Just a suggestion: we could simplify the code a bit if we put determination of register name under `switch` and avoid calling the same function 4 times.

AppCenterCrashes/AppCenterCrashesTests/MSErrorLogFormatterTests.mm
```diff
@@ -501,8 +501,12 @@ - (void)assertIsCrashProbeReportValidConverted:(NSString *)filename {
   NSArray *images = crashReport.images;
   for (PLCrashReportBinaryImageInfo *image in images) {
     if (image.codeType != nil && image.codeType.typeEncoding == PLCrashReportProcessorTypeEncodingMach) {
-      assertThat(errorLog.primaryArchitectureId, equalTo(@(image.codeType.type)));
-      assertThat(errorLog.architectureVariantId, equalTo(@(image.codeType.subtype)));
+      XCTAssertEqual(errorLog.primaryArchitectureId.unsignedLongLongValue, image.codeType.type,
+                     @"Report: %@, Image: %@", filename, [image.imageName lastPathComponent]);
+      // It might be not equal valid architectures is wider than actual compile targets.
```
annakocheshkova, line 506: Uncomment?

MatkovIvan: No. This is a comment WHY this check is commented/disabled.

Issue #2123, opened 2020-07-02
 MSAnalytics.trackEvent doesn't work from applicationWillTerminate


luci1329:
### **Description**

The application delegate's lifecycle method, applicationWillTerminate has a ~ 5 second time window to execute, according to apple docs, yet the simple event I am trying to log never show's up in the AppCenter event logs.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. log a simple event in the applicationWillTerminate
2. close the app
3. wait for the logs
4. die of waiting

### **Details**

1. Which SDK version are you using?
    - e.g. 2.1.0
2. Which OS version did you experience the issue on?
    - e.g. iOS 13
3. Which Cocoapods version are you using (run `pod --version`)?
    - e.g. 1.6.1
4. What device version did you see this error on?  Were you using an emulator or a physical device?
    - both, emulator and physical device (ipad mini 4)
5. What language are you using?
    - [x ] Objective C
    - [ ] Swift
6. What third party libraries are you using?
7. Please enable verbose logging for your app using `MSAppCenter.setLogLevel(.verbose)` before your call to `MSAppCenter.start(...)` for Swift, or `[MSAppCenter setLogLevel:MSLogLevelVerbose]` before `[MSAppCenter start: ...]` for Objective C and include the logs here: n/a

vvechkanov:
Hi @luci1329!

Thank you for contacting us!

I checked this situation, and I also do not get events from applicationWillTerminate. I'll create an issue for it in the team backlog, and we will contact you when this will be fixed.

luci1329:
Hi @vvechkanov thanks for your quick response

MatkovIvan:
This one should be fixed in 3.3.3. Feel free to comment/reopen if you have more questions or if you run into this problem again

PR #2144, opened 2020-08-07
Fix logs sent inside applicationWillTerminate

annakocheshkova:

* [x] Did you add unit tests?

## Description

Logs sent from `applicationWillTerminate` were not being saved because of `dispatch_async` call. 
To fix this, we add a task to the logsQueue and wait for it inside `applicationWillTerminate` event.

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-apple/issues/2123/

## Misc

[AB#81803](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/81803)

TODO: unit tests

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -214,8 +214,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 }
 
 - (void)applicationWillTerminate:(UIApplication *)application {
-  
-  dispatch_semaphore_wait(self.delayedProcessingSemaphore, 2);
+  dispatch_semaphore_wait(self.delayedProcessingSemaphore, dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC));
   
   // Block logs queue so that it isn't killed before app termination.
   dispatch_async(self.logsDispatchQueue, ^{
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -220,7 +220,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 }
 
 #if !TARGET_OS_OSX
-- (void)applicationWillTerminate:(UIApplication *)application {
+- (void)applicationWillTerminate:(__unused UIApplication *)application {
 
   // Block logs queue so that it isn't killed before app termination.
   dispatch_async(self.logsDispatchQueue, ^{
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -13,7 +13,11 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 
 @implementation MSChannelGroupDefault
 
+#if !TARGET_OS_OSX
+
 @synthesize delayedProcessingSemaphore = _delayedProcessingSemaphore;
+
+#endif
 
 #pragma mark - Initialization
 
@@ -34,7 +38,9 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
     if (ingestion) {
       _ingestion = ingestion;
     }
+#if !TARGET_OS_OSX
     _delayedProcessingSemaphore = dispatch_semaphore_create(0);
+#endif
   }
   return self;
 }
@@ -213,6 +219,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
    */
 }
 
+#if !TARGET_OS_OSX
 - (void)applicationWillTerminate:(UIApplication *)application {
 
   // Block logs queue so that it isn't killed before app termination.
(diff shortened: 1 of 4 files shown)


AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -177,6 +180,17 @@ - (void)channel:(id<MSChannelProtocol>)channel didResumeWithIdentifyingObject:(i
 
 - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
 
+#if !TARGET_OS_OSX
+  if (isEnabled) {
+    [MS_NOTIFICATION_CENTER addObserver:self
+    selector:@selector(applicationWillTerminate:)
```
MatkovIvan, line 186: Formatting

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -199,6 +213,15 @@ - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
    */
 }
 
+- (void)applicationWillTerminate:(UIApplication *)application {
+  dispatch_semaphore_wait(self.delayedProcessingSemaphore, dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC));
```
MatkovIvan, line 217: wait should be after `dispatch_async`. This way it won't work

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefaultPrivate.h
```diff
@@ -18,6 +18,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion;
 
+/**
+* Semaphore for blocking logs queue on applicationWillTerminate.
```
MatkovIvan, line 22: Formatting

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefaultPrivate.h
```diff
@@ -18,6 +18,11 @@ NS_ASSUME_NONNULL_BEGIN
...
+*/
+@property dispatch_semaphore_t delayedProcessingSemaphore;
```
MatkovIvan, line 24: Wrap into `#if !TARGET_OS_OSX`

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -177,6 +186,17 @@ - (void)channel:(id<MSChannelProtocol>)channel didResumeWithIdentifyingObject:(i
 
 - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
 
+#if !TARGET_OS_OSX
+  if (isEnabled) {
+    [MS_NOTIFICATION_CENTER addObserver:self
```
iamclement, line 191: We don't really know when the `applicationWillTerminate` of the developer will be called. He can use the delegate but also the notification as we do. Could be before or after. If it's happening after would the log be lost?

MatkovIvan: yes

annakocheshkova: Could be. One possible solution is to use a special flag inside our callback and handle it later but seems like a bit of overkill. What do you think?

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -61,6 +66,26 @@ - (void)tearDown {
   [super tearDown];
 }
 
+#if !TARGET_OS_OSX
+- (void)testAppIsKilled {
+
+  // If
+  [self.sut setEnabled:YES andDeleteDataOnDisabled:YES];
+  id sut = OCMPartialMock(self.sut);
```
MatkovIvan, line 74: Please explicitly stop mocking at the end

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -61,6 +66,26 @@ - (void)tearDown {
...
+
+  // When
+  [[NSNotificationCenter defaultCenter] postNotificationName:UIApplicationWillTerminateNotification object:sut];
+
+  // Then
+  OCMVerify([sut applicationWillTerminate:OCMOCK_ANY]);
+  XCTAssertNotNil(self.sut.logsDispatchQueue);
+  [self.sut setEnabled:NO andDeleteDataOnDisabled:YES];
+  [[NSNotificationCenter defaultCenter] postNotificationName:UIApplicationWillTerminateNotification object:sut];
+  OCMReject([sut applicationWillTerminate:OCMOCK_ANY]);
```
MatkovIvan, line 84: It should be BEFORE usage

MatkovIvan: Please add `OCMVerifyAll` as well

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
iamclement: Sounds overkill yes. I wonder if we can simply switch to synchronous storage for all logs when the app enters this phase. The app will be wiped out of the memory so we don't need to think about restoring the async storage. We'll need to make sure previous logs are saved.

If it turns out to introduce too much complexity we'll need to postpone the fix.

PR #2147, opened 2020-08-11
Fix for late applicationWillTerminate call

annakocheshkova:

~* [ ] Has `CHANGELOG.md` been updated?~
* [X] Did you add unit tests?

## Description

In case user subscription to applicationWillTerminate event fires later than ours, we still need to store logs synchronously.

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -214,8 +214,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 }
 
 - (void)applicationWillTerminate:(UIApplication *)application {
-  
-  dispatch_semaphore_wait(self.delayedProcessingSemaphore, 2);
+  dispatch_semaphore_wait(self.delayedProcessingSemaphore, dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC));
   
   // Block logs queue so that it isn't killed before app termination.
   dispatch_async(self.logsDispatchQueue, ^{
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#import <libkern/OSAtomic.h>
+
 #import "MSChannelGroupDefault.h"
 #import "AppCenter+Internal.h"
 #import "MSAppCenterIngestion.h"
@@ -13,7 +15,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 
 @implementation MSChannelGroupDefault
 
-static BOOL _hasEnteredApplicationWillTerminate = NO;
+static volatile uint32_t _hasEnteredApplicationWillTerminate;
 
 #if !TARGET_OS_OSX
 
@@ -32,7 +34,7 @@ static BOOL _hasEnteredApplicationWillTerminate = NO;
 
 - (instancetype)initWithIngestion:(nullable MSAppCenterIngestion *)ingestion {
   if ((self = [self init])) {
-    _hasEnteredApplicationWillTerminate = NO;
+    OSAtomicOr32Barrier(0, & _hasEnteredApplicationWillTerminate);
     dispatch_queue_t serialQueue = dispatch_queue_create(kMSLogsDispatchQueue, DISPATCH_QUEUE_SERIAL);
     _logsDispatchQueue = serialQueue;
     _channels = [NSMutableArray<id<MSChannelUnitProtocol>> new];
@@ -49,7 +51,7 @@ static BOOL _hasEnteredApplicationWillTerminate = NO;
 }
 
 + (BOOL)hasEnteredApplicationWillTerminate {
-  return _hasEnteredApplicationWillTerminate;
+  return _hasEnteredApplicationWillTerminate != 0;
 }
 
 - (id<MSChannelUnitProtocol>)addChannelUnitWithConfiguration:(MSChannelUnitConfiguration *)configuration {
@@ -228,9 +230,7 @@ static BOOL _hasEnteredApplicationWillTerminate = NO;
 
 #if !TARGET_OS_OSX
 - (void)applicationWillTerminate:(__unused UIApplication *)application {
-  @synchronized(self) {
-    _hasEnteredApplicationWillTerminate = YES;
-  }
+  OSAtomicOr32Barrier(1, & _hasEnteredApplicationWillTerminate);
 
   // Block logs queue so that it isn't killed before app termination.
   dispatch_async(self.logsDispatchQueue, ^{
--- a/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
@@ -15,7 +15,7 @@ static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQ
 
 @implementation MSChannelGroupDefault
 
-static volatile uint32_t _hasEnteredApplicationWillTerminate;
+static BOOL _applicationWillTerminate = NO;
 
 #if !TARGET_OS_OSX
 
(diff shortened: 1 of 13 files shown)


AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -166,7 +166,19 @@ - (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
         [self checkPendingLogs];
       }
     }
-  });
+  };
+
+  NSNumber *hasEnteredApplicationWillTerminate = [MS_APP_CENTER_USER_DEFAULTS objectForKey:kMSApplicationWillTerminateEnteredKey];
```
MatkovIvan, line 171: User defaults must not be used there

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -222,6 +228,7 @@ - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
 #if !TARGET_OS_OSX
 - (void)applicationWillTerminate:(__unused UIApplication *)application {
 
+  _hasEnteredApplicationWillTerminate = YES;
   // Block logs queue so that it isn't killed before app termination.
```
MatkovIvan, line 232: Empty line before comments

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -166,7 +166,17 @@ - (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
         [self checkPendingLogs];
       }
     }
-  });
+  };
+
+  if ([MSChannelGroupDefault hasEnteredApplicationWillTerminate]) {
```
MatkovIvan, line 171: This is not thread-safe.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -222,6 +228,7 @@ - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
 #if !TARGET_OS_OSX
 - (void)applicationWillTerminate:(__unused UIApplication *)application {
 
+  _hasEnteredApplicationWillTerminate = YES;
```
MatkovIvan, line 231: This is not thread-safe.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -10,6 +10,7 @@
 #import "MSLogDBStorage.h"
 
 static char *const kMSLogsDispatchQueue = "com.microsoft.appcenter.ChannelGroupQueue";
+static BOOL _hasEnteredApplicationWillTerminate = NO;
```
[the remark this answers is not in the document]
annakocheshkova: We do not hold a reference to channel group by its units, so it needs to be static

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -13,6 +15,8 @@
 
 @implementation MSChannelGroupDefault
 
+static volatile uint32_t _hasEnteredApplicationWillTerminate;
```
iamclement, line 18: A more direct form may be more readable?
```suggestion
static volatile uint32_t _applicationWillTerminate;
```

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
```diff
@@ -221,6 +230,7 @@ - (void)setEnabled:(BOOL)isEnabled andDeleteDataOnDisabled:(BOOL)deleteData {
 
 #if !TARGET_OS_OSX
 - (void)applicationWillTerminate:(__unused UIApplication *)application {
+  OSAtomicOr32Barrier(1, & _hasEnteredApplicationWillTerminate);
 
   // Block logs queue so that it isn't killed before app termination.
```
iamclement, line 235: Can't we just set the boolean in the log queue synchronously with dispatch_sync? Trying to simplify because we are introducing quite some complexity for just this scenario. Need to find a tradeoff or postpone the fix.

MatkovIvan: It won't make it thread safe

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefaultPrivate.h
```diff
@@ -5,12 +5,29 @@
...
 
 @property(nonatomic) NSMutableSet<NSString *> *pausedTargetKeys;
 
+#if !TARGET_OS_OSX
+/**
+ * Semaphore for blocking logs queue on applicationWillTerminate.
+ */
+@property dispatch_semaphore_t delayedProcessingSemaphore;
+
+@property BOOL applicationWillTerminateEntered;
```
MatkovIvan, line 22: add atomic explicitly

AppCenter/AppCenterTests/MSChannelUnitDefaultTests.m
```diff
@@ -83,6 +86,32 @@ - (void)tearDown {
 
 #pragma mark - Tests
 
+#if !TARGET_OS_OSX
+- (void)testAppIsKilled {
+
```
MatkovIvan, line 91: Format

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#import "TargetConditionals.h"
+#if !TARGET_OS_OSX
+#import <UIKit/UIKit.h>
+#endif
```
MatkovIvan, line 7: It doesn't look required

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
annakocheshkova: Will be confusing now that it is a property

AppCenter/AppCenterTests/MSChannelGroupDefaultTests.m
MatkovIvan: Not resolved

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -166,8 +177,28 @@ - (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
...
+  (self.applicationWillTerminateEntered ? dispatch_sync : dispatch_async)
+#else
+  dispatch_async
+#endif
+      (self.logsDispatchQueue, storeLogs);
+}
+
+#if !TARGET_OS_OSX
+- (void)applicationWillTerminate:(__unused UIApplication *)application {
+  self.applicationWillTerminateEntered = YES;
```
MatkovIvan, line 193: After thinking - it's still not thread safe. There is no garantee that `dispatch_async` won't be scheduled after that

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
```diff
@@ -166,8 +177,28 @@ - (void)enqueueItem:(id<MSLog>)item flags:(MSFlags)flags {
         [self checkPendingLogs];
       }
     }
+  };
+#if !TARGET_OS_OSX
+
+  // Process synchronously in case applicationWillTerminate was hit.
+  (self.applicationWillTerminateEntered ? dispatch_sync : dispatch_async)
```
MatkovIvan, line 184: Wait. `dispatch_sync` can be called not from the main thread. The execution time is the user's responsibility in this case?

iamclement: Yes we don't control the context in this case it's whatever caller context is running. Need to be careful.

AppCenter/AppCenter/Internals/Channel/MSChannelGroupDefault.m
iamclement: Right because by the time it's running some more logs may be queued asynchronously. Though, we can still set the bool before an empty dispatch sync. In this place we are guaranteed to run in the main thread so a dispatch sync should be safe. Of course unless the code dispatch_sync to the main thread but that would be bad.

AppCenter/AppCenter/Internals/Channel/MSChannelUnitDefault.m
iamclement: The more I am thinking about this the more I am thinking we should postpone this fix. It comes with some risks at this point.

Issue #2189, opened 2020-10-13
Crash in ISO8601DateFormatter


MBayard:
When simultaneously launching multiple Mac audio plugins using the the AppCenter SDK, we get occasional crashes in ISO8601DateFormatter.

Thread 1 Crashed:: Dispatch queue: com.microsoft.appcenter.ChannelGroupQueue
0   libobjc.A.dylib               	0x00007fff7091069d objc_msgSend + 29
1   com.apple.Foundation          	0x00007fff484210d9 -[NSDateFormatter setLocale:] + 149
2   com.dolby.atmos.binaural.settings.AAX	0x0000000129c05a91 +[MSUtility(Date) ISO8601DateFormatter] + 140
3   com.dolby.atmos.binaural.settings.AAX	0x0000000129c0593e +[MSUtility(Date) dateToISO8601:] + 53
4   com.dolby.atmos.binaural.settings.AAX	0x0000000129c0cf00 -[MSAbstractLog serializeToDictionary] + 233
5   com.dolby.atmos.binaural.settings.AAX	0x0000000129bf5c4a -[MSStartServiceLog serializeToDictionary] + 64
6   com.dolby.DolbyAtmosPanner.AAX 	0x000000012c385d8b __49-[MSLogContainer serializeLogWithPrettyPrinting:]_block_invoke + 29
7   com.apple.CoreFoundation      	0x00007fff461b07bb -[__NSArrayM enumerateObjectsWithOptions:usingBlock:] + 219
8   com.dolby.DolbyAtmosPanner.AAX 	0x000000012c385b75 -[MSLogContainer serializeLogWithPrettyPrinting:] + 176

Repro Steps:

Create multiple plugins with the AppCenter SDK, load them into your favorite audio workstation, see intermittent crashes on workstation launch.

Using AppCenter Apple 3.2.0
MacOS 10.14
Objective-C++ interface

russelarms:
Hi! Thanks for reaching out, we're looking at this now!

MBayard:
Thanks.  Attaching another crash log, this time from Ableton Live instead of Pro Tools.

[Live crash log.pdf](https://github.com/microsoft/appcenter-sdk-apple/files/5379957/Live.crash.log.pdf)

AnastasiaKubova:
The fix has been released in [4.0.0](https://github.com/microsoft/appcenter-sdk-apple/releases/tag/4.0.0) version of App Center.

PR #2192, opened 2020-10-15
Fix NSDateFormatter initialization in a concurrent environment

russelarms:

* [x] Did you add unit tests?
* [ ] ~~Did you check UI tests on the sample app? They are not executed on CI.~~

## Description

This PR synchronizes NSDateFormatter's initialization in a concurrent environment. It should fix the issue https://github.com/microsoft/appcenter-sdk-apple/issues/2189/

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-apple/issues/2189/
[AB#83374](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Ivan-Team/Backlog%20Items/?workitem=83374)


AppCenter/AppCenter/Internals/Util/MSACUtility+DatePrivate.h
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Utility class that is used throughout the SDK.
+ * Date part.
+ */
+@interface MSACUtility (DatePrivate)
+
+/**
+ * Method to reset the nsdateformatter singleton when running unit tests only.
+ */
++ (void)resetDateFormatterInstance;
```
MatkovIvan, line 15: No need for separate file, it ccan be declared right in the tests file

AppCenter/AppCenter/Internals/Util/MSACUtility+Date.m
```diff
@@ -1,7 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#import "MSACUtility+Date.h"
+#import "MSACUtility+DatePrivate.h"
+
+/**
+ * Singleton.
```
MatkovIvan, line 7: It's more like cache, not "singleton"

AppCenter/AppCenterTests/MSACUtilityTests.m
```diff
@@ -156,6 +156,42 @@ - (void)testNowInMilliseconds {
...
+  id nsDateFormatter = OCMClassMock([NSDateFormatter class]);
+  OCMStub([nsDateFormatter stringFromDate:OCMOCK_ANY]).andReturn(@"stub");
+  OCMStub([nsDateFormatter alloc]).andReturn(nsDateFormatter);
+
+  // When
+  int dispatchTimes = 10;
+  for (int i = 0; i <= dispatchTimes; i++) {
+    dispatch_async(concurrentQueue, ^{
+      if (i == dispatchTimes) {
+        [expectation fulfill];
```
DmitriyKirakosyan, line 174: Wouldn't we get here before even the first test starts? Code inside the `dispatch_async`'s will be called after we iterate 10 times and `i` becomes `10`, because it's async in another thread. And `[expectation fulfill]` will be called 10 times, because in each dispatch block `i == dispatchTimes` is `true`.

I think we should create a variable, that we would increase in each dispatch block and when that variable is equal to `dispatchTimes` we're calling fulfill (but after the `try` block, not before).

Please correct me if I'm wrong.

russelarms: The block is catching the i value when created, so each dispatch_async will have the corresponding value, e.g. 0,1,2,3, and so on. But since they are concurrent they are not guaranteed to run in any particular order. So, yes, it makes sense to increase the counter inside the dispatch block, I'll fix it.

AppCenter/AppCenterTests/MSACUtilityTests.m
```diff
@@ -156,6 +162,47 @@ - (void)testNowInMilliseconds {
...
+  int dispatchTimes = 10;
+  __block NSObject *lock = [NSObject new];
+  __block int counter = 0;
+  for (int i = 0; i < dispatchTimes; i++) {
+    dispatch_async(concurrentQueue, ^{
+      @try {
+        [MSACUtility dateToISO8601:[NSDate dateWithTimeIntervalSince1970:i]];
+      } @catch (NSException *exception) {
+        XCTFail(@"Expectation Failed with error: %@", exception);
+      } @
```
MatkovIvan, line 185: @?

russelarms: `./clang-format-changed-files.sh` did it for some reason. I haven't noticed.
The interesting thing is that it works this way :)

MatkovIvan (approved):

DmitriyKirakosyan (approved):

AnastasiaKubova (approved):

PR #2261, opened 2021-01-29
Fix exitting function in test app

vvechkanov:

* ~[ ] Has `CHANGELOG.md` been updated?~(These are changes in the test app)
* ~[ ] Did you add unit tests?~ (These are changes in the test app)

## Description

Sasquatch app implements the distributeWillExitApp delegate callback wrongly. It delays the thread for 7 seconds before it allows app to exit. Delay is now moved into the view controller code.

## Related PRs or issues

[AB#84300](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/84300)

## Misc

### DIFF
--- a/Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
@@ -96,7 +96,7 @@ class MSDistributeViewController: UITableViewController, AppCenterProtocol {
       case 1:
         appCenter.showDistributeDisabledAlert()
       case 3:
-        appCenter.closeApp()
+        distributeWillExitApp()
       default: ()
       }
     default: ()
@@ -105,5 +105,18 @@ class MSDistributeViewController: UITableViewController, AppCenterProtocol {
 
   @IBAction func customizedSwitchUpdated(_ sender: UISwitch) {
     UserDefaults.standard.set(sender.isOn, forKey: kSASCustomizedUpdateAlertKey)
+  }
+    
+  func distributeWillExitApp() {
+    print("distributeWillExitApp callback invoked");
+    let alertController = UIAlertController(title: NSLocalizedString("distribute_alert_willExit_title", tableName: "Sasquatch", comment: ""),
+            message: NSLocalizedString("distribute_alert_willExit_message", tableName: "Sasquatch", comment: ""),
+            preferredStyle: .alert)
+
+    // Show alert saying that the app is closing
+    self.present(alertController, animated: true)
+    DispatchQueue.main.asyncAfter(deadline: .now() + 7.0) {
+      self.appCenter.closeApp()
+    }
   }
 }
--- a/Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
+++ b/Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
@@ -95,8 +95,6 @@ class MSDistributeViewController: UITableViewController, AppCenterProtocol {
         }
       case 1:
         appCenter.showDistributeDisabledAlert()
-      case 3:
-        exitWithSimulateUpdate()
       default: ()
       }
     default: ()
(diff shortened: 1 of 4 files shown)


Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
```diff
@@ -106,4 +106,17 @@ class MSDistributeViewController: UITableViewController, AppCenterProtocol {
...
+    
+  func distributeWillExitApp() {
+    print("distributeWillExitApp callback invoked");
+    let alertController = UIAlertController(title: NSLocalizedString("distribute_alert_willExit_title", tableName: "Sasquatch", comment: ""),
+            message: NSLocalizedString("distribute_alert_willExit_message", tableName: "Sasquatch", comment: ""),
+            preferredStyle: .alert)
+
+    // Show alert saying that the app is closing
+    self.present(alertController, animated: true)
+    DispatchQueue.main.asyncAfter(deadline: .now() + 7.0) {
```
AnastasiaKubova, line 118: maybe make sense to move `7.0` value to the constant?

Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -395,18 +395,6 @@ - (void)distributeNoReleaseAvailable:(MSACDistribute *)distribute {
   [alert addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"OK", nil) style:UIAlertActionStyleDefault handler:nil]];
   [self.window.rootViewController presentViewController:alert animated:YES completion:nil];
 }
-
-- (void)distributeWillExitApp:(MSACDistribute *)distribute {
```
MatkovIvan, line 399: Delegate should be still implemented, that's why it's here

MatkovIvan (changes requested):

Sasquatch/SasquatchObjC/AppDelegate.m
DmitriyKirakosyan: The idea is to move `sleep` to view and remove it from here.

DmitriyKirakosyan: To reproduce do this:
- press Simulate App exit on the distribution page (should wait 7 sec and then close the app)
- update the app using standart distribution process (should close app for update without a delay)

vvechkanov: If we want to show an alert inside delegate, then we need to show another update before closing app. So for    **Simulate App exit** we will show another alert in code. It duplicates code a bit but works the necessary way.

Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
```diff
@@ -106,4 +106,18 @@ class MSDistributeViewController: UITableViewController, AppCenterProtocol {
...
+    let UPDATE_SIMUATION_TIMEOUT = 7.0
+    print("distributeWillExitApp callback invoked");
+    let alertController = UIAlertController(title: NSLocalizedString("distribute_alert_willExit_title", tableName: "Sasquatch", comment: ""),
+            message: NSLocalizedString("distribute_alert_willExit_message", tableName: "Sasquatch", comment: ""),
+            preferredStyle: .alert)
+
+    // Show alert saying that the app is closing
+    self.present(alertController, animated: true)
+    DispatchQueue.main.asyncAfter(deadline: .now() + UPDATE_SIMUATION_TIMEOUT) {
+      exit(0)
```
DmitriyKirakosyan, line 120: This is not correct. We should use Distribute API, otherwise this simulation is useless.

Something like showing an alert, then waiting 7 sec and then calling `appCenter.closeApp()` should do the job.

Sasquatch/SasquatchSwift/AppDelegate.swift
```diff
@@ -333,8 +333,6 @@ extension AppDelegate: DistributeDelegate {
       // Show alert saying that the app is closing
       self.window?.rootViewController?.present(alertController, animated: true)
     }
-
-    sleep(7)
```
DmitriyKirakosyan, line 337: The alert should be removed as well. It's useless if we don't block the thread with a delay, because the app is going to be closed right after this callback invocation.

Sasquatch/Sasquatch/ViewControllers/MSDistributeViewController.swift
DmitriyKirakosyan: Actually, it's also not the right way to simulate this.
The correct way would be creating a new distribute listener, or add a flag to AppDelegate.
Alternatively we can delete this simulation. It only exists in ios demo app anyways.
Let's discuss it offline.

MatkovIvan (approved):

DmitriyKirakosyan (approved):

PR #2267, opened 2021-02-08
Fixes build warning by removing duplicate key in dictionary

rennarda:
Just a simple one-liner to fix the build warning  ```warning: duplicate key in dictionary literal``` when building with the ``` [-Wobjc-dictionary-duplicate-keys]``` flag in Xcode 12.5 beta


MatkovIvan (approved):
Thanks for that!

Jamminroot (approved):
Thanks for this submission! :+1:

PR #2301, opened 2021-05-12
Fix building Distribute module on Xcode 12.5

DmitriyKirakosyan:

* ~[ ] Did you check UI tests on the sample app? They are not executed on CI.~
* ~[ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?~

## Description

Xcode 12.5 started to report "-Wpointer-to-int-cast" error while building the Distribute code.
Fixed by suppressing the error.

## Related PRs or issues

[AB#86427](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/86427)

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
@@ -687,7 +687,12 @@ static dispatch_once_t onceToken;
   UIApplication *sharedApp = [MSACUtility sharedApp];
 #pragma clang diagnostic push
 
-// Ignore "cast to smaller integer type 'BOOL' (aka 'signed char') from 'id'" for Xcode 12.5.
+// The performSelector method only supports methods that return nothing or an object,
+// so cast to BOOL produces "-Wpointer-to-int-cast".
+// The reason why it works is to do with the way values are returned by methods.
+// Integer-like values (NSInteger, BOOL and object pointers) are returned in a general purpose register.
+// The compiler will always load the result from the register used for returning `id` values,
+// and then perform any action required by the cast. For integer and boolean values the cast is a no-op.
 #pragma clang diagnostic ignored "-Wpointer-to-int-cast"
 
   return (BOOL)[sharedApp performSelector:@selector(openURL:) withObject:url];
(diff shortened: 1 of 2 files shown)


AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
```diff
@@ -685,7 +685,13 @@ - (nullable NSURL *)buildTokenRequestURLWithAppSecret:(NSString *)appSecret
 
 - (BOOL)openUrlUsingSharedApp:(NSURL *)url {
   UIApplication *sharedApp = [MSACUtility sharedApp];
+#pragma clang diagnostic push
+
+// Ignore "cast to smaller integer type 'BOOL' (aka 'signed char') from 'id'" for Xcode 12.5.
```
MatkovIvan, line 690: Warning was correct here. You cannot just ignore it because of "Xcode 12.5".
The good explanation here: https://stackoverflow.com/a/23902487/1703290
If you want to keep this cast, left the comment WHY it works

MatkovIvan (changes requested):

AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
DmitriyKirakosyan: At first I was planning to convert this to NSInvocation. But then tested with muting this error and it also worked.

This method is called when Distribute is trying to call the iOS Tester App for updates : https://github.com/microsoft/appcenter-sdk-apple/blob/develop/AppCenterDistribute/AppCenterDistribute/MSACDistribute.m#L437
So, it is probably not used as there is no ios tester app afaik.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### App Center Distribute
 
 * **[Fix]** Fix linking framework `AuthenticationServices`.
+* **[Fix]** Fix compiling Distribute module on Xcode 12.5.
```
MatkovIvan, line 17: It was just a warning, right? It will prevent building only with "warning as errors" setting

AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
DmitriyKirakosyan: Added better description, which is directly based on the answer from stackoverflow.

CHANGELOG.md
DmitriyKirakosyan: Yes, but if I get it right we compile it with this flag enabled. See [Config/Global.xcconfig#L154](https://github.com/microsoft/appcenter-sdk-apple/blob/develop/Config/Global.xcconfig#L154)

MatkovIvan:
```suggestion
* **[Fix]** Fix a warning in Distribute module that prevented using SDK as a source code on Xcode 12.5.
```

MatkovIvan (approved):

russelarms (approved):

PR #2307, opened 2021-05-24
Use an available device information from the PLCR crash report

russelarms:

* [x] Did you add unit tests?
* [ ] ~~Did you check UI tests on the sample app? They are not executed on CI.~~

## Description

This PR fixes some cases where the reported application information is incorrect due to a manually changed time on the device.

## Related PRs or issues

[AB#86579](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Mobile%20SDK/Backlog%20Items/?workitem=86579)

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/Internals/Util/MSACErrorLogFormatter.m
+++ b/AppCenterCrashes/AppCenterCrashes/Internals/Util/MSACErrorLogFormatter.m
@@ -738,14 +738,18 @@ static const char *findSEL(const char *imageName, NSString *imageUUID, uint64_t 
   MSACDevice *device = [[MSACDeviceTracker sharedInstance] deviceForTimestamp:timestamp];
   MSACDevice *alteredDevice = [MSACDevice new];
 
-  // Merge PLCR system information with our device information as the PLCR report
+  // Merge PLCR system information with the SDK`s device information as the PLCR report
   // is more relevant in cases when the time on the device has been manually changed.
-  alteredDevice.osVersion = report.systemInfo.operatingSystemVersion ? report.systemInfo.operatingSystemVersion : device.osVersion;
-  alteredDevice.osBuild = report.systemInfo.operatingSystemBuild ? report.systemInfo.operatingSystemBuild : device.osBuild;
-  alteredDevice.model = report.machineInfo.modelName ? report.machineInfo.modelName : device.model;
-  alteredDevice.appBuild = report.applicationInfo.applicationVersion ? report.applicationInfo.applicationVersion : device.appBuild;
-  alteredDevice.appVersion = report.applicationInfo.applicationMarketingVersion ? report.applicationInfo.applicationMarketingVersion : device.appVersion;
-  alteredDevice.appNamespace = report.applicationInfo.applicationIdentifier ? report.applicationInfo.applicationIdentifier : device.appNamespace;
+  // These fields are expected not to be empty:
+  // https://github.com/microsoft/plcrashreporter/blob/b7b88ee14bbc25ce408ae05464cb6f1cdd747948/Source/PLCrashReport.m#L135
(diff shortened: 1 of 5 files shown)


russelarms:
Tested with wrapper SDKs, works fine.

AppCenterCrashes/AppCenterCrashesTests/MSACErrorLogFormatterTests.mm
```diff
@@ -473,6 +471,79 @@ - (void)testErrorLogFromCrashReportWithWrapper {
...
+  device.wrapperRuntimeVersion = @"13.14";
+  device.liveUpdateReleaseLabel = @"Release Label";
+  device.liveUpdateDeploymentKey = @"Deployment Key";
+  device.liveUpdatePackageHash = @"Package Hash";
+
+  // When
+  MSACAppleErrorLog *errorLog = [MSACErrorLogFormatter errorLogFromCrashReport:report];
+
+  // Then
+  XCTAssertNotNil(errorLog.device.appBuild);
```
DmitriyKirakosyan, line 516: We are checking these values for notNil and notEqual to the original device field. But we don't check if the value equals to the value from crash log.
Based on the test name I would expect this test to verify if the device fields were set to exact values from the crash log.

AppCenterCrashes/AppCenterCrashesTests/MSACErrorLogFormatterTests.mm
```diff
@@ -97,7 +97,6 @@ - (void)testCreateErrorReport {
   XCTAssertEqual([errorReport.appErrorTime timeIntervalSince1970], [crashReport.systemInfo.timestamp timeIntervalSince1970] + 0.999);
   assertThat(errorReport.appStartTime, equalTo(crashReport.processInfo.processStartTime));
 
-  XCTAssertEqualObjects(errorReport.device, self.deviceMock);
```
DmitriyKirakosyan, line 100: Is it worth writing a function that would check the devices for being partly equal?
I guess you removed these lines because device is not equal now since it takes some data from crash report. But half of the object still should be equal.

AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
```diff
@@ -1304,6 +1304,13 @@ - (void)handleUserConfirmation:(MSACUserConfirmation)userConfirmation {
     id<MSACCrashesDelegate> delegate = self.delegate;
     if ([delegate respondsToSelector:@selector(attachmentsWithCrashes:forErrorReport:)]) {
       attachments = [delegate attachmentsWithCrashes:self forErrorReport:report];
+
+      // Use the device information from the error log, otherwise the current device information will be used.
+      for (MSACErrorAttachmentLog *attachment in attachments) {
+        if (attachment && !attachment.device) {
```
DmitriyKirakosyan, line 1310:
```suggestion
        if (attachment != nil && attachment.device == nil) {
```

AppCenterCrashes/AppCenterCrashes/Internals/Util/MSACErrorLogFormatter.m
```diff
@@ -736,6 +734,39 @@ + (MSACBinaryImageType)imageTypeForImagePath:(NSString *)imagePath processPath:(
 
 #pragma mark - Helpers
 
++ (MSACDevice *)deviceForTimestamp:(NSDate *)timestamp crashReport:(PLCrashReport *)report {
+  MSACDevice *device = [[MSACDeviceTracker sharedInstance] deviceForTimestamp:timestamp];
+  MSACDevice *alteredDevice = [MSACDevice new];
+
+  // Merge PLCR system information with our device information as the PLCR report
+  // is more relevant in cases when the time on the device has been manually changed.
+  alteredDevice.osVersion = report.systemInfo.operatingSystemVersion;
```
DmitriyKirakosyan, line 743: Shouldn't we fallback to the value from device tracker if crash reporter's field is empty or nil?

DmitriyKirakosyan (approved):

PR #2322, opened 2021-06-21
[demo-app-ios] Removing No Updates Available alert

AnatolyPristensky:
~~* [ ] Has `CHANGELOG.md` been updated?~~
~~* [ ] Are tests passing locally?~~
~~* [ ] Did you add unit tests?~~

## Description

Removing No Updates Available alert on app startup.

## Related PRs or issues

Please, correct this field.

### DIFF
--- a/Sasquatch/SasquatchObjC/AppDelegate.m
+++ b/Sasquatch/SasquatchObjC/AppDelegate.m
@@ -393,11 +393,6 @@ enum StartupMode { APPCENTER, ONECOLLECTOR, BOTH, NONE, SKIP };
 
 - (void)distributeNoReleaseAvailable:(MSACDistribute *)distribute {
   NSLog(@"distributeNoReleaseAvailable invoked");
-  UIAlertController *alert = [UIAlertController alertControllerWithTitle:nil
-                                                                 message:NSLocalizedString(@"No updates available", nil)
-                                                          preferredStyle:UIAlertControllerStyleAlert];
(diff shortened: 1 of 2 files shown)


Sasquatch/SasquatchObjC/AppDelegate.m
```diff
@@ -393,11 +393,6 @@ - (BOOL)distribute:(MSACDistribute *)distribute releaseAvailableWithDetails:(MSA
 
 - (void)distributeNoReleaseAvailable:(MSACDistribute *)distribute {
   NSLog(@"distributeNoReleaseAvailable invoked");
-  UIAlertController *alert = [UIAlertController alertControllerWithTitle:nil
```
AnastasiaKubova, line 396: Why are they removed? How these dialogs are displayed now?

AnastasiaKubova (changes requested):

Sasquatch/SasquatchObjC/AppDelegate.m
AnatolyPristensky: These dialogs have been completely removed due to their unnecessariness. Also they were uninformative and conflicting with "Crash reports" alerts.
Now they appearing only in logs.
@DmitriyKirakosyan approved this approach.

AnastasiaKubova: This dialog is used to check that NoReleaseAvailable callback works, because Demo apps are tested without checking logs.

AnastasiaKubova: Also, what about the update dialog? Does it affect the sending of logs?

Issue #2140, opened 2020-08-02
Support SwiftPM command line


shpakovski:
Hello, I’m using your SDK as a Swift Package and it prevents `swift test` from working with my own package. Steps to reproduce:

```txt
git clone https://github.com/microsoft/appcenter-sdk-apple
cd appcenter-sdk-apple
swift test
```

```
...
/Users/vadim/Projects/appcenter-sdk-apple/AppCenterCrashes/AppCenterCrashes/Model/MSErrorAttachmentLog.m:4:9: fatal error: 'MSCrashesUtil.h' file not found
#import "MSCrashesUtil.h"
        ^~~~~~~~~~~~~~~~~
clang: warning: -framework Foundation: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -framework Foundation: 'linker' input unused [-Wunused-command-line-argument]
/Users/vadim/Projects/appcenter-sdk-apple/AppCenterCrashes/AppCenterCrashes/Internals/Util/MSErrorLogFormatter.m:49:9: fatal error: 'MSAppCenterInternal.h' file not found
#import "MSAppCenterInternal.h"
        ^~~~~~~~~~~~~~~~~~~~~~~
/Users/vadim/Projects/appcenter-sdk-apple/AppCenterCrashes/AppCenterCrashes/Internals/Util/MSApplicationForwarder.m:6:9: fatal error: 'MSAppCenterInternal.h' file not found
#import "MSAppCenterInternal.h"
        ^~~~~~~~~~~~~~~~~~~~~~~
/Users/vadim/Projects/appcenter-sdk-apple/AppCenterCrashes/AppCenterCrashes/MSCrashes.mm:10:9: fatal error: 'MSAppCenterInternal.h' file not found
#import "MSAppCenterInternal.h"
        ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
...
```

1. Which SDK version are you using?
    - 3.3.2
2. Which OS version did you experience the issue on?
    - macOS 11 Beta 3, Xcode 12 Beta 3

Thanks!

Jamminroot:
Hey, @shpakovski, thanks for bringing this to our sight. We've investigated this problem, did reproduce it on our side and the problem seem to be a bit deeper than it might at first look.
In short: we currently don't support SPM build through console.
In details:
[...]

PR #2397, opened 2022-02-01
Propose a workaround for SwiftPM's lack of recursive header search paths

tonyarnold:
## Description

This PR proposes a workaround for the issue described in #2140. 

## Related PRs or issues

- #2140

## Misc

Ideally the list of header search paths could be scanned from the filesystem, however I think there are potentially bigger reliability issues with doing that than with what I've proposed.

### DIFF
--- a/Package@swift-5.3.swift
+++ b/Package@swift-5.3.swift
@@ -35,11 +35,14 @@ let package = Package(
             name: "AppCenter",
             path: "AppCenter/AppCenter",
             exclude: ["Support"],
-            cSettings: [
-                .define("APP_CENTER_C_VERSION", to:"\"4.4.2\""),
-                .define("APP_CENTER_C_BUILD", to:"\"1\""),
-                .headerSearchPath("**"),
-            ],
+            cSettings: {
+                var settings: [CSetting] = [
+                    .define("APP_CENTER_C_VERSION", to: "\"4.4.2\""),
+                    .define("APP_CENTER_C_BUILD", to: "\"1\"")
+                ]
+                settings.append(contentsOf: cHeaderSearchPaths)
+                return settings
+            }(),
             linkerSettings: [
                 .linkedLibrary("z"),
                 .linkedLibrary("sqlite3"),
@@ -55,10 +58,7 @@ let package = Package(
             dependencies: ["AppCenter"],
             path: "AppCenterAnalytics/AppCenterAnalytics",
             exclude: ["Support"],
-            cSettings: [
-                .headerSearchPath("**"),
-                .headerSearchPath("../../AppCenter/AppCenter/**"),
-            ],
+            cSettings: cHeaderSearchPaths,
             linkerSettings: [
                 .linkedFramework("Foundation"),
                 .linkedFramework("UIKit", .when(platforms: [.iOS, .tvOS])),
@@ -73,10 +73,7 @@ let package = Package(
             ],
             path: "AppCenterCrashes/AppCenterCrashes",
             exclude: ["Support", "Internals/MSACCrashesBufferedLog.hpp"],
-            cSettings: [
-                .headerSearchPath("**"),
-                .headerSearchPath("../../AppCenter/AppCenter/**"),
-            ],
+            cSettings: cHeaderSearchPaths,
             linkerSettings: [
                 .linkedFramework("Foundation"),
                 .linkedFramework("UIKit", .when(platforms: [.iOS, .tvOS])),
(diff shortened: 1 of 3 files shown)


AnastasiaKubova:
Hi! Thanks for your contribution PR! 
I will check it and back to you with the results.

Package@swift-5.3.swift
```diff
@@ -35,11 +35,14 @@ let package = Package(
             name: "AppCenter",
             path: "AppCenter/AppCenter",
             exclude: ["Support"],
-            cSettings: [
-                .define("APP_CENTER_C_VERSION", to:"\"4.4.2\""),
-                .define("APP_CENTER_C_BUILD", to:"\"1\""),
-                .headerSearchPath("**"),
-            ],
+            cSettings: {
```
AnastasiaKubova, line 38: Hi! I checked your changes, but it doesn't work for me.
When I'm trying to call the `swift test` command (from the feature description)  I get the next error:
```
error: manifest parse error(s):
Unknown error parsing manifest for appcenter-sdk-apple 
```

tonyarnold: Oh, right, I need to update that. That was `swift test` in local projects that use AppCenter SDK. 

To test this for the project, you would need to specify `swift build`.

tonyarnold: My apologies, there was an error in the committed code. 

Could you please try running:

```sh
swift build --product AppCenterAnalytics
swift build --product AppCenterCrashes
```

These should work. I'm looking to see if I can make plain `swift build` work, but there's iOS code mixed in via the AppCenterDistribute target that isn't being excluded from the Swift package properly.

AnastasiaKubova: Thanks for the update! 
Now it works for me with the `swift build` command, but fails with the `swift test` with error:
```
fatal error: 'UIKit/UIScreen.h' file not found
#import <UIKit/UIScreen.h>
        ^~~~~~~~~~~~~~~~~~
```

We will discuss these changes with the team and come back with the results.

tonyarnold: Thank you. I realise it's not an ideal solution to the problem, but AppCenter not working with SwiftPM outside of Xcode is an issue that's going to continue to grow as people migrate their projects to use more Swift packages.

DmitriyKirakosyan:
Hi @tonyarnold, I tested your solution and it does work fine for `AppCenterAnalytics` and `AppCenterCrashes` products. It doesn't work for `AppCenterDistribute` though. It is not building because of the UIKit dep. Is there any reason for adding Distribute dependency for a Swift Package project? If there is no such need, I guess we can ignore the broken Distribute product.

As an alternative solution, I was trying to add header paths recursively using a helper function with FileManager etc, however the solution wasn't elegant, so specifying paths explicitly looks better.

There is an issue in your changes though. If you try to integrate AppCenter SDK via SPM from your branch and add all modules, it would not compile because you missed some header paths for Distribute module.
I think we need to add "**" to the header path list to make Xcode happy when integrating SDK to an iOS project.
Could you please fix that?

tonyarnold:
Thanks, @DmitriyKirakosyan. I've added "**" as a header search path in 09e5806231aba77ce6533a0a4ac7c02a6a5968a2. 

> Is there any reason for adding Distribute dependency for a Swift Package project? 

Not for macOS projects - Distribute is iOS-only, right? It should work when targeting iOS. 

>As an alternative solution, I was trying to add header paths recursively using a helper function with FileManager etc, however the solution wasn't elegant, so specifying paths explicitly looks better.

I tried the same, but yeah, I found the same outcome - it got messy quickly. This isn't great, but perhaps someone from Microsoft could file an issue on https://bugs.swift.org/ about supporting recursive header search paths in Swift Package Manager? (it's going to hold more weight than if I did it!)

DmitriyKirakosyan:
@tonyarnold Thanks for the update.

> Not for macOS projects - Distribute is iOS-only, right? It should work when targeting iOS.

Yes, but what about iOS package projects? Probably worth mentioning that AppCenterDistribute product is not building via swift cli due to its iOS dependency.

> This isn't great, but perhaps someone from Microsoft could file an issue on https://bugs.swift.org/ about supporting recursive header search paths in Swift Package Manager?

I'm actually surprised no one has filed such feature request yet. I created one here: https://bugs.swift.org/browse/SR-15851.
Hopefully they fix it and we will be able to get rid of this workaround.

tonyarnold:
I've updated the changelog to reflect that only the Analytics and Crashes targets will build via `swift test`, etc.

DmitriyKirakosyan (approved):

DmitriyKirakosyan:
@tonyarnold merged. Thank you for your contribution!

tonyarnold:
Thanks for accepting the patch! Hopefully Apple addresses the recursive header search limitations in SwiftPM, so that we can drop this workaround. 

Is there a release of this project planned anytime soon? It would be good to know when I can stop using my workarounds locally to get this working.

DmitriyKirakosyan:
Next release is planned at the end of this quarter. Meantime, you can just use `develop` branch.

Issue #2382, opened 2021-12-15
Details of Swift errors not reported?


arielelkin:
### **Description**

I have a module in my app that can throw the following errors:

```swift
public enum MyError: Error {
    case errorOne
    case errorTwo
    case errorThree
```

Every time an error is thrown, I track it via App Center thus:

```swift
Crashes.trackError(error, properties: properties, attachments: nil)
```

But this results in the following report:

![image](https://user-images.githubusercontent.com/1756909/146199202-764bd755-2d45-40f7-b2cb-8b278b7f2917.png)

There were 95 reports of `MyError` being thrown, but no details on which specific error it was. How can I find those details? 

### **Details**

1. Which SDK version are you using?
    - e.g. 4.3.0
2. Which OS version did you experience the issue on?
    - e.g. macOS 12
3. Which Xcode version did you build the app with?
    - e.g. Xcode 13.1
5. Which Cocoapods version are you using (run `pod --version`)?
    - e.g. 1.10.1
6. What device version did you see this error on?  Were you using an emulator or a physical device?
    - e.g. MacBook Pro
7. What language are you using?
    - [x] Swift

arielelkin:
Hello @DmitriyKirakosyan 

I've tried to crash my app by clicking a button that calls this:

```swift
func buttonClicked()
    func makeCrash() throws {
        throw MyError.errorOne
    }
    try! makeCrash()
}
```

I click the button and it crashes the app as expected. But when I run it again `Crashes.hasCrashedInLastSession` is false, and `Crashes.lastSessionCrashReport` is nil. Could it be because I have the debugger attached? 

How do you suggest I crash the app with these errors?

DmitriyKirakosyan:
HI @arielelkin , yes the debugger should not be attached, otherwise the crash reporter will not able to catch the crash.
1. Turn ON verbose logging : `AppCenter.logLevel = .verbose`
2. Disable debugger (`Edit scheme -> Run -> Turn off Debugger`) and run the app
3. Crash the app
4. Enable debugger again and run the app
5. Log with the crash should be displayed in output and sent to the App Center Portal

arielelkin:
OK, I followed those steps and got this stack trace, which seems symbolicated:

```
0x00007ff82ce84df4.$sSTsE7forEachyyy7ElementQzKXEKF libswiftCore
0x00007ff81f75cad8._dispatch_call_block_and_release libdispatch
[...]

PR #2401, opened 2022-02-22
Add null check for AppCenter.trackError

AnastasiaKubova:

## Description

[#2382](https://github.com/microsoft/appcenter/issues/2382/)

## Related PRs or issues

[AB#91298](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/91298)

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
@@ -209,7 +209,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSACC
 
 #pragma mark - Public Methods
 
-+ (NSString *)trackError:(NSError *)error
++ (NSString *)trackError:(NSError *_Nonnull)error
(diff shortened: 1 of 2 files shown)


AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
```diff
@@ -209,15 +209,15 @@ @implementation MSACCrashes
 
 #pragma mark - Public Methods
 
-+ (NSString *)trackError:(NSError *)error
++ (NSString *)trackError:(NSError *_Nonnull)error
```
DmitriyKirakosyan, line 212: It doesn't restrict you from passing `nil` here in objective-c, but only generates a warning if I pass `nil`.
Check https://developer.apple.com/swift/blog/?id=25.
I think we should check with `if` as well, shouldn't we?

DmitriyKirakosyan: Decided to avoid adding additional conditions. Annotation should be enough.

DmitriyKirakosyan (approved):

PR #2441, opened 2022-10-27
Mark Validate Project Settings script phases as always out of date

ksuther:
## Description

This silences warnings in Xcode 14 about build scripts running every time because input and output files are defined. The Verify No Build Settings scripts now explicitly run each build (before it was happening implicitly and showing the warning).

## Misc

Another option would be to specify input and output files so that these scripts only run when necessary.

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes.xcodeproj/project.pbxproj
+++ b/AppCenterCrashes/AppCenterCrashes.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 46;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
@@ -2279,6 +2279,7 @@
 		};
 		C2B25A21230F03A200511D49 /* Verify No Build Settings */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
 			buildActionMask = 2147483647;
 			files = (
 			);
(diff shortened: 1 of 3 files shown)


AppCenterCrashes/AppCenterCrashes.xcodeproj/project.pbxproj
```diff
@@ -444,6 +444,13 @@
 			remoteGlobalIDString = C9A91F5D230BFDB70068070D;
 			remoteInfo = "AppCenterCrashes macOS Framework";
 		};
+		371668A8290B177B006A1E12 /* PBXContainerItemProxy */ = {
```
AnatolyPristensky, line 447: Why do we need these changes?

AppCenterCrashes/AppCenterCrashes.xcodeproj/project.pbxproj
```diff
@@ -1832,10 +1847,10 @@
 			remoteRef = C24C9D022472B3C30072C35D /* PBXContainerItemProxy */;
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
-		C24C9D072472B3C30072C35D /* Tests tvOS.xctest */ = {
+		C24C9D072472B3C30072C35D /* CrashReporter tvOS Tests.xctest */ = {
```
AnatolyPristensky, line 1850: Why do we need this renaming?

AnatolyPristensky:
Hello @ksuther and thank you for your contribution!
Left some comments, it's not clear why do we need a new targets and renaming.

ksuther:
Xcode made these changes automatically. I tend to not fight Xcode when it wants to make these changes, as it usually means any other change to the project will still result in these automatic changes being applied. If you'd prefer I can back them out.

MikhailSuendukov:
Hi @ksuther, yes, could you please revert these redundant changes?

ksuther:
Force pushed with just the necessary changes.

DmitriyKirakosyan (approved):

PR #2449, opened 2022-12-12
Synchronize self.delegates instead of self

ternovtsi:

* [ ] Did you add unit tests?

## Description
Synchronize self.delegates in MSACChannelGroupDefault the same like it's done in MSACChannelUnitDefault.m.

## Related PRs or issues
[AB#89553](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/89553)

### DIFF
--- a/AppCenter/AppCenter/Internals/Channel/MSACChannelGroupDefault.m
+++ b/AppCenter/AppCenter/Internals/Channel/MSACChannelGroupDefault.m
@@ -75,20 +75,20 @@ static char *const kMSACLogsDispatchQueue = "com.microsoft.appcenter.ChannelGrou
 #pragma mark - Delegate
 
 - (void)addDelegate:(id<MSACChannelDelegate>)delegate {
-  @synchronized(self) {
+  @synchronized(self.delegates) {
     [self.delegates addObject:delegate];
   }
 }
 
 - (void)removeDelegate:(id<MSACChannelDelegate>)delegate {
-  @synchronized(self) {
+  @synchronized(self.delegates) {
     [self.delegates removeObject:delegate];
   }
 }
 
 - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSACChannelDelegate> delegate))block {
   NSArray *synchronizedDelegates;
-  @synchronized(self) {
+  @synchronized(self.delegates) {
 
     // Don't execute the block while locking; it might be locking too and deadlock ourselves.
     synchronizedDelegates = [self.delegates allObjects];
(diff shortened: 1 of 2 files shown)


AppCenter/AppCenter/Internals/Channel/MSACChannelGroupDefault.m
```diff
@@ -87,17 +87,13 @@ - (void)removeDelegate:(id<MSACChannelDelegate>)delegate {
...
-  NSArray *synchronizedDelegates;
-  @synchronized(self) {
-
-    // Don't execute the block while locking; it might be locking too and deadlock ourselves.
-    synchronizedDelegates = [self.delegates allObjects];
-  }
-  for (id<MSACChannelDelegate> delegate in synchronizedDelegates) {
-    if ([delegate respondsToSelector:selector]) {
-      block(delegate);
+    @synchronized(self.delegates) {
```
DmitriyKirakosyan, line 90: You can't really do this. Check the comment. If `block(delegate)` calls this instance, it will end up with a deadlock.

ternovtsi: There is MSACDeadLockTests i mentioned in description

AppCenter/AppCenter/Internals/Channel/MSACChannelGroupDefault.m
```diff
@@ -87,13 +87,8 @@ - (void)removeDelegate:(id<MSACChannelDelegate>)delegate {
...
 
 - (void)enumerateDelegatesForSelector:(SEL)selector withBlock:(void (^)(id<MSACChannelDelegate> delegate))block {
-  NSArray *synchronizedDelegates;
-  @synchronized(self) {
-
-    // Don't execute the block while locking; it might be locking too and deadlock ourselves.
-    synchronizedDelegates = [self.delegates allObjects];
-  }
-  for (id<MSACChannelDelegate> delegate in synchronizedDelegates) {
+  NSArray *delegatesCopy = [self.delegates.allObjects copy];
```
DmitriyKirakosyan, line 90: Is it safe to remove `@synchronized(self)`? What would happen if we call `enumerateDelegatesForSelector` at the same time from multiple threads?

ternovtsi: I don't think we need it here. I simple don't understand why should we assign it under synchronized. If we call enumerateDelegatesForSelector at the same time from multiple threads nothing bad should happen i think, maybe im wrong

DmitriyKirakosyan (approved):

PR #2461, opened 2023-01-25
Fix NSLog congestion on Apple's Framework Thread

shadyabarada:

## Description

Thus AppCenter Logs are being dispatched to different threads that may not be executed consecutively and they are being congested in the thread pool. 

The reason behind the above issue is that the NSLog is currently being dispatched to a global queue provided by apple framework

## Related PRs or issues

[AB#98398](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/98398)

## Misc


AppCenter/AppCenter/MSACNSLogQueueManager.m
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "MSACNSLogQueueManager.h"
+#import <dispatch/dispatch.h>
+
+@implementation MSACNSLogQueueManager
```
DmitriyKirakosyan, line 7: Why do we need a new singleton class for keeping one serialQueue which is only used in `MSACLogger`?
Can we just add a static variable to `MSACLogger`, eg `static dispatch_queue_t loggerDispatchQueue`. And instantiate in `+ (void)initialize`.

AppCenter/AppCenter/MSACNSLogQueueManager.h
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import <Foundation/Foundation.h>
+
+@interface MSACNSLogQueueManager : NSObject
+
++ (MSACNSLogQueueManager *)sharedManager;
+@property(nonatomic, strong) dispatch_queue_t mySerialQueue;
```
DmitriyKirakosyan, line 9: `mySerialQueue` -> `loggerDispatchQueue`

AppCenter/AppCenter/MSACLogger.m
```diff
@@ -78,12 +79,12 @@ + (void)logMessage:(MSACLogMessageProvider)messageProvider
               file:(const char *)file
           function:(const char *)function
               line:(uint)line {
-    
-    dispatch_async(dispatch_get_global_queue( DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^(void){
-        if (currentLogHandler) {
-            currentLogHandler(messageProvider, loglevel, tag, file, function, line);
-        }
-    });
+
```
DmitriyKirakosyan, line 82: let's remove empty line

AppCenter/AppCenter/MSACLogger.m
```diff
@@ -78,12 +79,12 @@ + (void)logMessage:(MSACLogMessageProvider)messageProvider
...
+  dispatch_async([[MSACNSLogQueueManager sharedManager] mySerialQueue], ^{
```
DmitriyKirakosyan, line 83: Better to put `dispatch_async` under the condition `if (currentLogHandler)` to avoid dispatching if there is no log handler.

AppCenter/AppCenter/MSACNSLogQueueManager.m
shadyabarada: Sure we can have it as a static variable on the class level, but I went with the singleton class just in case we later had to implement NLogging in different classes/modules to use the same serial queue instead of having to create another serial queue with the same purpose.

DmitriyKirakosyan: There shouldn't be any places using NSLog other than MSACLogger. All logging should go through MSACLogger. The idea is to being able to replace the log handler which is NSLog by default. There is a public API for that `AppCenter.logHandler`. So if this dispatcher is only for NSLog, I believe it should be in MSACLogger. 
If for some special reason we would need to implement another logger or use NSLog in other places, it wouldn't be difficult to create a class for the dispatch queue.
For now, I don't think creating code just in case it may be used is a good idea.


Btw, I noticed we are using NSLog instead of MSACLogger in one place:
https://github.com/microsoft/appcenter-sdk-apple/blob/0361dbd7c98e29af468fd36db1441826ef55aa39/AppCenter/AppCenter/MSACAppCenter.m#L224

You can fix it in your PR.

thewulf7 (approved):

PR #2470, opened 2023-03-07
Always use isDirectory parameter to improve performace

DmitriyKirakosyan:
* ~[ ] Did you check UI tests on the sample app? They are not executed on CI.~

## Description

Using `[NSURL URLByAppendingPathComponent:]` may cause performance issues if we don't specify `isDirectory`.
Quote from apple docs:

> If the receiver is a file URL and pathComponent does not end with a trailing slash, this method may read file metadata to determine whether the resulting path is a directory. This is done synchronously, and may have significant performance costs if the receiver is a location on a network mounted filesystem. You can instead call the [URLByAppendingPathComponent:isDirectory:](https://developer.apple.com/documentation/foundation/nsurl/1413953-urlbyappendingpathcomponent) method if you know whether the resulting path is a directory to avoid this file metadata operation.

This PR adds all missing `isDirectory` arguments.

~~Note, that `deleteItemForPathComponent` is used for both files and directories. However, I specify  `isDirectory:YES` assuming it would not break anything. Alternatively I would need to add an additional parameter to this method, which affects multiple parts of the code, so I would better to avoid that.~~

## Related PRs or issues

[AB#98782](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/98782)

### DIFF
--- a/AppCenter/AppCenter/Internals/Util/MSACUtility+File.m
+++ b/AppCenter/AppCenter/Internals/Util/MSACUtility+File.m
@@ -25,7 +25,7 @@ static NSString *const kMSACAppCenterBundleIdentifier = @"com.microsoft.appcente
                       forceOverwrite:(BOOL)forceOverwrite {
   @synchronized(self) {
     if (filePathComponent) {
-      NSURL *fileURL = [[self appCenterDirectoryURL] URLByAppendingPathComponent:filePathComponent];
+      NSURL *fileURL = [[self appCenterDirectoryURL] URLByAppendingPathComponent:filePathComponent isDirectory:NO];
 
       // Check if item already exists. We need to check this as writeToURL:atomically: can override an existing file.
       if (!forceOverwrite && [fileURL checkResourceIsReachableAndReturnError:nil]) {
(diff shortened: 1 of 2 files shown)


Priskorn (approved):

physX23 (approved):
👍

MikhailSuendukov:
What about [that](https://github.com/microsoft/appcenter-sdk-apple/pull/2470/files#diff-26c0cc207e95d58391fe5a4deb18292c10477be4efc3d7741fc386ff6b9f2c40R54) line? This is universal function?

DmitriyKirakosyan:
> What about [that](https://github.com/microsoft/appcenter-sdk-apple/pull/2470/files#diff-26c0cc207e95d58391fe5a4deb18292c10477be4efc3d7741fc386ff6b9f2c40R54) line? This is universal function?

Yes, actually I want to leave it as it is. This function is used for both files and directories. Initially I used `isDirectory:YES` but then reverted the changes.

AnatolyPristensky (approved):

Issue #2497, opened 2023-08-29
Spawning lots of short lived threads when initialising `MSACCrashes` instance


vmagaziy:
### **Description**

When `MSACCrashes` instance is created, `ms_crashes_log_buffer_size` buffer files are created in `- [MSACCrashes setupLogBuffer]` via `dispatch_group_async` and `+ [MSACUtility createFileAtPathComponent:withData:atomically:forceOverwrite:]`. The issue is that `MSACUtility` synchronises its file creation and so it leads to spawning (of in the worst case scenario `ms_crashes_log_buffer_size`) short lived threads.

Provided the nature of the SDK, the mentioned instance is created during the process bootstrapping, and creation of these threads puts an extra pressure on the runtime which may lead to crashes.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Create any iOS project and add the AppCenter SDK (for instance using Swift Package Manager)
2. Initiate creation of the service for crashes using the following line:
```
AppCenterCrashes.AppCenter.start(withAppSecret: "12345", services: [Crashes.self])
```
3. Make breakpoint on a statement right before that and observe a number of threads

<img width="500" alt="Screenshot 2023-08-29 at 9 32 46 PM" src="https://github.com/microsoft/appcenter-sdk-apple/assets/1112653/531b3de3-fd2f-4086-af1f-feee06a1d2bc">

or alternatively create a breakpoint for `start_wqthread` as below and observe the burst of thread creations:

<img width="300" alt="Screenshot 2023-08-29 at 9 42 32 PM" src="https://github.com/microsoft/appcenter-sdk-apple/assets/1112653/a328b3a4-7d4e-4e83-9405-c90a89e2b482">

Notice: If there is no mentioned initialisation, the situation is different:

<img width="500" alt="Screenshot 2023-08-29 at 9 33 11 PM" src="https://github.com/microsoft/appcenter-sdk-apple/assets/1112653/fc4d1a39-62b1-42d4-b597-df1ea17d600a">

### **Details**

1. Which SDK version are you using?
    All available versions
2. Which OS version did you experience the issue on?
    All iOS (and apparently macOS versions)
5. Which Xcode version did you build the app with?
    Xcode 14.2
6. Which Cocoapods version are you using (run `pod --version`)?
    n/a
7. What device version did you see this error on?  Were you using an emulator or a physical device?
    All supported models
8. What language are you using?
    - [ ] Objective C
    - [x] Swift
[...]

PR #2507, opened 2023-09-20
Reduce number of threads when setting up log buffer for crashes

vmagaziy:

However the following test was failing **before** made changes:
Screenshot 2023-09-20 at 8 08 43 AM

* [ ] Did you add unit tests?
No need to add new tests as the affected functionality has been already covered
Yes, the test app which was used to identify the issue (see #2497) does not have the number of threads observed with the proposed change.

## Description

When log buffer for crashes is setup, `ms_crashes_log_buffer_size` asynchronous operations are scheduled on a hight priority GCD queue and because of the shared lock when a file is created, it potentially leads to spawning of the mentioned number of lightweight threads. This is causing the runtime pressure as executing process usually does this setup along with other initial app specific setup (which requires additional threads for asynchronous work to be done), and so this may lead to sporadic crashes.

The fix is to limit a number of concurrent operations in the mentioned setup to 5 by using `NSOperationQueue` instead of GCD dispatch queue (and group for testing purpose).

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-apple/issues/2497

## Misc

Despite the fact, dispatch queue is not supposed to spawn a number of threads over some limit leading to runtime pressure and crashes.

### DIFF
--- a/AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
+++ b/AppCenterCrashes/AppCenterCrashes/MSACCrashes.mm
@@ -174,12 +174,7 @@ __attribute__((noreturn)) static void uncaught_cxx_exception_handler(const MSACC
 /**
  * Queue with high priority that will be used to create the log buffer files. The default main queue is too slow.
  */
-@property(nonatomic) dispatch_queue_t bufferFileQueue;
-
-/**
- * A group to wait for creation of buffers in the test.
- */
-@property(nonatomic) dispatch_group_t bufferFileGroup;
+@property(nonatomic) NSOperationQueue *bufferFileQueue;
 
 /**
  * Semaphore for exclusion with "startDelayedCrashProcessing" method.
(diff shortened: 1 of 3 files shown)


drewmccormack:
This has started affecting us too, entirely on macOS 11 for some reason. Crashes on launch due to this issue.

vmagaziy:
@MikhailSuendukov -- how can failing of that test can be related to made changes? The changes are made in `App Center Crashes`, and have nothing to do with channels.

MikhailSuendukov:
I tried rerunning the tests, and it looks like this test failed due to a delay. So far, all tests have been successful.

MikhailSuendukov (approved):

DmitriyKirakosyan (approved):

vmagaziy:
What's the difference between `microsoft:develop` and `develop` branches?

DmitriyKirakosyan:
It is the same branch. It is displayed `microsoft:develop` for this PR because it was opened from a fork repository.

Issue #2297, opened 2021-05-02
Distribute as binary Swift package


apps4everyone:
Hey, would like not to compile the SDK every time. maybe it is possible to release also as binary frameworks as Swift package to make it possible to consume like this: (maybe the binaryTarget `zip/XCFrameworks` has to be split in `AppCenter`, `AppCenterAnalytics`, `AppCenterCrashes` and `AppCenterDistribute`)

```swift
// swift-tools-version:5.3

import PackageDescription

let package = Package(
    name: "AppCenter",
    products: [
        .library(name: "AppCenter", targets: ["AppCenter"]),
        .library(name: "AppCenterAnalytics", targets: ["AppCenterAnalytics"]),
        .library(name: "AppCenterCrashes", targets: ["AppCenterCrashes"]),
        .library(name: "AppCenterDistribute", targets: ["AppCenterDistribute"])
    ],
    dependencies: [
    ],
    targets: [
        .binaryTarget(
            name: "AppCenter",
            url: "https://github.com/microsoft/appcenter-sdk-apple/releases/download/4.1.1/AppCenter-SDK-Apple-XCFramework-4.1.1.zip",
            checksum: "some-checksum"
        )
    ]
)
```

Infos here: https://www.wwdcnotes.com/notes/wwdc20/10147/

DmitriyKirakosyan:
Hi @apps4everyone 

Sorry for the delay and thank you for bringing it to our attention.

I created a work item for this feature request.

apps4everyone:
thx @DmitriyKirakosyan, can I somehow see the progress (ticket)? Also for https://github.com/microsoft/appcenter-sdk-apple/issues/2298 ?

DmitriyKirakosyan:
We will keep you updated in your github issues if anything new is merged or released.

arielelkin:
@DmitriyKirakosyan @aleksandr-dorofeev @AnastasiaKubova 

This feature would be very useful, because compiling the entire swift package takes up a lot of time (both locally and on CI)

Any updates on this?

DmitriyKirakosyan:
Hi @arielelkin , I agree it would be better to distribute xcframework via SPM as we do with Cocoapods, however there is an issue with SPM which I was not able to fix. It doesn't properly link translation resources for Distribute module. Here is my attempt to distribute binary via SPM: https://github.com/DmitriyKirakosyan/mobile-center-sdk-ios/blob/spm-xcframework/Package.swift.
[...]

PR #2524, opened 2024-01-08
add support to override default bundle for resources

ilendemli:

* [ ] Did you add unit tests?

## Description

Adds a nullable property which can be used to override the default bundle for Distribute to search for its resources.

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-apple/issues/2297

## Misc

This is a workaround to use binary swift packages to improve build times. Each xcframework has to be zipped and uploaded separately. Example swift package release can be found here https://github.com/ilendemli/appcenter-sdk-apple-spm/releases/tag/5.0.4

### DIFF
--- a/AppCenterDistribute/AppCenterDistribute/MSACDistribute.h
+++ b/AppCenterDistribute/AppCenterDistribute/MSACDistribute.h
@@ -53,6 +53,11 @@ typedef NS_ENUM(NSInteger, MSACUpdateTrack) {
 @property(class, nonatomic) MSACUpdateTrack updateTrack;
 
 /**
+ * Overrides default bundle for resources to be searched in
+ */
+@property(class, nullable, nonatomic) NSBundle* resourceBundle;
+
+/**
  * Distribute delegate
  *
  * @discussion If Distribute delegate is set and releaseAvailableWithDetails is returning <code>YES</code>, you must call
--- a/AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
+++ b/AppCenterDistribute/AppCenterDistribute/MSACDistribute.m
@@ -72,6 +72,8 @@ static dispatch_once_t onceToken;
 
 @synthesize updateTrack = _updateTrack;
 
+@synthesize resourceBundle = _resourceBundle;
+
 #pragma mark - Service initialization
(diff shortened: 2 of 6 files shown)


ilendemli:
@microsoft-github-policy-service agree

ilendemli:
I am also creating a workflow, to automatically separate the XCFrameworks from this repository and to release it on GitHub. You can feel free to fork that repository too after this has been merged, so it is official: https://github.com/ilendemli/appcenter-sdk-apple-spm/tree/workflow

MikhailSuendukov:
Hi @ilendemli Hello and thank you for your contribution. Could you please describe in more detail how your pull request fixes the problem that was described in the issue and what is the use case for this feature?

MikhailSuendukov:
Thank you very much for the information, but could you please clarify which application you mean?

ilendemli:
I meant for you to create a new app project, and to integrate AppCenter SDK using a precompiled binary swift package.

MikhailSuendukov:
The fact that an additional repository is used to solve this is a problem. Is it possible to fix the problem within our repository only?

ilendemli:
Why is an additional repository a problem? That repository will maintain itself automatically with GitHub actions and people will have the option to either use appcenter from code or from binary.

DmitriyKirakosyan:
@ilendemli While we appreciate the effort you've put into implementing this feature, to ensure proper functionality, the integration needs to be conducted directly from this repository. As such, using an alternative integration method, like adding SPM through your repository's link, remains unsupported. This, in turn, means that the API introduced in this pull request will also be unsupported. Consequently, introducing an API that is known to be unsupported from the outset may not be the best course of action. If you still require this feature, I would suggest creating a fork and using that for your integration purposes.

ilendemli:
Interesting take. Maybe I did not express myself clearly but anyway. The API changes does not affect any functionality but opens up possibilities. But if this is not wanted, then I have to respect that decision.

ilendemli:
Someone else might come up with a better solution. For now, I have updated my repository to pull this repo, apply a patch, compile and bundle a binary swift package. You might still want to take the translation fixes I made here though.

PR #2530, opened 2024-02-07
[Fix] Fixes the system exception when initializing NSString from `exceptionReason` with nil value

ForeverQ:
## Description

We encountered a system exception on Outlook iOS when `[NSString stringWithString:report.exceptionInfo.exceptionReason];` tried to create a string from a nil value. This PR prevents this by checking that `exceptionReason` is not nil before using it.


MikhailSuendukov:
Hello and thanks for your contribution!
We noticed that you also opened a PR in the PLCrashReporter repository, where you also processed this behavior. I suggest continuing the discussion in that [pull request](https://github.com/microsoft/plcrashreporter/pull/298/).

DmitriyKirakosyan:
Closing due to inactivity.

